PR #18, opened 2023-08-11
[REG-1049] Add ability to customize player replay models

RG-nAmKcAz:
- Adds `isRuntimeObject` to identify assets in the scene that are dynamically spawned
- Converts the ReplayModelManager to a ScriptableObject and supports persistence of the object within the context of the user's project itself
- Adds a button to the replay window as a shortcut to editing the ReplayModelManager asset in the inspector, including a warning text if none have been configured
  - This will auto populate the ReplayModelAssociations any time you load a replay data zip with any missing entity types that aren't static
- Utilizes `isRuntimeObject` flag to know which entities in the replay to automatically add to the list of things needing custom models.
- Updates the Regression Games menu layout options
- Refactors naming to clarify player vs entity from replay data

See also (https://github.com/Regression-Games/RegressionDocs/pull/22)


Runtime/Scripts/RGBotConfigs/RGState.cs
```diff
@@ -43,6 +47,7 @@ public class RGState: MonoBehaviour
             ["id"] = this.transform.GetInstanceID(),
             ["type"] = objectType,
             ["isPlayer"] = isPlayer,
+            ["isStatic"] = isStatic,
```
RG-nAmKcAz, line 50: Maybe we name this `isSceneElement` or `isDynamicallySpawned` ?

Would those names better represent that this is a thing that isn't defined in the scene itself, but it dynamically loaded at runtime based on the game flow , player connection, game event, etc.

vontell: @rcornwall-reg any idea on what developers call this? GameObjects that are dragged into the scene within the editor vs spawned in after the scene starts?

vontell (approved):
Approving assuming we have an answer for isStatic naming (even if we keep the same name)

Editor/Scripts/RGBotReplayWindow.cs
```diff
@@ -90,7 +90,25 @@ private void OnGUI()
...
+            GUILayout.FlexibleSpace();
+                        
+
             EditorGUILayout.EndHorizontal();
+            
+            GUILayout.Space(5);
+            
+            if (!ReplayModelManager.GetInstance().HasEntries())
+            {
+                EditorGUILayout.HelpBox("Custom Replay Models have not been configured for this project.\nLoad your first replay zip to auto populate the entity types, then configure their model associations using the `Configure Custom Replay Models` button.\n(You can also manually add, edit, or remove associations manually at any time.)", MessageType.Warning, true);
```
abeizer, line 103:
```suggestion
                EditorGUILayout.HelpBox("Custom Replay Models have not been configured for this project.\nLoad your first replay zip to auto populate the entity types, then configure their model associations using the `Configure Custom Replay Models` button.\n(You can also manually add, edit, or remove associations at any time.)", MessageType.Warning, true);
```

abeizer (approved):

rcornwall-reg (approved):
Looks good to me, I like the scriptable object approach on this

rcornwall-reg:
@RG-nAmKcAz Some suggested names instead of 'isStatic':

'isRuntimeObject'
'isRuntimeGameObject'
'runtimeGameObject'
'runtimeObject'

rcornwall-reg:
@RG-nAmKcAz Approved before the changes, but everything looks good to me 👍

PR #21, opened 2023-08-15
[REG-1205] Log error for instant bot queueing

rcornwall-reg:
Looking at the comments here:
https://linear.app/regression-games/issue/REG-1205/log-error-for-instant-bot-queueing

It's possible no further information is available, except to determine which service has failed. Is the value added by the warning messages worth the extra inspector logs?

### DIFF
--- a/Runtime/Scripts/RGOverlayMenu.cs
+++ b/Runtime/Scripts/RGOverlayMenu.cs
@@ -28,8 +28,8 @@ namespace RegressionGames
         public TMP_Dropdown nextBotDropdown;
         
         private static List<RGBotInstance> activeBots = new List<RGBotInstance>();
-
-        private List<long> invalidBotIds = new List<long>();
+        
+        private ConcurrentBag<long> invalidBotIds = new ConcurrentBag<long>();
 
         private int lastCount = -1;
 
--- a/Runtime/Scripts/RGOverlayMenu.cs
+++ b/Runtime/Scripts/RGOverlayMenu.cs
@@ -29,6 +29,8 @@ namespace RegressionGames
         
         private static List<RGBotInstance> activeBots = new List<RGBotInstance>();
 
+        private List<long> invalidBotIds = new List<long>();
+
         private int lastCount = -1;
 
         private static RGOverlayMenu _this = null;
@@ -205,6 +207,7 @@ namespace RegressionGames
                                 },
                                 () =>
                                 {
+                                    invalidBotIds.Add(bot.id);
                                     if (Interlocked.Increment(ref count) >= bots.Length)
                                     {
                                         // we hit the last async result.. do the update processing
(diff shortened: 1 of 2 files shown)


Runtime/Scripts/RGServiceManager.cs
```diff
@@ -198,6 +203,7 @@ public async Task QueueInstantBot(long botId, Action<RGBotInstance> onSuccess, A
                 },
                 onFailure: async (f) =>
                 {
+                    RGDebug.LogWarning($"Failed to queue instance bot: {f}");
```
RG-nAmKcAz, line 206: This is already logged by the caller of this method (but without the id) in `RGOverlayMenu.AddBot`

Runtime/Scripts/RGServiceManager.cs
```diff
@@ -220,6 +226,7 @@ public async Task GetRunningInstancesForBot(long botId, Action<RGBotInstance[]>
                 },
                 onFailure: async (f) =>
                 {
+                    RGDebug.LogWarning($"Failed to get running instance for bot: {f}");
```
RG-nAmKcAz, line 229: This seems valid, but very spammy.  Instead, we should handle this in `RGOverlayMenu.UpdateBots` and consolidate the errors into one log statement after all the calls finish.

Runtime/Scripts/RGServiceManager.cs
```diff
@@ -172,7 +173,11 @@ public async Task GetExternalConnectionInformationForBotInstance(long botInstanc
                         RGDebug.LogDebug($"RG Bot Instance external connection info: {connInfo}");
                         onSuccess.Invoke(connInfo);
                     },
-                    onFailure: async (f) => { onFailure.Invoke(); }
+                    onFailure: async (f) =>
+                    {
+                        RGDebug.LogWarning($"Failed retrieving external connection information: {f}");
```
RG-nAmKcAz, line 178: This one we definitely do not want to log here.  This API polls the backend waiting for the bot to start and will repeatedly throw an http error code or timeout while the bot is still starting up.  I had a log like this in here originally and it fills up the log with spam on every bot startup as part of our normal operation.

Runtime/Scripts/RGServiceManager.cs
```diff
@@ -240,6 +247,7 @@ public async Task StopBotInstance(long botInstanceId, Action onSuccess, Action o
                 },
                 onFailure: async (f) =>
                 {
+                    RGDebug.LogWarning($"Failed to stop bot instance: {f}");
```
RG-nAmKcAz, line 250: All http errors are already logged at debug level, but if we want this specifically logged at warning, we should add the botInstanceId into this message so that we can help them debug from our end.

RG-nAmKcAz (changes requested):

Runtime/Scripts/RGServiceManager.cs
rcornwall-reg: I like that idea - I consolidated them in the calling method

Runtime/Scripts/RGServiceManager.cs
rcornwall-reg: Agreed, I removed it

Runtime/Scripts/RGServiceManager.cs
rcornwall-reg: True, it looks like all calling methods handle their errors so I'm going to remove it from this location

Runtime/Scripts/RGOverlayMenu.cs
```diff
@@ -200,32 +202,40 @@ public void UpdateBots()
                                     if (Interlocked.Increment(ref count) >= bots.Length)
                                     {
                                         // we hit the last async result.. do the update processing
-                                        processBotUpdateList(instances);
+                                        ProcessBotUpdateList(instances);
                                     }
                                 },
                                 () =>
                                 {
+                                    invalidBotIds.Add(bot.id);
```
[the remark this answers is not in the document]
rcornwall-reg: Updated from a List to a thread-safe ConcurrentBag 👍

RG-nAmKcAz (approved):

PR #22, opened 2023-08-15
[REG-1206] Namespace the SDK

rcornwall-reg:
Added namespaces:
* RGState
* RGFindUtils

Fixed namespaces:
* RGIconPulse
* ReplayModelManager

### DIFF
--- a/Runtime/Scripts/RGBotConfigs/RGState.cs
+++ b/Runtime/Scripts/RGBotConfigs/RGState.cs
@@ -24,7 +24,8 @@ namespace RegressionGames.RGBotConfigs
         [Tooltip("Is this object spawned during runtime, or a fixed object in the scene?")]
         public bool isRuntimeObject = false;
 
-        [Header("3D Positioning")] public bool syncPosition = true;
+        [Header("3D Positioning")] 
+        public bool syncPosition = true;
         public bool syncRotation = true;
 
         /**
--- a/Runtime/Scripts/RGBotConfigs/RGState.cs
+++ b/Runtime/Scripts/RGBotConfigs/RGState.cs
@@ -9,51 +9,52 @@ using UnityEngine;
  * TODO: Can we use a generic type instead of a dictionary? That way users can
  *       debug and use the states within their own code?
  */
-public class RGState: MonoBehaviour
+namespace RegressionGames.RGBotConfigs
 {
+    public class RGState : MonoBehaviour
+    {
 
-    [Header("General Information")]
-    [Tooltip("Does this object represent a human/bot player ?")]
-    public bool isPlayer;
-    [Tooltip("A type name for associating like objects in the state")]
-    public string objectType;
+        [Header("General Information")] [Tooltip("Does this object represent a human/bot player ?")]
+        public bool isPlayer;
 
-    [Header("3D Positioning")]
-    public bool syncPosition = true;
-    public bool syncRotation = true;
+        [Tooltip("A type name for associating like objects in the state")]
+        public string objectType;
 
-    /**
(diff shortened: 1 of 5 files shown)


Runtime/Scripts/RGOverlayMenu.cs
```diff
@@ -173,6 +173,17 @@ public void StopBotInstance(long id)
             RGBotServerListener.GetInstance()?.TeardownClient((uint) id);
         }
 
+        /// <summary>
+        /// Tear down all currently active bots
+        /// </summary>
+        public void StopAllBots()
```
RG-nAmKcAz, line 179: Looks like this accidentally included work from your other branch

rcornwall-reg:
Also, we need to be aware that manipulating namespaces breaks backwards compatibility.

Scripts currently inheriting from RGState, or accessing an RG script via a different namespace will throw compiler errors until the correct imports are added

RG-nAmKcAz:
> Also, we need to be aware that manipulating namespaces breaks backwards compatibility.
> 
> Scripts currently inheriting from RGState, or accessing an RG script via a different namespace will throw compiler errors until the correct imports are added

These should be fixed as part of this Linear Task, whether in this project, or with PRs for the affected projects.

rcornwall-reg:
We can fix our examples (UnitySample, BossRoom, and ThirdPersonDemo). I was more thinking any customers currently using the SDK, but if it's only our samples then it's not too bad

RG-nAmKcAz:
Yep, just our samples,  Release notes for next release would touch on this type of thing for customer (Unleashed Games) if they've been able to start by then.

Runtime/Scripts/RGOverlayMenu.cs
rcornwall-reg: Resolved. I committed a delete to avoid conflicts with other pulled versions. It _does_ muddle up the commit history so if you'd prefer to see a rollback + force push just let me know

RG-nAmKcAz (approved):

Runtime/Scripts/RGBotConfigs/RGState.cs
```diff
@@ -9,56 +9,56 @@
...
-    public string objectType;
-    
-    // this is used in our toolkit to understand which things would need dynamic models
-    [Tooltip("Is this object spawned during runtime, or a fixed object in the scene?")]
-    public bool isRuntimeObject = false;
-    
-    [Header("3D Positioning")]
-    public bool syncPosition = true;
-    public bool syncRotation = true;
+        [Header("3D Positioning")] public bool syncPosition = true;
```
vontell, line 27: Is this new formatting intentional?

vontell (approved):

Runtime/Scripts/RGBotConfigs/RGState.cs
rcornwall-reg: Adding the namespace at the top layer of the file reformats the entire file with an indentation:
**Before**
```
public class MyClass {
   [Header("My Header")]
   public bool myVar = true;
}
```

**After**
```
namespace RegressionGames
{
   public class MyClass {
      [Header("My Header")]
      public bool myVar = true;
   }
}
```

I think that's what we're seeing here

vontell: Ah here I am referring to the fact that the header is on the same line as the syncPosition variable, is that ok/standard?

rcornwall-reg: Ohh I see what you're saying. I've seen both, as long as its consistent. In this case consistency calls for a new line 👍

PR #25, opened 2023-08-22
[REG-1099] Utils and env vars for CI/CD

vontell:
Adds a few features to the SDK to support CI/CD usage.

* A new "HasBotsRunning()" function that is used during tests to figure out when to terminate (see other PR at end of this description for an example)
* Adds API Key and Host env vars that can override the project settings in CI/CD

Example: https://github.com/Regression-Games/RGUnitySample/pull/6

### DIFF
--- a/Runtime/Scripts/RGAgentTypes.meta
+++ b/Runtime/Scripts/RGAgentTypes.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: f802639f659c54a1e94229184b7052c9
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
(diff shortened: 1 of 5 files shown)


vontell:
@rcornwall-reg I'd like to get this merged quickly so I can test this new code with my sample (which is downloaded via github, so I need this to be on the main branch). Example here: https://github.com/Regression-Games/RGUnitySample/pull/6

Once I get that sample working, I will be making a tutorial with BossRoom.

rcornwall-reg (approved):
I don't see anything sensitive here, but is there any risk in keeping keys like API_KEY checked into our source?

vontell:
@rcornwall-reg can you elaborate? With this code, the API Key is never checked into the source - it is added as an environment variable within the GitHub Actions secrets feature, so the API key will never be saved into a project or in the repo itself :D

rcornwall-reg:
That's what I was wondering ^ Sounds good 👍

PR #28, opened 2023-08-23
[REG-1210] Add a custom text mesh asset to avoid conflicts

rcornwall-reg:
Adds an internal TMP font and updates reference on Regression prefabs to reference internal font.

Speaking to some concerns here: https://linear.app/regression-games/issue/REG-1210/add-a-custom-text-mesh-asset-to-avoid-conflicts

I believe include a custom TMP font and material, instead of all TMP assets, in our SDK is enough to remove any conflicts with user TMP configurations. No updates on the samples should be required. It's also possible including more than a font asset (like an internal TMP copy) could lead to conflicts while installing the SDK in apps that already have their own TMP versions & configurations.

I tested on two new projects - the first had a TMP installed already and a custom text configuration. Installing the SDK used the packaged font asset and materials, so it looked as expected. The second project was empty without TMP installed. Adding the SDK prompted the install of TMP essentials, after which it worked as expected.


vontell (approved):

vontell:
@rcornwall-reg once this is ready for review again, make sure to re-request Zack

RG-nAmKcAz (approved):

PR #32, opened 2023-09-07
[REG-1212] Wrap up RGAction implementation

rcornwall-reg:
**Changes**
* RGActions can be defined through a [RGAction] attribute
```
[RGAction]
public void Jump()
{
    _rigidbody.AddForce(Vector3.up * jumpPower, ForceMode.Impulse);
}

[RGAction("Big Jump")]
public void DoubleJump(int height, UnityEngine.Vector3 position)
{
    Debug.Log("Double Jump");
}
```
* Code Generators will create the following files, in `Assets/RGScripts`. It will create the folder if it does not exist yet:
1. `RGSerialization.cs` - Custom JSON deserializers for non-primitive parameters. It uses a simple `ToJson` / `FromJson` method of serialization, so it only works for simple cases, like Vector2, Vector3. Nested custom classes as parameters currently aren't working due to this limitation. Future PR's should expand on this capability to remove any limits on parameter types
2. `RGAction_{ActionName}.cs` - Each method with the `RGAction` attribute will generate a class that inherits from RGAction. It will automatically create a link back to the original method, with parameters, and fill out RGAction's required methods (GetActionName, StartAction). It will also pull out the parameters from the `Dictionary<string, object>` passed into `StartAction`. It gets the dictionary key from the parameter's name, which means the parameters passed to `StartAction` need to match the names of the original method.
3. `RGActionMap.cs` - A helper method that adds the correct RGAction scripts to the GameObject at Runtime. I still need some input on this, but so far a generated helper is much better than adding them manually!

**Note**
Go to `RegressionGames->Compile` to run the script analysis and generation. I'm going to attach it to the normal script compilation process, but this is better for testing


Editor/Scripts/CodeGenerators/GenerateRGActionClasses.cs
```diff
@@ -0,0 +1,257 @@
...
+                string tryParseStatement;
+
+                if (paramType.ToLower() == "string" || paramType.ToLower() == "system.string")
+                {
+                    tryParseStatement = $"{paramName} = {paramName}Input;";
+                }
+                else if (RGUtils.IsCSharpPrimitive(paramType))
+                {
+                    tryParseStatement = $"{paramType}.TryParse({paramName}Input, out {paramName});";
+                }
```
vontell, line 213: Instead of treating a string separately, can't we have the first pull from the `input[paramName]` be and `object` type, and then do a cast? Rather than a parse? Not a big deal, just curious if you tried that too.

Editor/Scripts/CodeGenerators/GenerateRGActionClasses.cs
```diff
@@ -0,0 +1,257 @@
...
+
+                var tryBlock = SyntaxFactory.Block(SyntaxFactory.SingletonList<StatementSyntax>(
+                    SyntaxFactory.ParseStatement(tryParseStatement)
+                ));
+
+                var catchBlock = SyntaxFactory.CatchClause()
+                                              .WithDeclaration(SyntaxFactory.CatchDeclaration(SyntaxFactory.ParseTypeName("Exception")))
+                                              .WithBlock(SyntaxFactory.Block(SyntaxFactory.SingletonList<StatementSyntax>(
+                                                  SyntaxFactory.ParseStatement($"RGDebug.LogError(\"Failed to parse '{paramName}'\");")
```
vontell, line 226: Can we also add the exception message / stack trace to the log?

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
```diff
@@ -0,0 +1,111 @@
...
+                        ActionName = actionName,
+                        Parameters = parameterList
+                    });
+                }
+            }
+
+            string jsonResult =
+                JsonConvert.SerializeObject(new RGActionsInfo { BotActions = botActionList }, Formatting.Indented);
+            
+            // TODO: send json result to server for typedef generation
```
RG-nAmKcAz, line 85: TODOs need to refer to a linear task #

RG-nAmKcAz (changes requested):
Looks ok, but I don't have a feel for how this will workout without a PR showing how it integrates into the Sample Project or 3rd Party demo.  Do you have a PR for one of those projects we can use as a reference, maybe whatever you used for your testing ?

- I think the `RegressionGames->Compile` menu option should be named something like `Generate Action Classes` to be more explicit.
- Without the sample PR, I can't yet grasp RGActionMap's use case.  Won't this end up needing to reference potentially hundreds of game objects to make these associations ?  I'm feeling like explicitly putting the Actions on the GameObjects in the editor or on their prefabs will be more clear, but will wait to see a demo PR.

Editor/Scripts/CodeGenerators/GenerateRGActionMapClass.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.IO;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using UnityEditor;
+
+namespace RegressionGames
+{
+    public static class GenerateRGActionMapClass
```
vontell, line 12: A comment showing what exactly this generates would be helpful!

Editor/Scripts/CodeGenerators/GenerateRGSerializationClass.cs
```diff
@@ -0,0 +1,105 @@
...
+                                        SyntaxKind.SimpleMemberAccessExpression,
+                                        SyntaxFactory.ParseTypeName("JsonUtility"),
+                                        SyntaxFactory.GenericName(SyntaxFactory.Identifier("FromJson"))
+                                            .WithTypeArgumentList(SyntaxFactory.TypeArgumentList(
+                                                SyntaxFactory.SingletonSeparatedList<TypeSyntax>(
+                                                    SyntaxFactory.ParseTypeName(parameter.Type))))),
+                                    SyntaxFactory.ArgumentList(SyntaxFactory.SingletonSeparatedList(
+                                        SyntaxFactory.Argument(SyntaxFactory.IdentifierName("paramJson"))))))));
+
+                        methodDeclarations.Add(method);
```
vontell, line 84: Little hard to follow this - might be useful to add a comment that shows an example of what may be generated here

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
vontell: Sorry @rcornwall-reg yeah we may have not covered this - for any TODO, we make a linear task with more info about the TODO and then link it in, i.e.

```
// TODO(REG-1234): send json result to server for typedef generation
```

Editor/Scripts/RGActionInfo.cs
```diff
@@ -0,0 +1,15 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace RegressionGames
+{
+    [System.Serializable]
+    public class RGActionInfo
+    {
+        public string Object;
```
vontell, line 10: Can we add some comments here? Wasn't sure what Object was.

vontell (changes requested):
Looking good! As @RG-nAmKcAz  mentioned, would love to see an example, and I'll try this out too. Looking forward to the docs as started from @abeizer.

Most of my comments are about comments!

Editor/Scripts/CodeGenerators/GenerateRGActionClasses.cs
```diff
@@ -0,0 +1,257 @@
...
+            var methodParameters = new List<ParameterSyntax>
+            {
+                SyntaxFactory.Parameter(SyntaxFactory.Identifier("input"))
+                             .WithType(SyntaxFactory.ParseTypeName("Dictionary<string, object>"))
+            };
+
+            var parameterParsingStatements = new List<StatementSyntax>();
+            var methodInvocationArguments = new List<string>();
+
+            foreach (var parameter in action.Parameters)
```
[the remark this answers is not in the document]
rcornwall-reg: Added 👍

Editor/Scripts/CodeGenerators/GenerateRGActionMapClass.cs
rcornwall-reg: Added 👍

Editor/Scripts/CodeGenerators/GenerateRGSerializationClass.cs
rcornwall-reg: Added 👍

Editor/Scripts/RGActionInfo.cs
rcornwall-reg: Added 👍

vontell (approved):
Almost LGTM! I think just one thing was unadressed? I'll approve on my end for now assuming that gets addressed (my comment about exception message).

Also, in the future, we always have the reviewers resolve the conversations, since they are the ones verifying that you addressed the issue. Usually what we do is say "Fixed" or some message indicating we resolved it, but leave the actual resolving to the reviewer.

RG-nAmKcAz (approved):
Approving pending aaron's review.

also see my note in Linear on task REG-1251

Editor/Scripts/CodeGenerators/GenerateRGActionClasses.cs
rcornwall-reg: Fixed in latest commit

vontell:
@rcornwall-reg if you change works, then feel free to merge! LGTM!

PR #37, opened 2023-09-19
[REG-1232] Prototype, design, and implement a new RGState approach that uses attributes

rcornwall-reg:
Define RGStates through an attribute:

1. Public fields
   ```
   [RGState]
   public int myVariable;
   ```
2. Public getter methods. (No parameters)
   ```
   [RGState]
   public int GetVariable() { return myVariable; }
   ```
   
Override the name of the state by passing in a string to the attribute
```
[RGState("My State")]
public int foo;
```
```
[RGState("My State")]
public int GetFoo() { return foo;}
```

Once the attributes are added, run `Regression Games -> Generate Scripts` from the menu

Generated State script will go to `Assets/RGScripts/RGStates/`

Add the state scripts to the GameObject and fill out the inspector fields for `isPlayer`, `objectType`, `isRuntimeObject`, and position/rotation sync

Screen Shot 2023-09-19 at 11 33 01 AM

### DIFF
--- a/Editor/Scripts/CodeGenerators/GenerateRGStateClasses.cs
+++ b/Editor/Scripts/CodeGenerators/GenerateRGStateClasses.cs
@@ -62,7 +62,7 @@ namespace RegressionGames
                 string headerComment = "/*\n* This file has been automatically generated. Do not modify.\n*/\n\n";
 
                 // Save to 'Assets/RGScripts/RGSerialization.cs'
-                string subfolderName = "RGScripts";
+                string subfolderName = Path.Combine("RGScripts", "RGStates");               
                 string fileName = $"{className}.cs";
                 string filePath = Path.Combine(Application.dataPath, subfolderName, fileName);
                 string fileContents = headerComment + formattedCode;
--- a/Editor/Scripts/CodeGenerators/GenerateRGStateClasses.cs
+++ b/Editor/Scripts/CodeGenerators/GenerateRGStateClasses.cs
@@ -1,0 +1,276 @@
+using Newtonsoft.Json;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using System;
+using UnityEditor;
+using UnityEngine;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace RegressionGames
+{
+    public static class GenerateRGStateClasses
+    {
+        public static void Generate(string jsonInput)
+        {
+            var rgStateInfos = JsonConvert.DeserializeObject<RGStateInfoWrapper>(jsonInput).RGStateInfo;
+
+            foreach (var rgStateInfo in rgStateInfos)
+            {
+                var className = $"{rgStateInfo.Object}_RGState";
+                var componentType = rgStateInfo.Object;
+
+                // Create a new compilation unit
+                var compilationUnit = SyntaxFactory.CompilationUnit()
+                    .AddUsings(SyntaxFactory.UsingDirective(SyntaxFactory.ParseName("System")),
+                        SyntaxFactory.UsingDirective(SyntaxFactory.ParseName("System.Collections.Generic")));
+
+                // Create a new class declaration with the desired name
+                var classDeclaration = SyntaxFactory.ClassDeclaration(className)
+                    .AddModifiers(SyntaxFactory.Token(SyntaxKind.PublicKeyword))
+                    .AddBaseListTypes(SyntaxFactory.SimpleBaseType(SyntaxFactory.ParseTypeName("RGState")));
+
+                // Create the private field
+                var fieldDeclaration = SyntaxFactory.FieldDeclaration(
+                        SyntaxFactory.VariableDeclaration(SyntaxFactory.ParseTypeName(componentType))
+                            .AddVariables(SyntaxFactory.VariableDeclarator("myComponent")))
+                    .AddModifiers(SyntaxFactory.Token(SyntaxKind.PrivateKeyword));
+
+                // Create the Start method
+                var startMethod = GenerateStartMethod(componentType, rgStateInfo.State);
+
+                // Create the GetState method
+                var getStateMethod = GenerateGetStateMethod(componentType, rgStateInfo.State);
+
+                // Add the members to the class declaration
+                classDeclaration = classDeclaration
+                    .AddMembers(fieldDeclaration, startMethod, getStateMethod);
+
+                // Create namespace
+                var namespaceDeclaration = SyntaxFactory.NamespaceDeclaration(SyntaxFactory.ParseName("RegressionGames.RGBotConfigs"))
+                                                       .AddMembers(classDeclaration);
+
+                // Add the namespace declaration to the compilation unit
+                compilationUnit = compilationUnit.AddMembers(namespaceDeclaration);
+
+                // Get the full code text
+                var formattedCode = compilationUnit.NormalizeWhitespace().ToFullString();
+
+                // Write the code to a .cs file
+                string headerComment = "/*\n* This file has been automatically generated. Do not modify.\n*/\n\n";
+
+                // Save to 'Assets/RGScripts/RGSerialization.cs'
+                string subfolderName = "RGScripts";
+                string fileName = $"{className}.cs";
+                string filePath = Path.Combine(Application.dataPath, subfolderName, fileName);
+                string fileContents = headerComment + formattedCode;
+                Directory.CreateDirectory(Path.GetDirectoryName(filePath));
+                File.WriteAllText(filePath, fileContents);
+                RGDebug.Log($"Successfully Generated {filePath}");
+                AssetDatabase.Refresh();
+            }
+        }
+
+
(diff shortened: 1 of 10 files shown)


Editor/Scripts/CodeGenerators/GenerateRGStateClasses.cs
```diff
@@ -0,0 +1,276 @@
...
+                // Add the namespace declaration to the compilation unit
+                compilationUnit = compilationUnit.AddMembers(namespaceDeclaration);
+
+                // Get the full code text
+                var formattedCode = compilationUnit.NormalizeWhitespace().ToFullString();
+
+                // Write the code to a .cs file
+                string headerComment = "/*\n* This file has been automatically generated. Do not modify.\n*/\n\n";
+
+                // Save to 'Assets/RGScripts/RGSerialization.cs'
```
vontell, line 64: Should this say `"Save to 'Assets/RGScripts/RGStates/{name}.cs'"`

Editor/Scripts/CodeGenerators/GenerateRGStateClasses.cs
```diff
@@ -0,0 +1,276 @@
...
+using UnityEngine;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace RegressionGames
+{
+    public static class GenerateRGStateClasses
+    {
+        public static void Generate(string jsonInput)
```
vontell, line 16: Is there an example in some repo of the generated file?

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
```diff
@@ -100,6 +106,103 @@ private static void SearchForBotActionMethods()
...
+                            Object = className,
+                            State = stateList
+                        });
+                    }
+                }
+            }
+
+            string jsonResult = JsonConvert.SerializeObject(new { RGStateInfo = rgStateInfoList }, Formatting.Indented);
+            
+            // TODO (REG-1267): send json result to server for typedef generation
```
vontell, line 189: Is this still a TODO? It seems like rn we are fine with the typdef generation happening here? Or by typedef do you mean code generation for bots, not generating the JSON?

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
```diff
@@ -100,6 +106,103 @@ private static void SearchForBotActionMethods()
...
+            
+            // remove previous RGActions
```
vontell, line 191: This should say RGStates I believe?

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
```diff
@@ -100,6 +106,103 @@ private static void SearchForBotActionMethods()
...
+                        rgStateInfoList.Add(new RGStatesInfo
+                        {
+                            Object = className,
+                            State = stateList
+                        });
+                    }
+                }
+            }
+
+            string jsonResult = JsonConvert.SerializeObject(new { RGStateInfo = rgStateInfoList }, Formatting.Indented);
```
vontell, line 187: Seems like we are serializing just to deserialize in the generate call? Is this because we eventually plan to send it somewhere?

vontell: I guess this is the spot where we could write it and then that is what Abby will be able to upload?

Runtime/Scripts/RGBotConfigs/RGState.cs
```diff
@@ -28,6 +30,30 @@ public class RGState : MonoBehaviour
...
+        protected object InvokeMethod(string methodName, Type returnType)
+        {
+            if (methodMap.TryGetValue(methodName, out var del))
+            {
+                return Convert.ChangeType(del.DynamicInvoke(), returnType);
+            }
+
+            RGDebug.LogError($"Method '{methodName}' not found");
+            return null;
+        }
```
vontell, line 55: Is there a reason why we need to provide an Invoke pattern, vs just calling the function directly in the generated code?

vontell (changes requested):

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
rcornwall-reg: Actually for now, because we're doing user upload of the data, it would be something like:
```
// TODO (REG-1267) Export JSON for user upload
```
The ticket: https://linear.app/regression-games/issue/REG-1267/unity-agent-builder-crossover-export-actions-and-state

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
rcornwall-reg: Yeah exactly - I'm serializing to export it as a `.json` file, so they can upload it for @abeizer 's work

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
vontell: Ah ok, thanks!

Runtime/Scripts/Attributes/RGStateAttribute.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        // This constructor allows the attribute to be used without a name parameter, e.g., [RGState]
+        public RGStateAttribute()
+        {
+        }
+
+        // This constructor allows a name parameter to be specified, e.g., [RGState("Some Name")]
+        public RGStateAttribute(string stateName)
+        {
+            StateName = stateName;
```
abeizer, line 18: This question is likely out of scope for this task, but in the spirit of thinking ahead... based on Aaron's proposal for the bot dashboard design, we know we're going to want additional params that customize how the attribute is displayed in the client (ex. type = "bar", min = "0", max = maxHealth). So question is, can these attributes accept a map or an instance of a "config" object as a parameter? Something that would allow me to specify only the fields that I want to (ie. I don't need to pass a name or type if I really only want to set a min and max value), and won't require us to have many different constructors. From a quick search, it looks like this may not be possible, so we'll probably want to think about alternatives to keep these easy for the user to define and customize

abeizer:
> Add the state scripts to the GameObject and fill out the inspector fields for isPlayer, objectType, isRuntimeObject, and position/rotation sync

There's been a lot of discussion around RGStates this past cycle. I'd like to have a clear understanding of what our final implementation should look like for these and what the user experience for that will look like, and what work we still need to do to reach that based on this PR. For example, we've discussed things like squashing state scripts versus allowing multiple state scripts and therefore multiple objectTypes, whether we want users to define the ObjectType inside the script using an [RGObjectType] attribute or via the editor, etc. Right now it looks like the user will need to set `isPlayer`, `ObjectType`, `isRuntimeObject`, and sync position + rotation for every RGState script they add to their prefab. This could make sense for the ObjectType depending on whether we're going for the multiple-object-type approach, but it doesn't make sense to me for the other fields... if one script says "yes I am a player" and another says "no I am not a player", then which one wins? Same for position + rotation?

Editor/Scripts/CodeGenerators/GenerateRGStateClasses.cs
rcornwall-reg: Fixed

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
rcornwall-reg: Fixed

Runtime/Scripts/RGBotConfigs/RGState.cs
rcornwall-reg: I can't see any downside - updated the generator to invoke directly

Editor/Scripts/CodeGenerators/GenerateRGStateClasses.cs
rcornwall-reg: PR'd in Unity Sample with generated state file: https://github.com/Regression-Games/RGUnitySample/pull/8

rcornwall-reg:
> > Add the state scripts to the GameObject and fill out the inspector fields for isPlayer, objectType, isRuntimeObject, and position/rotation sync
> 
> There's been a lot of discussion around RGStates this past cycle. I'd like to have a clear agreement on what our final implementation should look like for these, what the user experience for that will look like, and what work we still need to do to reach that based on this PR. For example, we've discussed things like squashing state scripts versus allowing multiple state scripts and therefore multiple objectTypes, whether we want users to define the ObjectType inside the script using an [RGObjectType] attribute or via the editor, etc. Right now it looks like the user will need to set `isPlayer`, `ObjectType`, `isRuntimeObject`, and sync position + rotation for every RGState script they add to their prefab. This could make sense for the ObjectType depending on whether we're going for the multiple-object-type approach, but it doesn't make sense to me for the other fields... if one script says "yes I am a player" and another says "no I am not a player", then which one wins? Same for runtime, position + rotation?

I agree - I think we should discuss where to go from here after this PR is in. I think it's a solid v1 but there's still a lot of unknowns

vontell (approved):

abeizer (approved):

PR #39, opened 2023-09-20
[REG-992] Show bot states on overlay and in logs

RG-nAmKcAz:
- Now shows Unity side view of the Bot state in both the overlay and in the logs

- Side Note: While doing this I noted it may be nice in the future to filter the bots shown to only be those that your Unity client is tracking, instead of all active Unity botInstances for your user

image


abeizer (approved):

rcornwall-reg (approved):

RG-nAmKcAz:
@vontell any feedback/changes before I merge this ?

vontell (approved):
Nope, looks good to me!

PR #45, opened 2023-10-04
[REG-1291] RG Bot overlay changes for Local Bots

RG-nAmKcAz:
- Update the overlay to handle local bots interleaved with Remote bots
- Gives Unity side bots a unique botId and botInstanceId based on their unique machine identifier to avoid conflicts with RG service side ids when/if we sync these
- Fixes bug(s) from REG-1298 that broke RGBot object serialization in JSON responses
- Fixes bot/instance ids to respect the 'long' value as much as possible 
  - (also see RGBossRoom PR here .... (https://github.com/Regression-Games/RGBossRoom/pull/32))
  - I think this is safe for the Sample and 3rd person projects based on a quick search in their code
- Fixes up settings menu and dynamic bot starting to handle local and remote bots interleaved

image

image


RG-nAmKcAz:
Note: I still need to merge with REG-988 once it is merged to main as noted in `RGBot.cs`

RG-nAmKcAz:
This PR is FUBAR.. re-opening... /hate Github

PR #47, opened 2023-10-05
[REG-1291] Overlay support for Unity local bots

RG-nAmKcAz:
- Update the overlay to handle local bots interleaved with Remote bots
- Gives Unity side bots a unique botId and botInstanceId based on their unique machine identifier to avoid conflicts with RG service side ids when/if we sync these
- Fixes bug(s) from REG-1298 that broke RGBot object serialization in JSON responses
- Fixes bot/instance ids to respect the 'long' value as much as possible 
  - (also see RGBossRoom PR here .... (https://github.com/Regression-Games/RGBossRoom/pull/32))
  - I think this is safe for the Sample and 3rd person projects based on a quick search in their code
- Fixes up settings menu and dynamic bot starting to handle local and remote bots interleaved

image

image


Runtime/Scripts/RGBotLocalRuntime/SampleBot/RGSampleBot.cs
```diff
@@ -20,7 +21,7 @@ protected override RGBotLifecycle GetBotLifecycle()
 
         public override void ConfigureBot(RG rgObject)
         {
-            var classIndex = new System.Random().Next(4);
+            var classIndex = new Random().Next(4);
             var classes = new string[] {"Mage", "Rogue", "Tank", "Archer"};
```
vontell, line 25: I know this was in a previous PR, but note that sometime in the next week or two, we may want to provide a default implementation that works out of the box for any game, like just printing the state that was received.

vontell (approved):
Lots of changes here... let's make sure to do a lot of testing during UAT (I will wait for this to be merged before doing my "Getting Started" walkthrough UAT task).

Runtime/Scripts/Types/RGBot.cs
```diff
@@ -11,6 +10,12 @@ public class RGBot
         public string gameEngine;
         public string programmingLanguage;
         public string codeSourceType;
+        
+        //TODO (post REG-988): Cleanup and/or remove left side of || statements
```
abeizer, line 14: Can this be addressed now?

abeizer (approved):

PR #50, opened 2023-10-10
[REG-1330] Add RGEntity that takes the place of RGAgent, and includes state settings

rcornwall-reg:
This PR is a sub-task of [REG-1294](https://linear.app/regression-games/issue/REG-1294/unity-agent-builder-crossover-add-objecttype-to-actions-and-state) In order to grab the object type of the object, we must have exactly one object type present per object. To do this, we're refactoring `RGAgent` and `RGState`. We're creating a new `RGEntity` component that will replace RGAgent and includes object-specific state information.

Changes:
1. `RGAgent` marked as obsolete. It will still work in all samples and for all users, but will show a warning to migrate over to `RGEntity`
2. `RGEntity` added, which includes all functionality of RGAgent, as well as object state information like `isPlayer`, `objectType`, `syncPosition`, `syncRotation`, and `isRuntimeObject`
3. Refactored references to RGAgent to RGEntity


Runtime/Scripts/RGBotConfigs/RGEntity.cs
```diff
@@ -0,0 +1,53 @@
...
+        public bool isRuntimeObject = false;
+
+        [Header("3D Positioning")] 
+        public bool syncPosition = true;
+        public bool syncRotation = true;
+        
+        // maps action names to RGAction components
+        private Dictionary<string, RGAction> actionMap = new Dictionary<string, RGAction>();
+            
+        // The client Id that owns this agent
```
vontell, line 26: Should this say "owns this entity if it is an agent?"

Runtime/Scripts/RGBotConfigs/RGEntity.cs
```diff
@@ -0,0 +1,53 @@
...
+        // Used as a performance optimization for mapping the ClientId into the state payloads
+        public long? ClientId = null;
+        
+        /**
+     * Updates the registry with a game object that has RGAction scripts attached to it
+     */
```
vontell, line 32: Indentation issue?

Runtime/Scripts/RGBotConfigs/RGEntity.cs
```diff
@@ -0,0 +1,53 @@
...
+        void Start()
+        {
+            var actions = GetComponents<RGAction>();
+            foreach (var action in actions)
+            {
+                actionMap[action.GetActionName()] = action;
+            }
+
+            RGDebug.LogDebug($"Agent registered with {actionMap.Count} actions");
```
vontell, line 41: So in my mental model (correct me if I am wrong), we have entities, some of which may be agents (i.e. the ones with actions are agents I suppose). If that is correct, maybe we only print this if there is at least one action? That way entities with no actions don't have a console log that say the word "Agent"

vontell (changes requested):
Some initial comments, mostly around the mental model

Runtime/Scripts/RGBotConfigs/RGEntity.cs
rcornwall-reg: That's a solid way to think about it. Still I think because we're depreciating references to RGAgent, using 'agent' as meaning 'entity with at least one action' might be confusing once 'agent' is no longer an actual class or separate object. In cases like this perhaps a better approach is 

```
if (actionMap.Count > 0)
{
    RGDebug.LogDebug($"Entity registered with {actionMap.Count} actions");
}
```

Runtime/Scripts/RGBotConfigs/RGEntity.cs
rcornwall-reg: Fixed

Runtime/Scripts/RGBotConfigs/RGEntity.cs
vontell: That makes a lot of sense to me! Happy to not add extra points of confusion :+1:

vontell (approved):

vontell:
Approved based on my remarks, curious to see if @abeizer and @RG-nAmKcAz seeing anything I missed in Unity land and Agent Builder land. Thanks!

Runtime/Scripts/RGBotConfigs/RGButtonState.cs
```diff
@@ -9,9 +9,9 @@ public class RGButtonState : RGState
 
         public RGButtonState()
         {
-            if (string.IsNullOrEmpty(this.objectType))
+            if (string.IsNullOrEmpty(rgEntity.objectType))
```
RG-nAmKcAz, line 12: I think if you try out the menu and ability bot for BossRoom with these changes you will find it to be broken.

This change would fundamentally alter how we had buttons modeled.

There is only 1 Agent for buttons and it is attached to the RGOverlayCanvas.  Any RGButtonState on any canvas is mapped to that 1 Agent.  Thus all UI button input is managed through a single Agent which basically acts as the `Human User` that would click said buttons.

I would expect that RGAgent on the RGOverlayCanvas would be changed to an RGEntity for this set of changes, but I would NOT expect every button to require an RGEntity on it as that would change the Entity/Agent simulating a single human user model.

Runtime/Scripts/RGBotConfigs/RGEntity.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        [Tooltip("A type name for associating like objects in the state")]
+        public string objectType;
+
+        // this is used in our toolkit to understand which things would need dynamic models
+        [Tooltip("Is this object spawned during runtime, or a fixed object in the scene?")]
+        public bool isRuntimeObject = false;
+
+        [Header("3D Positioning")] 
+        public bool syncPosition = true;
```
RG-nAmKcAz, line 20: @addisonbgross Note that the location of these fields to remove for REG-1303 is being moved to a new class here.

Runtime/Scripts/RGBotConfigs/RGEntity.cs
```diff
@@ -0,0 +1,56 @@
...
+        public bool syncRotation = true;
+        
+        // maps action names to RGAction components
+        private Dictionary<string, RGAction> actionMap = new Dictionary<string, RGAction>();
```
RG-nAmKcAz, line 24: ` = new ();`

RG-nAmKcAz (changes requested):
- Some big comments/discussions about `RGButtonState.cs` and its relation to Agents/Entities.
- RGOverlayCanvas prefab needs updates to move from RGAgent to RGEntity.
- Are there PRs for the other projects (BossRoom, Sample, 3rd Person, match3, etc) to update for this ?

Runtime/Scripts/RGBotConfigs/RGButtonState.cs
```diff
@@ -1,26 +1,35 @@
...
-        public RGButtonState()
-        {
-            if (string.IsNullOrEmpty(rgEntity.objectType))
-            {
-                rgEntity.objectType = "Button";
-            }
-        }
-
-        public override Dictionary<string, object> GetState()
+        public Dictionary<string, object> GetState()
```
RG-nAmKcAz, line 13: this method is left over from when it extended RGState.  I think for buttons we can just move this code into `GetGameObjectState`.

Runtime/Scripts/RGBotServerListener.cs
```diff
@@ -518,36 +518,40 @@ private void FixedUpdate()
                 // if this object is a 'player' ... put the clientId that owns it into the state
                 if (isPlayer)
                 {
-                    var rgEntity = rgState.GetComponentInParent<RGEntity>();
-                    if (rgEntity != null)
+                    var statefulObject = rgState as RGState;
+                    if (statefulObject)
```
RG-nAmKcAz, line 522: doesn't this need `is not null` or `!= null` ?

RG-nAmKcAz (changes requested):

Runtime/Scripts/RGBotServerListener.cs
rcornwall-reg: Fixed

Runtime/Scripts/RGBotConfigs/RGButtonState.cs
rcornwall-reg: Fixed

RG-nAmKcAz (approved):

PR #52, opened 2023-10-11
[REG-1308] Update characterConfig to Map and RG api enhancements

RG-nAmKcAz:
- Changes characterConfig from a plain string to a `Dictionary<string, object>` to make writing local bots somewhat sane
  - Handles/Compensates for existing remote bots sending poorly formatted JSON strings for backward compatibility
- Exposes `SceneName` as a field on `RG` accessor object
- Adds Filter Function support to `FindNearestEntity` for parity with javascript version
- Adds a `object GetField(string fieldName)` method onto `RGStateEntity` to simplify coding patterns for getting fields from state entities

See also https://github.com/Regression-Games/RegressionDocs/pull/36
See also https://github.com/Regression-Games/RGBossRoom/pull/34

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,4 @@ crashlytics-build.properties
 mono_crash*.json
 
 **/.DS_Store
+.DS_Store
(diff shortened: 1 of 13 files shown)


vontell (approved):
Makes sense to me!

abeizer (approved):

PR #55, opened 2023-10-16
[REG-1304] Move generated scripts under RegressionGames asset path

RG-nAmKcAz:
- Moves generated scripts from `Assets/RGScripts` to `Assets/RegressionGames/Runtime/GeneratedScripts`
  - Handles removing the old folder to avoid classname conflicts
- Fixes an issue where generating scripts more than once would fail to generate the zip
- Updated path for the temp zip folder to avoid confusion with the `RegressionGames` folder in `Assets`

See also https://github.com/Regression-Games/RegressionDocs/pull/37

### DIFF
--- a/Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
+++ b/Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
@@ -16,6 +16,7 @@ namespace RegressionGames
         [MenuItem("Regression Games/Generate Scripts")]
         private static void GenerateRGScripts()
         {
+            //TODO: Someone/Anyone... remove this delete RGScripts directory code after November 1st, 2023... This is temporary to help devs migrate easily
             // remove old 'RGScripts' folder that is no longer used
             string dataPath = Application.dataPath;
             string directoryToDelete = Path.Combine(dataPath, "RGScripts").Replace("\\", "/");
--- a/Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
+++ b/Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
@@ -16,6 +16,17 @@ namespace RegressionGames
         [MenuItem("Regression Games/Generate Scripts")]
         private static void GenerateRGScripts()
         {
+            // remove old 'RGScripts' folder that is no longer used
+            string dataPath = Application.dataPath;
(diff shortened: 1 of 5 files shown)


Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
```diff
@@ -16,6 +16,17 @@ public class RGActionCodeGenerator
...
+            // remove old 'RGScripts' folder that is no longer used
+            string dataPath = Application.dataPath;
+            string directoryToDelete = Path.Combine(dataPath, "RGScripts").Replace("\\", "/");
+
+            if (Directory.Exists(directoryToDelete))
+            {
+                Directory.Delete(directoryToDelete, true);
+                File.Delete(directoryToDelete + ".meta");
+                AssetDatabase.Refresh();
+            }
```
vontell, line 28: We can probably make a ticket to remove this at some point? I think we should get in the habit of cleaning up old cold that becomes outdated. Thoughts on that process?

vontell (approved):

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
RG-nAmKcAz: I'm almost inclined to put a TODO here without a task # that lists a remove after X date indicator on it.  If someone is in there changing and sees it great.

I'm hesitant to suggest we create tasks for cleanup of single lines like this as historically all I've ever seen that do is add tasks to the backlog that never get prioritized :/

vontell: That's fine with me! I think a TODO is fine just so we know that this code can be removed in the future.

vontell: The other approach, if you wanted to do a ticket, is to make a ticket and explicitly put it into the cycle that is 2 cycles ahead from now, but I am fine with either way.

addisonbgross (approved):

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
RG-nAmKcAz: todo added

PR #56, opened 2023-10-16
[REG-1302] Adds typed `RGActionRequest`s to generated code

RG-nAmKcAz:
- Generates new classes extending RGActionRequest for each action to provide typed action requests when coding in C#
- Fixes the issue with namespaces for actions in the action map and in the generated action class
- Sample
```csharp
    public class RGActionRequest_PerformAttack : RGActionRequest
    {
        public RGActionRequest_PerformAttack(bool isEnabled, string actionName, int attackNumber, float damageAmount)
        {
            action = "PerformAttack";
            Input = new ()
            {{"isEnabled", isEnabled}, {"actionName", actionName}, {"attackNumber", attackNumber}, {"damageAmount", damageAmount}, };
        }
    }
```

See also https://github.com/Regression-Games/RegressionDocs/pull/38

### DIFF
--- a/Editor/Scripts/CodeGenerators/GenerateRGActionClasses.cs
+++ b/Editor/Scripts/CodeGenerators/GenerateRGActionClasses.cs
@@ -19,6 +19,7 @@ namespace RegressionGames
             // Iterate through BotActions
             foreach (var botAction in actionsInfo.BotActions)
             {
+                
                 // Create a new compilation unit
                 CompilationUnitSyntax compilationUnit = SyntaxFactory.CompilationUnit()
                     .AddUsings(
@@ -26,7 +27,10 @@ namespace RegressionGames
                         SyntaxFactory.UsingDirective(SyntaxFactory.ParseName("System.Collections.Generic")),
                         SyntaxFactory.UsingDirective(SyntaxFactory.ParseName("RegressionGames")),
                         SyntaxFactory.UsingDirective(SyntaxFactory.ParseName("RegressionGames.RGBotConfigs")),
+                        SyntaxFactory.UsingDirective(SyntaxFactory.ParseName("RegressionGames.StateActionTypes")),
+                        SyntaxFactory.UsingDirective(SyntaxFactory.ParseName(botAction.Namespace)),
                         SyntaxFactory.UsingDirective(SyntaxFactory.ParseName("UnityEngine"))
+                        
                     )
                     .AddMembers(
                         // Namespace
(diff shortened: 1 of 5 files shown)


Editor/Scripts/CodeGenerators/GenerateRGActionClasses.cs
```diff
@@ -19,36 +19,51 @@ public static void Generate(string jsonData)
...
+                
                 // Create a new compilation unit
                 CompilationUnitSyntax compilationUnit = SyntaxFactory.CompilationUnit()
                     .AddUsings(
                         SyntaxFactory.UsingDirective(SyntaxFactory.ParseName("System")),
                         SyntaxFactory.UsingDirective(SyntaxFactory.ParseName("System.Collections.Generic")),
                         SyntaxFactory.UsingDirective(SyntaxFactory.ParseName("RegressionGames")),
                         SyntaxFactory.UsingDirective(SyntaxFactory.ParseName("RegressionGames.RGBotConfigs")),
+                        SyntaxFactory.UsingDirective(SyntaxFactory.ParseName("RegressionGames.StateActionTypes")),
+                        SyntaxFactory.UsingDirective(SyntaxFactory.ParseName(botAction.Namespace)),
```
vontell, line 31: Does this fix the namespace issue I was seeing?

vontell (changes requested):
Looks good! Two requests:

* Is there a PR / docs somewhere so others can use this?
* Is there an example PR that shows what this would look like?

Also, does this have any breaking changes for other usages of the SDK so far in our own samples?

Editor/Scripts/CodeGenerators/GenerateRGActionClasses.cs
vontell: Resolved because I see you answer this in the PR description.

rcornwall-reg:
Does the dev need to know if the actions they're exposing are going to be called by local bots or remote bots? I'm assuming they either a
```
RGActionRequest_PerformAttack
``` 
or a 
```
RGAction_PerformAttack
```
component back to the game object, correct? Or do they always add both?

RG-nAmKcAz:
@rcornwall-reg `RGActionRequest_PerformAttack` doesn't have to be added to anything, it is used as the request object for queueing an action when writing a bot

vontell (approved):

rcornwall-reg (approved):

abeizer (approved):

PR #57, opened 2023-10-16
[REG-1303] non-nullable, always synced RGEntity position and rotation

addisonbgross:
## Summary

- RGEntities now have mandatory, non-nullable `positions` and `rotations`. They will always be synced
- Added a check to ensure that skills aren't used on UI components. They are found the same as enemies or allies in BossRoom, but lack a `position`

## Validation

To test this I created a remote bot that logs the position of all player characters in a game of BossRoom. I tested the game using the remote bot and a local bot. In the remote bot logs we can see the Vector3 position of the player characters update:

https://www.loom.com/share/7c6e22fbc0b94af6b46965b572b0a4be

## Remote Bot Logging Code (If you are curious)

```
const state = rg.getState().gameState;
const res = Object.keys(state)
    .filter(key => state[key].isPlayer)
    .map(key => `${state[key].position.x}, ${state[key].position.y}, ${state[key].position.z}`);
```

### DIFF
--- a/Editor/Scripts/RGTickInfoActionManager.cs
+++ b/Editor/Scripts/RGTickInfoActionManager.cs
@@ -50,7 +50,7 @@ namespace RegressionGames.Editor
                         var ti = replayData.tickInfo[i];
                         if (ti == null || ti.state.position == null) break;
 
-                        linePoints.Push(ti.state.position.Value);
+                        linePoints.Push(ti.state.position);
                     }
 
                     if (linePoints.Count > 0) return linePoints.ToArray();
--- a/Editor/Scripts/RGTickInfoActionManager.cs
+++ b/Editor/Scripts/RGTickInfoActionManager.cs
@@ -48,7 +48,7 @@ namespace RegressionGames.Editor
                     for (var i = tickNumber - 1; i >= 0; i--)
                     {
                         var ti = replayData.tickInfo[i];
-                        if (ti == null || ti.state.position == null) break;
+                        if (ti == null) break;
 
                         linePoints.Push(ti.state.position);
                     }
--- a/Runtime/Scripts/RGBotConfigs/RGState.cs
+++ b/Runtime/Scripts/RGBotConfigs/RGState.cs
@@ -45,7 +45,7 @@ namespace RegressionGames.RGBotConfigs
                 ["type"] = rgEntity.objectType,
                 ["isPlayer"] = rgEntity.isPlayer,
                 ["isRuntimeObject"] = rgEntity.isRuntimeObject,
-				["state"] = theTransform.position;
+				["position"] = theTransform.position;
 				["rotation"] = theTransform.rotation;
             };
 
--- a/Runtime/Scripts/RGBotConfigs/RGState.cs
+++ b/Runtime/Scripts/RGBotConfigs/RGState.cs
@@ -45,8 +45,8 @@ namespace RegressionGames.RGBotConfigs
                 ["type"] = rgEntity.objectType,
                 ["isPlayer"] = rgEntity.isPlayer,
                 ["isRuntimeObject"] = rgEntity.isRuntimeObject,
-				["position"] = theTransform.position;
-				["rotation"] = theTransform.rotation;
+				["position"] = theTransform.position,
+				["rotation"] = theTransform.rotation,
             };
 
             var dict = GetState();
--- a/Runtime/Scripts/RGBotConfigs/RGState.cs
+++ b/Runtime/Scripts/RGBotConfigs/RGState.cs
@@ -45,10 +45,9 @@ namespace RegressionGames.RGBotConfigs
                 ["type"] = rgEntity.objectType,
                 ["isPlayer"] = rgEntity.isPlayer,
                 ["isRuntimeObject"] = rgEntity.isRuntimeObject,
+				["state"] = theTransform.position;
+				["rotation"] = theTransform.rotation;
             };
-
-            if (rgEntity.syncPosition) state["position"] = theTransform.position;
-            if (rgEntity.syncRotation) state["rotation"] = theTransform.rotation;
 
             var dict = GetState();
             foreach (var entry in dict)
--- a/Runtime/Scripts/RGBotLocalRuntime/RG.cs
+++ b/Runtime/Scripts/RGBotLocalRuntime/RG.cs
@@ -61,12 +61,15 @@ namespace RegressionGames.RGBotLocalRuntime
         /**
          * <summary>Used to find the closest Entity to the given position.</summary>
          * <param name="objectType">{string | null} Search for entities of a specific type</param>
-         * <param name="position">{Vector3} Position to search from</param>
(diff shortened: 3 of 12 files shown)


Runtime/Scripts/RGBotConfigs/RGState.cs
```diff
@@ -45,11 +45,10 @@ public RGStateEntity GetGameObjectState()
                 ["type"] = rgEntity.objectType,
                 ["isPlayer"] = rgEntity.isPlayer,
                 ["isRuntimeObject"] = rgEntity.isRuntimeObject,
+				["position"] = theTransform.position,
+				["rotation"] = theTransform.rotation,
```
rcornwall-reg, line 49: Formatting issue?

rcornwall-reg (changes requested):

rcornwall-reg (approved):

Runtime/Scripts/RGBotLocalRuntime/RG.cs
```diff
@@ -61,15 +61,12 @@ public List<RGStateEntity> GetMyPlayers()
...
-         *     {Vector3 | null} Position to search from.  If not passed, attempts to use the client's bot
-         *     position in index 0.
-         * </param>
+         * <param name="position">{Vector3} Position to search from</param>
          * <param name="filterFunction">{Func&lt;RGStateEntity, bool&gt; | null} Function to filter entities.</param>
          * <returns>{RGStateEntity} The closest Entity matching the search criteria, or null if none match.</returns>
          */
         [CanBeNull]
-        public RGStateEntity FindNearestEntity(string objectType = null, Vector3? position = null,
+        public RGStateEntity FindNearestEntity(Vector3 position, string objectType = null,
```
abeizer, line 69: This should still allow a null position. This method is a helper for users to find entities from their botcode based on their distance from the specified position. null will default to searching from the bot's current position, but I can also search for the closest entity to some other position - whether that's a GameObject's or some arbitrary waypoint.

Runtime/Scripts/RGBotLocalRuntime/SampleBot/RGSampleBot.cs
```diff
@@ -38,17 +38,21 @@ public override void ProcessTick(RG rgObject)
                 {
                     var target = entities[new Random().Next(entities.Count)];
 
-                    var targetPosition = target.position ?? Vector3.zero;
-                    //TODO (REG-1302): If Actions were strongly typed we wouldn't need to build this weird map...
-                    var action = new RGActionRequest("PerformSkill", new Dictionary<string, object>
+                    // ensure target is not UI component
+                    if (target.ContainsKey("position"))
```
abeizer, line 42: I'm a bit confused by this, won't position always be defined since UI components have a rect transform?

abeizer (changes requested):

Runtime/Scripts/RGBotLocalRuntime/SampleBot/RGSampleBot.cs
RG-nAmKcAz: i agree with you abby, I would expect this check to be un-necessary

SampleBots/BossRoom/MenuAbilityBot_1/BotEntryPoint.cs
```diff
@@ -160,16 +160,20 @@ public override void ProcessTick(RG rgObject)
                             {
                                 var target = entities[new Random().Next(entities.Count)];
 
-                                var targetPosition = target.position ?? Vector3.zero;
-                                var action = new RGActionRequest("PerformSkill", new Dictionary<string, object>
+                                // ensure target is not UI component
+                                if (target.ContainsKey("position"))
```
RG-nAmKcAz, line 164: tagging them all, even ui components should have position

RG-nAmKcAz (changes requested):

RG-nAmKcAz:
@addisonbgross @abeizer 

RGButtonState doesn't inherit from `RGState` anymore after the latest round of `RGEntity` work...

We 'can' put the position and rotation of buttons into the state of `RGButtonState` if we want to but I don't think it is necessary.

That said, I still don't think we need the checks for `position` in these bots when we are performing actions.  The entities we are searching for are known ahead of time to be non buttons in those cases.

Runtime/Scripts/RGBotConfigs/RGState.cs
vontell: @addisonbgross @rcornwall-reg this comment is marked as resolved but I still see the issue

Runtime/Scripts/RGBotLocalRuntime/SampleBot/RGSampleBot.cs
```diff
@@ -37,18 +37,18 @@ public override void ProcessTick(RG rgObject)
...
-
-                    var targetPosition = target.position ?? Vector3.zero;
+                    
+                    var targetPosition = (Vector3)target["position"];
                     //TODO (REG-1302): If Actions were strongly typed we wouldn't need to build this weird map...
-                    var action = new RGActionRequest("PerformSkill", new Dictionary<string, object>
+                    var action = new RGActionRequest("PerformSkill", new Dictionary<string, object>()
                     {
-                        { "skillId", new Random().Next(2) },
+                        { "skillId",  new Random().Next(2)},
```
vontell, line 45: random new space added?

vontell (approved):
Glad to see this change! The null position was kinda annoying in my code. Approving assuming others' requests are handled.

Editor/Scripts/RGTickInfoActionManager.cs
```diff
@@ -50,7 +50,7 @@ public Vector3[] GetPathForEntityId(int tickNumber, long entityId)
                         var ti = replayData.tickInfo[i];
                         if (ti == null || ti.state.position == null) break;
```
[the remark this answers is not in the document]
addisonbgross: Good eye, removed the check for the position

addisonbgross:
@RG-nAmKcAz @abeizer I removed all of the checks in the bots I had for UI components.

I was getting errors that an entity's position did not exist, and have now realized that its the bots in the RGBossRoom repo that cause this issue. If I don't check for the position key we get constant errors. I'm going to dig into this right now, but your other comments should be addressed

Runtime/Scripts/RGBotLocalRuntime/RG.cs
addisonbgross: I have changed the function back. There are just a couple of adjustments to handle the always present position vectors

RG-nAmKcAz:
@addisonbgross Re the position checks and `RGStateEntity`.  We should avoid situations where user bot code can lead to exceptions in our SDK as much as possible.  Short term fix, I saw we add position/rotation to RGButtonState until we evaluate a better long term answer.

RG-nAmKcAz (approved):

abeizer (approved):

PR #60, opened 2023-10-17
[REG-1273] Harden network connectivity and re-connects

RG-nAmKcAz:
Not Perfect, but should reconnect remote bots automatically in most scenarios.

- Thread safety / concurrency fixes around network connection management
  - Resolves an issue where sometimes we wouldn't re-connect as necessary due to a concurrency issue setting/reading a variable
- Changes to better detect most socket state failures before read and write
- Added socket read/write timeouts to make failures detect properly
- Fixed socket read code to address issues handling transient errors / empty buffer during read attempts

### DIFF
--- a/Runtime/Scripts/RGBotSpawnManager.cs
+++ b/Runtime/Scripts/RGBotSpawnManager.cs
@@ -280,7 +280,15 @@ namespace RegressionGames
             {
                 // First, allow the developer to configure any frontend/backend for a player about to spawn,
                 // such as character selection.
-                SeatBot(botToSpawn);
+                try
+                {
+                    SeatBot(botToSpawn);
+                }
+                catch (Exception ex)
+                {
+                    RGDebug.LogError($"Error seating bot - {ex}");
+                }
+
                 RGBotServerListener rgBotServerListener = RGBotServerListener.GetInstance();
                 if (rgBotServerListener != null)
                 {
--- a/Runtime/Scripts/RGBotSpawnManager.cs
+++ b/Runtime/Scripts/RGBotSpawnManager.cs
@@ -280,14 +280,8 @@ namespace RegressionGames
             {
                 // First, allow the developer to configure any frontend/backend for a player about to spawn,
                 // such as character selection.
-                try
-                {
-                    SeatBot(botToSpawn);
-                }
-                catch (Exception ex)
-                {
-                    RGDebug.LogError($"Error seating bot - {ex}");
-                }
+
+                SeatBot(botToSpawn);
 
                 RGBotServerListener rgBotServerListener = RGBotServerListener.GetInstance();
                 if (rgBotServerListener != null)
--- a/Runtime/Scripts/Types/RGClientConnection_Remote.cs
+++ b/Runtime/Scripts/Types/RGClientConnection_Remote.cs
@@ -1,21 +1,26 @@
 using System;
 using System.Buffers.Binary;
+using System.Collections.Generic;
 using System.Net;
 using System.Net.Sockets;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
-using System.Timers;
+using Codice.Client.Common;
 using JetBrains.Annotations;
 using Newtonsoft.Json;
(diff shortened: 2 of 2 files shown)


Runtime/Scripts/RGBotSpawnManager.cs
```diff
@@ -280,7 +280,15 @@ protected internal void CallSeatBot(BotInformation botToSpawn)
...
                 // First, allow the developer to configure any frontend/backend for a player about to spawn,
                 // such as character selection.
-                SeatBot(botToSpawn);
+                try
+                {
+                    SeatBot(botToSpawn);
+                }
+                catch (Exception ex)
+                {
+                    RGDebug.LogError($"Error seating bot - {ex}");
```
vontell, line 289: Should we also just return here or raise an exception? Not sure if we want to continue getting the bot ready if the seating failed.

vontell (changes requested):
Just one question from me in the code. Also, should we add a UAT, or I'm assuming this will be covered by existing UAT as well?

Runtime/Scripts/RGBotSpawnManager.cs
RG-nAmKcAz: removed

vontell (approved):

Runtime/Scripts/Types/RGClientConnection_Remote.cs
```diff
@@ -114,113 +143,116 @@ public override bool Connected()
...
+            {
+                lock (this)
+                {
+                    if (!_connecting && !Connected())
+                    {
+                        _connecting = true;
+                        shouldIConnect = true;
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
RG-nAmKcAz: I've never used Lazy for a primitive value before. In c# primitives are atomic for a single operation.  There are still limits to cross thread visibility, especially in multi processor systems, but as long as you keep those conditions in mind you can avoid the overhead of Lazy's lock for everything pattern.

I avoided using lazy here as `ExecutionAndPublication` always locks when checking its value and the intent of the double checked locking pattern is to avoid locking when not necessary.  As I understand it, it would be roughly equivalent to Lazy's `ExecutionAndPublication` from a locking perspective to just lock and then check instead of using double checked locking; not exactly the same cpu cost, but logically similar.

In this particular case you could argue that double checked locking is only truly safe for preventing 2 simultaneous connection attempts, but may possibly miss `_connecting` transitioning to false as set by another thread and not connect in this method call when it technically could.  This statement is true... however, the system will still see on the next pass (RGBotServerListener.Update) that we aren't connected and will attempt the connection again.  This outer code already exists to detect random disconnect scenarios and is not specific to interactions with this locking code.

Long story short, I updated the code to use a Semaphore here instead of a boolean.  It doesn't change anything from a functionality perspective as described in the prior sentences, but it may be a bit more clear for others reading the code.

Runtime/Scripts/Types/RGClientConnection_Remote.cs
```diff
@@ -61,13 +66,40 @@ private bool SendSocketMessage(string type, string data)
...
+                {
+                    var vt = _client.GetStream().WriteAsync(finalBuffer);
+                    vt.ConfigureAwait(false).GetAwaiter().OnCompleted(() =>
+                    {
+                        // see if the socket is no longer writeable
+                        if (!vt.IsCompletedSuccessfully)
+                        {
+                            HandleDeadWriteState();
+                        }
+                    });
```
addisonbgross, line 80: is there a need to implement a retry policy if this write fails?

RG-nAmKcAz: updates are not guaranteed to get there.  If the connection is down, all the states between 'now' and 'future now' are not forwarded to the client as they should have no need to process them anyway

in fact.. on the client side if it is processing slowly and we get multiple new states before we finish processing old states, we also explicitly trash the states in between

rcornwall-reg (approved):

PR #61, opened 2023-10-17
[REG-1336] allow multiple state scripts per GameObject

abeizer:
* When a GameObject has multiple scripts extending RGState, squash them into one object for tickinfo
* Inherit `isPlayer` and `isRuntimeObject` from GameObject's `RGEntity` instead of letting state scripts override these values

Tested this against my platformer project + BossRoom local and remote bots

### DIFF
--- a/Runtime/Scripts/RGBotServerListener.cs
+++ b/Runtime/Scripts/RGBotServerListener.cs
@@ -498,6 +498,8 @@ namespace RegressionGames
             }
         }
 
+        
+        
         /**
          * Gets the entire game state by searching for all RGState game objects and gather their
          * states.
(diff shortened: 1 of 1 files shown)


rcornwall-reg (approved):

Runtime/Scripts/RGBotServerListener.cs
```diff
@@ -498,6 +498,8 @@ private void FixedUpdate()
             }
         }
 
+        
+
```
vontell, line 502: Random new lines added?

Runtime/Scripts/RGBotServerListener.cs
```diff
@@ -558,9 +558,20 @@ private void FixedUpdate()
                         }
                     }
                 }
-                totalState[state["id"].ToString()] = state;
+                
+                var key = gameObjectState.id.ToString();
+                if (fullGameState.TryGetValue(key, out var combinedGameObjectState))
+                {
+                    // if GameObject has multiple state scripts attached to it, then combine fields into one RGStateEntity
+                    gameObjectState.ToList().ForEach(x => combinedGameObjectState[x.Key] = x.Value);
```
vontell, line 566: Do you have an example of what this ends up looking like? From my reading, it seems like a random agent key/id is selected, and under that the other ids are nested? I am thinking I maybe be misunderstanding...

vontell (approved):

Runtime/Scripts/RGBotServerListener.cs
abeizer: for each component with RGState...
  * if entity hasn't been added to the full state object yet then add all state info
  * else append state info to existing entity in the full state object. Any key conflicts will be overwritten.

Ex. If I have one state script that defines `isIdle` and another state script that defines `inventory` then we end up with what we've been expecting:
```
<entity id>: {
  "isPlayer": true,
  "isRuntimeObject": true,
  "position": ... , 
  "rotation": ... ,
  "isIdle": true,
  "inventory": [...] ,
  // other state stuff
}
```

vontell: Got it, thank you!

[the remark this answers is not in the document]
abeizer: Added logs for fields that we don't expect to be duplicated (things like `position`, `rotation`, `isPlayer`, etc. default to the values on the GameObject's `RGEntity` so we won't log warnings for those)

Runtime/Scripts/RGBotServerListener.cs
```diff
@@ -498,18 +498,25 @@ private void FixedUpdate()
...
          * Gets the entire game state by searching for all RGState game objects and gather their
          * states.
          */
         private Dictionary<string, RGStateEntity> GetGameState()
         {
             var overlayAgent = this.gameObject.GetComponent<RGEntity>();
-
             var statefulObjects = FindObjectsOfType<MonoBehaviour>(true).OfType<IRGState>();
             var fullGameState = new Dictionary<string, RGStateEntity>();
+            var duplicatedFields = new List<string>()
```
RG-nAmKcAz, line 510: Let's define this outside the method so it doesn't get re-allocated on every call

RG-nAmKcAz (changes requested):

Runtime/Scripts/RGBotServerListener.cs
abeizer: Done 👍

RG-nAmKcAz (approved):

PR #62, opened 2023-10-17
[REG-1333] - Add namespace to generated state files

rcornwall-reg:
Fixes compile errors after `Regression Games -> Generate Scripts` 

**To Test**
Add `[RGAction]` and `[RGState]` attributes to a class with a custom namespace

```
using UnityEngine;
using RegressionGames;

namespace TestNamespace
{
  public class Test : MonoBehaviour
  {
     [RGState]
     public int testState = 5;
     
     [RGAction]
     public void TestAction()
     {
       Debug.Log("Test Action");
     }
     
     private void Start(){}
  }
}
```

Generate scripts. `Regression Games->Generate Scripts`. Generated files should include `using TestNamespace;`

### DIFF
--- a/Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
+++ b/Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
@@ -171,11 +171,11 @@ namespace RegressionGames
 
                 foreach (var classDeclaration in classDeclarations)
                 {
-                    string className = classDeclaration.Identifier.ValueText;
-                    List<RGStateInfo> stateList = new List<RGStateInfo>();
-
                     string nameSpace = classDeclaration.Ancestors().OfType<NamespaceDeclarationSyntax>().FirstOrDefault()?.Name.ToString();
 
+                    string className = classDeclaration.Identifier.ValueText;
+                    List<RGStateInfo> stateList = new List<RGStateInfo>();
+                    
                     var membersWithRGState = classDeclaration.Members
                         .Where(m => m.AttributeLists.Any(a => a.Attributes.Any(attr => attr.Name.ToString() == "RGState")));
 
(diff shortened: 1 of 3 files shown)


vontell (approved):
LGTM!

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
```diff
@@ -164,6 +164,8 @@ private static string SearchForBotStates()
                     string className = classDeclaration.Identifier.ValueText;
                     List<RGStateInfo> stateList = new List<RGStateInfo>();
 
+                    string nameSpace = classDeclaration.Ancestors().OfType<NamespaceDeclarationSyntax>().FirstOrDefault()?.Name.ToString();
```
RG-nAmKcAz, line 167: can you update this higher in the file in the other location so that the code pattern matches

~line 75

RG-nAmKcAz (changes requested):

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
rcornwall-reg: Updated to be the first line of the for loop. Is that what you meant? If not let me know 👍

RG-nAmKcAz: no, I mean on line ~75 where this similar operation occurs, can you update that other code location to be consistent with this one

RG-nAmKcAz (approved):

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
rcornwall-reg: I updated the declaration to match. Let me know if that's more like what your saying

PR #64, opened 2023-10-18
[REG-1335] Debug features for lines, indicators, and billboard text

vontell:
As I was testing my bot, I found that console logs were not enough to understand what my bot was doing, and it was hard to debug issues. This PR adds three core functionalities:

* Ability to draw lines from the agent to a position
* Ability to draw an indicator (i.e. a sphere) at a position
* Ability to draw a debug text billboard above the agent

Video demo of my bot showing it's distance and angle to the next waypoint, highlighting nearby waypoints, and drawing a line to the next waypoint:
https://www.loom.com/share/33cc6f2712bd435598e9679a3f7a4a3c?sid=ceec6c65-d309-4bd6-8ab2-8f6b3bd8811c

Here is a video of the sample bot with these changes as well:
https://www.loom.com/share/2445c4039ff44481a38e50ee8f4d4ab4?sid=7cc16c01-2b67-4a0a-b79a-f8205095b6f2

Some limitations:
* The billboard feature is basic, as in there is no control on the font and sizing, so it can only fit limited text. However, you can control the offset above the agent. I am not sure if it makes sense to handle this resizing issue now.

Interesting aspects of this:
* Making canvases and text programmatically was so difficult... so I made a prefab. @RG-nAmKcAz I used a loading approach here similar to how we load certs, but I actually am not sure if that will work in builds?

**Requests from reviewers**
Right now - give a first initial pass and let me know if this is useful or sounds good. If so, I will start my documentation PR for the feature! In the meantime, I did have proper C# comments.
After merging - Try it out extensively yourself, provide feedback, and add more functionality to it!
Desired extensions:
* More control over the text billboard

**How I tested this:**
I added this to my own sample, and tested each function. You can see this operation in the video above. I also tested this in bossroom.

**How you can test this:**
Please do this! It's really easy - just checkout this branch in your RGUnityBots folder if your project uses local bots, (or alternatively just add a package via git url using `https://github.com/Regression-Games/RGUnityBots.git#aaron/reg-1335`). Then, inside your bot code, just add the following (I'm guessing you have at least one position you can reference!):

**NOTE: Gizmos must be turned on, you can find this while the game is playing in the top right.**

```csharp
// Examples from my scene
RGGizmos.CreateLine(thisAgent.id, nextWaypoint.position, Color.green, "MyLine");
RGGizmos.CreateSphere(nextWaypoint.position,  Color.red, 0.3f, true, "MySphere");
RGGizmos.CreateText(thisAgent.id, $"Distance: {distance}\nAngle: {angle}");
```


Runtime/Scripts/DebugUtils/BillboardText.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            SetText(_content);
+            SetYOffset(_yOffset);
+        }
+    }
+
+    // Update is called once per frame
+    void LateUpdate()
+    {
+        transform.LookAt(_mainCamera.transform);
```
vontell, line 27: This is slightly wrong (you can see the effect more prominently in the boss room video in the PR description - my billboard tilts too much). Any advice?

Runtime/Scripts/DebugUtils/BillboardText.cs
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    public void SetText(string content)
+    {
+        // In some cases, the content is set when the object is not awake. Instead, wait until the object is
+        // awake
+        _content = content;
+        if (_text == null) return;
+        _text.text = content;
```
vontell, line 36: I hate what I did here, going to simplify this I think.

Runtime/Scripts/RGBotConfigs/DefaultActions/DrawIndicator.cs
```diff
@@ -0,0 +1,67 @@
...
+                _indicators.Remove(indicatorName);
+            }
+            else
+            {
+                var targetPosition = (Vector3) input["position"];
+                var color = (Color) input["color"];
+                var size = (float) input["size"];
+                var indicatorParams = Tuple.Create(targetPosition, color, size);
+                _indicators[indicatorName] = indicatorParams;
+            }
```
vontell, line 53: You can see hear that my actions are "overloaded". Since my actions keep the references to the indicators (and also line and text in the other actions), I decided to back the remove operation into this as well. I could have put all indicator/line/text logic into somewhere on RGEntity, but that seemed much worse.

Runtime/Scripts/DebugUtils/BillboardText.cs
addisonbgross: maybe if you made the forward vector of billboard the camera's forward vector * -1?

Runtime/Scripts/RGBotConfigs/DefaultActions/DrawLineToAction.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+
+            var lineName = (string) input["name"];
+            if ((bool) input.GetValueOrDefault("remove", false))
+            {
+                _lines.Remove(lineName);
+            }
+            else
+            {
+                var targetPosition = (Vector3) input["position"];
```
RG-nAmKcAz, line 47: This seems error prone.  We are getting the target position in `StartAction` and then continuing to render this indicator until they tell us to remove it, but the line's start and end point aren't updating each frame.

We definitely should have this be able to accept and `entityId` as the target so that the line can keep moving both its start and end from frame to frame

Runtime/Scripts/RGBotLocalRuntime/RG.cs
```diff
@@ -235,7 +235,136 @@ internal List<RGValidationResult> FlushValidations()
...
+         * used to update and remove this line later.
+         * <remarks>Gizmos must be enabled to see this line.</remarks>
+         * <param name="position">The position in the world to draw the line toward</param>
+         * <param name="color">The color of the line</param>
+         * <param name="name">A name to reference this line by for later operations</param>
+         * <example><code>rgObject.DrawLineTo(target.position, Color.red, "TargetEnemy");</code></example>
+         * <seealso cref="RemoveLineTo"/>
+         * <seealso cref="RemoveAllLinesTo"/>
+         */
+        public void DrawLineTo(Vector3 position, Color color, string name)
```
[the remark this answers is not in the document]
RG-nAmKcAz: After reading through all of this, I would now vote that we name these something like `Create` and `Destroy` as we are creating long living indicators across many frames, basically for infinite time until they remove them.

Runtime/Scripts/RGBotConfigs/DefaultActions/DrawIndicator.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        }
+
+        private void OnDrawGizmos()
+        {
+            foreach (var indicatorParams in _indicators.Values)
+            {
+                Gizmos.color = indicatorParams.Item2;
+                Gizmos.DrawSphere(indicatorParams.Item1, indicatorParams.Item3);
```
RG-nAmKcAz, line 62: we should use `WireSphere` as `Sphere` is a solid fill object that blocks visibility to scene elements under it

Runtime/Scripts/RGBotConfigs/DefaultActions/DrawIndicator.cs
```diff
@@ -0,0 +1,66 @@
...
+     *    - removeAll (bool): Set to true
+     */
+    public class DrawIndicator: RGAction
+    {
+
+        private Dictionary<string, Tuple<Vector3, Color, float>> _indicators = new ();
+
+        public override string GetActionName()
+        {
+            return "DrawIndicator";
```
RG-nAmKcAz, line 29: Let's name this what it is `DrawSphere` or `DrawWireSphere` as there are many useful indicator types in `Gizmos`

Runtime/Scripts/RGBotConfigs/RGEntity.cs
```diff
@@ -38,8 +38,23 @@ void Start()
 
             if (actionMap.Count > 0)
             {
-                RGDebug.LogDebug($"Entity registered with {actionMap.Count} actions");
+
+                // If this is an entity with other actions, also give it default actions
+                var defaultActions = new RGAction[]
+                {
+                    gameObject.AddComponent<DrawLineToAction>(),
```
RG-nAmKcAz, line 45: instead of putting these debug actions on every entity, i would restructure them as a special case where the from entity and to entity/position/etc are part of the payload.

Adding these to all those game objects seems weird when it is debug code.

Let's separate debug from actual code and make debug be debug.

RG-nAmKcAz: The gameObject with the BotServerListener/etc on it can have a debugactionhandler on it that handles rendering of all debug actions.

RG-nAmKcAz (changes requested):
Let's get the .DS_Store files out of the project and gitignored.

All of these scripts should be in something like a `Runtime/Scripts/DebugUtils` sub folder.  We should isolate/hide these actions from being used directly as much as possible.

Runtime/Scripts/RGBotLocalRuntime/RG.cs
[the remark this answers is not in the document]
vontell: I was intentionally only supporting client-owned Entity to position line drawing, and was hoping someone would extend it later to add another function that drew a line between to random points or entities, but I suppose I can update that now!

Runtime/Scripts/RGBotLocalRuntime/RG.cs
```diff
@@ -235,7 +235,136 @@ internal List<RGValidationResult> FlushValidations()
...
+        public void RemoveAllLinesTo()
+        {
+            PerformAction(new RGActionRequest("DrawLineTo", new Dictionary<string, object>()
+            {
+                { "removeAll", true}
+            }));
+        }
+
+        /**
+         * Draws a snippet of text above this agent. Optionally change the offset which defines how far
```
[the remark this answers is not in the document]
vontell: Yup my bad - I'm curious about going back to why we made this decision... in my mind they are two different things (i.e. an entity with actions and controlled by a client is different from just an entity sitting in the scene providing state). Thoughts? Maybe a discussion for another time, but I do think this could confuse users.

Runtime/Scripts/RGBotLocalRuntime/RG.cs
```diff
@@ -235,7 +235,136 @@ internal List<RGValidationResult> FlushValidations()
...
+         * above the agent the text is drawn.
+         * <param name="content">The text to place above the agent</param>
+         * <param name="yOffset">The distance from this agent's origin to this text's Y position. Defaults to 2.</param>
+         * <example><code>rgObject.DrawText($"Ability {chosenAbility} on enemy {enemyId}", 3.0f);</code></example>
+         * <seealso cref="RemoveText"/>
+         */
+        public void DrawText(string content, float yOffset = 2.0f)
```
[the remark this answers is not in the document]
RG-nAmKcAz: when you get to the comment about managing this in a central location, not on each entity, it may make more sense

Runtime/Scripts/RGBotConfigs/DefaultActions/DrawLineToAction.cs
vontell: This is correct and was intentional based on my current use case. Just a heads up though, the line's start **is** updated every frame, since it's based on the agent's origin. I was thinking that a different function or parameter would be provided if you wanted to follow a certain transform.

Runtime/Scripts/RGBotConfigs/DefaultActions/DrawIndicator.cs
vontell: In my scene, the regular sphere was much better for me - maybe I'll add a parameter / option for that?

Runtime/Scripts/DebugUtils/BillboardText.cs
addisonbgross: This problem is stuck in my head. I think this will work:

```
transform.LookAt(_mainCamera.transform);
transform.up = _mainCamera.up;
```

I think setting the `up` like that will keep your billboard looking correct regardless of the camera's rotation

Runtime/Scripts/DebugUtils/RGGizmosWrapper.cs
```diff
@@ -0,0 +1,383 @@
...
+        private readonly Dictionary<int, Tuple<string, float>> _billboardsToDraw = new();
+        private readonly Dictionary<int, GameObject> _drawnBillboards = new();
+        
+        // Billboard text objects
+        private readonly GameObject _billboardAsset;
+
+        public RGGizmos()
+        {
+            var path = "Packages/gg.regression.unity.bots/Runtime/Prefabs/AgentBillboardText.prefab";
+            _billboardAsset = AssetDatabase.LoadAssetAtPath<GameObject>(path);
```
vontell, line 33: @rcornwall-reg you mention that there are better ways of doing this... after much research and testing, I still couldn't figure it out. Have any ideas/references/want to hop on a call today to help walk me through it?

Runtime/Scripts/DebugUtils/BillboardText.cs
```diff
@@ -0,0 +1,47 @@
...
+            SetYOffset(_yOffset);
+        }
+    }
+    
+    void LateUpdate()
+    {
+        Vector3 newRotation = _mainCamera.transform.eulerAngles;
+        newRotation.x = 0;
+        newRotation.z = 0;
+        transform.eulerAngles = newRotation;
```
vontell, line 28: This is still better, but wrong... I liked how they have it in BossRoom, but it seems like what they do is get the position of the character on the screen and then draw the UI onto a canvas that covers the entire screen, rather than actually have a canvas on the player. Maybe this is good enough for now, but I notice sometimes it can be janky.

Runtime/Scripts/RGBotLocalRuntime/RG.cs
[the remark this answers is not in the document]
vontell: Fixed/new code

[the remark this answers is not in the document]
vontell: Fixed/new code

Runtime/Scripts/RGBotLocalRuntime/RG.cs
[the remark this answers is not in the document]
vontell: Fixed/new code

Runtime/Scripts/RGBotConfigs/DefaultActions/DrawLineToAction.cs
vontell: Fixed/new code

Runtime/Scripts/RGBotConfigs/DefaultActions/DrawIndicator.cs
vontell: Fixed/new code

Runtime/Scripts/RGBotConfigs/DefaultActions/DrawIndicator.cs
vontell: Fixed/new code

vontell:
This is re-open for review!

analogrelay (commented):
Pretty much just some C# style and idiom feedback. I do have a question about some general `Exception` catching we're doing, which I'm usually pretty wary of.

Runtime/Scripts/RGBotLocalRuntime/RGUserBot.cs
```diff
@@ -13,13 +15,24 @@ public abstract class RGUserBot : ScriptableObject
 
         private long _botId;
         private string _botName;
+        private RGGizmos _rgGizmos;
+
+        /**
+         * A set of debug utilities for drawing gizmos and text on top of entities in your scene.
+         */
```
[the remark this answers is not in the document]
RG-nAmKcAz: sidenote: /** block style is still allowed as long as you use the xml tags.

Runtime/Scripts/DebugUtils/RGGizmosWrapper.cs
```diff
@@ -0,0 +1,383 @@
...
+{
+    
+    /**
+     * A set of debug utilities for drawing gizmos and text on top of entities in your scene.
+     */
+    // ReSharper disable once InconsistentNaming
+    public class RGGizmos
+    {
+        
+        private readonly Dictionary<string, Tuple<int, Vector3, Color>> _linesFromEntityToPosition = new();
```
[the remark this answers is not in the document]
RG-nAmKcAz: good to learn 👍

Runtime/Scripts/RGBotLocalRuntime/RGUserBot.cs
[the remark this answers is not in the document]
analogrelay: Oh man, TIL. I've never seen the block stand before but it is indeed valid. I believe you're still required to wrap the summary content in a `<summary>` tag for all documentation tools to pick it up.

If you feel silly wrapping summary content in an XML tag, I'm 1000% with you on that 😓. It's a pattern developed way back in C# 1 (circa 2002) when XML was Cool and Hip.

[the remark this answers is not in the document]
vontell: Added!

Runtime/Scripts/DebugUtils/RGGizmosWrapper.cs
```diff
@@ -0,0 +1,383 @@
...
+                    }
+                
+                    // Then set the parameters
+                    var billboardText = _drawnBillboards[billboardParams.Key].GetComponent<BillboardText>();
+                    billboardText.SetText(billboardParams.Value.Item1);
+                
+                    // If an offset is given, use that for placing it above the agent
+                    billboardText.SetYOffset(billboardParams.Value.Item2);
+                }
+                catch (Exception e)
```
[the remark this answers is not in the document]
vontell: Yeah good point... got lazy here and you caught me 😬. Basically I am trying to handle the situation when an object is deleted from the scene, and so `_drawnBillboards[billboardParams.Key]` may be destroyed. I'll test this scenario and see what the exception is!

Runtime/Scripts/DebugUtils/RGGizmosWrapper.cs
```diff
@@ -0,0 +1,383 @@
...
+         * RGGizmos.CreateLine(new Vector3(0, 0, 0), 1, Color.red, "myLine");
+         * </code>
+         * </example>
+         * <seealso cref="DestroyLine"/>
+         * <seealso cref="DestroyAllLines"/>
+         */
+        public void CreateLine(Vector3 startPosition, int endEntityId, Color color, string name)
+        {
+            CreateLine(endEntityId, startPosition, color, name);
+        }
```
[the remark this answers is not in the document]
vontell: Thanks! Changed because I prefer this way too

Runtime/Scripts/DebugUtils/RGGizmosWrapper.cs
```diff
@@ -0,0 +1,383 @@
...
+         * // Create a line from the entity with id 1 to the position (0, 0, 0) with the color red.
+         * RGGizmos.CreateLine(1, new Vector3(0, 0, 0), Color.red, "myLine");
+         * </code>
+         * </example>
+         * <seealso cref="DestroyLine"/>
+         * <seealso cref="DestroyAllLines"/>
+         */
+        public void CreateLine(int startEntityId, Vector3 endPosition, Color color, string name)
+        {
+            _linesFromEntityToPosition[name] = Tuple.Create(startEntityId, endPosition, color);
```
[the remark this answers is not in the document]
vontell: Fixed!

Runtime/Scripts/DebugUtils/RGGizmosWrapper.cs
```diff
@@ -0,0 +1,383 @@
...
+        private readonly Dictionary<string, Tuple<int, Color, float, bool>> _spheresAtEntity = new();
+        private readonly Dictionary<int, Tuple<string, float>> _billboardsToDraw = new();
+        private readonly Dictionary<int, GameObject> _drawnBillboards = new();
+        
+        // Billboard text objects
+        private readonly GameObject _billboardAsset;
+
+        public RGGizmos()
+        {
+            var path = "Packages/gg.regression.unity.bots/Runtime/Prefabs/AgentBillboardText.prefab";
```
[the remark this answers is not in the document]
vontell: Fixed!

Runtime/Scripts/DebugUtils/RGGizmosWrapper.cs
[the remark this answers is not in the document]
vontell: Ahhhhh so much better! Thanks!

vontell:
Thanks for all the comments @analogrelay! Always eager to learn all the syntactic sugar :) working on the exception comment but addressed your other points.

Runtime/Scripts/DebugUtils/BillboardText.cs
RG-nAmKcAz: have a look at `LookAtCameraInEditor` in our SDK

I apply this to the `NameTag` prefab in our tools so that the textmesh pro gameObjects floating above the characters are always facing to a point 'infinitely' behind the camera (not really infinite.. but a long long way).

There are other solutions to the problem, but you should be able to do something nearly identical here and/or move & rename & adjust that current script to work in editor and non-editor mode.

RG-nAmKcAz (approved):
Approving

Left some suggestions for how to handle the text direction issues

Runtime/Scripts/DebugUtils/RGGizmosWrapper.cs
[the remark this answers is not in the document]
vontell: Discovered the exception and updated! During this process, I also found a concurrency issue... which should be very uncommon with how this code should work, but still happened. Added ConcurrentDictionary over Dictionary.

Runtime/Scripts/DebugUtils/RGGizmosWrapper.cs
rcornwall-reg: It's not a _bad_ way of doing it, and it should work, I just think directly referencing prefabs by their file path is bad practice for Unity development, Here are some alternatives, but of course feel free to disregard

1. I would recommend the `Resource.Load` method. It doesn't require hardcoding the path to the prefab, but it does require that the prefab be located in a Resources folder, and you’d refer to it by its name.
 * Place prefab into Resources folder: `Resources/AgentBillboardText.prefab`
 * Load prefab with `_billboardAsset = Resources.Load<GameObject>("AgentBillboardText");`
 * Resources.Load will still work even if Resources folder is in a package
 
 Other than that you could:
 2. Expose the prefab as a field (either public or mark with `[SerializeField]`, although it'd require linking it in the inspector so it's not my favorite
 
 3. Including prefab as part of a 'RGGizmosSettings` scriptable object that is added to the wrapper - again probably as an inspector field. I'd recommend this method if there's going to be a lot of prefabs referenced or other settings values
 
Any of these would work and make the prefab loading a lot less prone to error

Runtime/Scripts/DebugUtils/BillboardText.cs
rcornwall-reg: Did you try the `LookAt` method previously? What's the issue with the billboarding?
Ex:

```
if (_mainCamera != null)
{
   transform.LookAt(transform.position + _mainCamera.transform.rotation * Vector3.forward,
       _mainCamera.transform.rotation * Vector3.up);
}
```

Runtime/Scripts/DebugUtils/RGGizmosWrapper.cs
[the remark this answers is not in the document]
analogrelay: Yeah, if we know there are concurrent invocations at play, using a `ConcurrentDictionary` is probably best. That will change some of the patterns though since you'll need to avoid races.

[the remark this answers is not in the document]
analogrelay: > Basically I am trying to handle the situation when an object is deleted from the scene, and so _drawnBillboards[billboardParams.Key] may be destroyed. I'll test this scenario and see what the exception is!

It'd be ideal if we could check if the billboard has been destroyed _before_ attempting to work with it. Something like:

```csharp
var billboard = _drawnBillboards[billboardParams.Key];

if(!billboard.IsDestroyed) {
    // ... work with it
}
```

That means that in _most_ cases, the exception is never thrown. Exceptions are _quite_ expensive in .NET (though to be fair, that cost probably isn't really noticable here) and it's best to avoid them entirely as much as possible. Of course, if there's concurrent code at play here, that pattern I showed above wouldn't be perfect and we'd still want the `try..catch` inside the `if(!billboard.IsDestroyed)` to cover unexpected cases, but we'd be able to avoid the Exception in _most_ cases and that's ideal.

Runtime/Scripts/DebugUtils/RGGizmosWrapper.cs
vontell: I will try the Resource approach! From my reading online... it seemed like packages actually don't support the resource folder, but if that works, that would make things easier!

Runtime/Scripts/DebugUtils/RGGizmosWrapper.cs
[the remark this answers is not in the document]
vontell: Thanks for the suggestion! Added this to my code. Btw, once you are satisfied with changes as it relates to a comment, make sure that you mark them as resolved - we have the comment writer, not the PR owner, mark conversations as resolved so they can verify that the PR owner made the requested changes properly.

Runtime/Scripts/DebugUtils/BillboardText.cs
vontell: @RG-nAmKcAz's approach worked, took it from the existing sample in our SDK! Closing this comment

Runtime/Scripts/DebugUtils/BillboardText.cs
```diff
@@ -0,0 +1,47 @@
...
+    
+    void Awake()
+    {
+        _mainCamera = Camera.main;
+        _text = GetComponentInChildren<TextMeshProUGUI>();
+        if (_content != null)
+        {
+            SetText(_content);
+            SetYOffset(_yOffset);
+        }
```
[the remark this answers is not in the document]
vontell: @rcornwall-reg I had it like this because this billboard text is instantiated on the fly and the `SetText` method gets called before the object is awake. So even though it's private and doesn't have a default, it immediately gets set.

But based on your comments, I drastically reduced the complexity of this file now - let me know if that looks good now!

Runtime/Scripts/DebugUtils/RGGizmosWrapper.cs
vontell: Thanks for this! Got it to work with Resource.Load and seems cleaner, thanks!

vontell:
@reviewers thanks for all the comments! I believe everything is addressed. I will be starting and making a PR soon for documentation in our site for this. 

NOTE: I need some help with a few more bugs though! 1) Sometimes my billboard text randomly decides to not face the camera, and 2) my text mesh gets stretched on certain entities.

RG-nAmKcAz:
@vontell 
re:`my text mesh gets stretched on certain entities` and other text camera rotation issues

This all happens because of ...
`billboardObject.transform.SetParent(entityGameObject.transform);`

It is tempting/easy to set the billBoard object as a child of the gameObject it represents.  However, that doesn't just affect positioning.  It also affects rotation, size scaling, and other inherited transform parameters.

The best way to resolve this would be to parent all of the billboard objects under some fixed position RGDebugGizmos empty game object that you spawn into the scene at 0,0,0 with the default rotation.

Your billBoard object position/rotation behavior script can have a setter method for which game object or transform it is supposed to follow.  Then on lateupdate you can adjust the position of your object to be correct relative to the position of the object it is tracking; in addition to the rotational positioning code you have now.

vontell:
Added docs page: https://github.com/Regression-Games/RegressionDocs/pull/40

vontell:
> @vontell re:`my text mesh gets stretched on certain entities` and other text camera rotation issues
> 
> This all happens because of ... `billboardObject.transform.SetParent(entityGameObject.transform);`
> 
> It is tempting/easy to set the billBoard object as a child of the gameObject it represents. However, that doesn't just affect positioning. It also affects rotation, size scaling, and other inherited transform parameters.
> 
> The best way to resolve this would be to parent all of the billboard objects under some fixed position RGDebugGizmos empty game object that you spawn into the scene at 0,0,0 with the default rotation.
> 
> Your billBoard object position/rotation behavior script can have a setter method for which game object or transform it is supposed to follow. Then on lateupdate you can adjust the position of your object to be correct relative to the position of the object it is tracking; in addition to the rotational positioning code you have now.

I'll give that a try!

Runtime/Scripts/DebugUtils/RGGizmos.cs
```diff
@@ -0,0 +1,396 @@
...
+                        var entityGameObject = RGFindUtils.Instance.FindOneByInstanceId<RGEntity>(billboardParams.Key);
+                        if (entityGameObject == null) continue;
+                        var billboardObject = Object.Instantiate(_billboardAsset, entityGameObject.transform.position,
+                            Quaternion.identity);
+                        billboardObject.transform.SetParent(entityGameObject.transform);
+                        _drawnBillboards[billboardParams.Key] = billboardObject;
+                    }
+                
+                    // Then set the parameters
+                    var billboard = _drawnBillboards[billboardParams.Key];
```
analogrelay, line 375: The "`dict.ContainsKey(key)` then get/set `dict[key]`" pattern isn't thread-safe, even when working with a `ConcurrentDictionary`. Nothing prevents another thread from adding/removing the dictionary entry between when you called `ContainsKey` and when you create/update the entry.

Instead, you should use `GetOrAdd`:

```csharp
var billboard = _drawnBillboards.GetOrAdd(billboardParams.Key, (key) => {
    // This is called if the billboard didn't exist in the dictionary.
    // Initialize it and return the value.
});
```

It's important to note that your "Add" callback _may be called multiple times for the same key_. The lock that ConcurrentDictionary uses to protect against multiple concurrent calls to GetOrAdd is _only_ taken **after** the callback has run in order to protect actually placing the new object in the dictionary. So, you may still end up calling `Object.Instantiate` multiple times for the same billboard key, if `OnDrawGizmos` is run concurrently.

Do we know how `OnDrawGizmos` could be called concurrently? It seems like we should prevent that by taking a lock of some kind. If we believe that calling `OnDrawGizmos` concurrently is incorrect, there are some fairly cheap mechanisms we can use. Even a traditional lock _should_ have low-to-no contention. Trying to draw the same Gizmos concurrently seems incorrect to me, but I'm not totally clear on the usage.

Runtime/Scripts/DebugUtils/RGGizmos.cs
```diff
@@ -0,0 +1,396 @@
...
+                    if (billboard == null || !billboard.activeInHierarchy) continue;
```
analogrelay, line 376: Here's another bit of funky C# pattern matching for you to consider. Again, a style thing, so I'm fine if it feels less clear. This could be:

```suggestion
                    if (billboard is not { activeInHierarchy: true }) continue;
```

This pattern breaks down as:

* `billboard is not ...` - Only call the `if` block if `billboard` does **not** match the specified pattern
* `{ activeInHierarchy: true }` - `billboard` must be a non-`null` object **with** the `activeInHierarchy` property set to `true`.

It does read fairly well to me. You're saying "If the billboard is not `activeInHierarchy`, then `continue`". A `null` billboard is clearly not `activeInHierarchy`.

Runtime/Scripts/DebugUtils/RGGizmos.cs
analogrelay: Disallowing concurrent calls to `OnDrawGizmos` would also solve other thread-safety issues with the other `ConcurrentDictionary`s in this method. IMO, it's still reasonable to use `ConcurrentDictionary`s in this type since I think it's _desirable_ to allow the `CreateN`/`DestroyN` methods to be called concurrently. I just also think it's reasonable to disallow concurrent Drawing.

analogrelay (commented):
More fancy C# idioms ;P. But also a question about concurrent Gizmo drawing (and a suggestion that we disallow that)

Runtime/Scripts/DebugUtils/BillboardText.cs
```diff
@@ -0,0 +1,36 @@
...
+    
+    void Awake()
+    {
+        _text = GetComponentInChildren<TextMeshProUGUI>();
+    }
+    
+    void LateUpdate()
+    {
+        Transform camTransform = Camera.main != null ? Camera.main.transform : null;
+        if (camTransform != null)
```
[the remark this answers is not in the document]
vontell: Weirdly enough, Unity specifically didn't like that first one due to some "bypassing lifetime access" issue 🤷 so going with the zesty approach! I think I actually used that somewhere else in our code once, so should be fine!

Runtime/Scripts/DebugUtils/BillboardText.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            // Set the point of focus of the object to far behind the camera
+            Vector3 lookPoint = camTransform.position +
+                                ( camTransform.forward) * 100_000;
```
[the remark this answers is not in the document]
vontell: Oops XD yes changed to the second one just for clarity

Runtime/Scripts/DebugUtils/BillboardText.cs
```diff
@@ -0,0 +1,36 @@
...
+            transform.LookAt(lookPoint);
+        }
+        else
+        {
+            Debug.Log("UOH");
```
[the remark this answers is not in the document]
vontell: UOH! I forgot to remove my typo-ed debug statement XD

Runtime/Scripts/RGBotLocalRuntime/RGUserBot.cs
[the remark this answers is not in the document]
vontell: Oops I lied the first time, now it is added.

Runtime/Scripts/DebugUtils/RGGizmos.cs
vontell: Luckily OnDrawGizmos should **not** ever be called concurrently, since it is only executed on the main thread by the Unity engine. The only concurrent part of this is that multiple threads may try to write to the dictionaries, but this particular function is only called by one thread.

So I think if I add your GetOrAdd suggestion, we should be good?

vontell: @analogrelay alright this should be addressed properly now, I think...

Runtime/Scripts/DebugUtils/RGGizmos.cs
vontell: Looks better to me as well! Added.

vontell:
> > @vontell re:`my text mesh gets stretched on certain entities` and other text camera rotation issues
> > This all happens because of ... `billboardObject.transform.SetParent(entityGameObject.transform);`
> > It is tempting/easy to set the billBoard object as a child of the gameObject it represents. However, that doesn't just affect positioning. It also affects rotation, size scaling, and other inherited transform parameters.
> > The best way to resolve this would be to parent all of the billboard objects under some fixed position RGDebugGizmos empty game object that you spawn into the scene at 0,0,0 with the default rotation.
> > Your billBoard object position/rotation behavior script can have a setter method for which game object or transform it is supposed to follow. Then on lateupdate you can adjust the position of your object to be correct relative to the position of the object it is tracking; in addition to the rotational positioning code you have now.
> 
> I'll give that a try!

@RG-nAmKcAz this worked much better, thanks!

Runtime/Scripts/DebugUtils/RGGizmos.cs
analogrelay: > So I think if I add your GetOrAdd suggestion, we should be good?

Yep, that should be fine. Especially since it looks like `_drawnBillboards` is only updated by `OnDrawGizmos`. I'd love it if we could have something in place to enforce that `OnDrawGizmos` is only run by a single thread, but I don't know a quick way to do that off-hand. This should be fine as-is.

Runtime/Scripts/DebugUtils/BillboardText.cs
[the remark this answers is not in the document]
analogrelay: Ah interesting. It looks like [Unity has some special stuff in place to handle `null` checks](https://github.com/JetBrains/resharper-unity/wiki/Possible-unintended-bypass-of-lifetime-check-of-underlying-Unity-engine-object).

Given that, I'm now not sure that the pattern matching approach is desirable either. Pattern matching explicitly disregards overloaded `==` operators, but it looks like Unity is using that to handle a liveness check on the underlying native object.

I think you should probably go back to the original logic, given that. I think even the pattern approach has the risk of us referring to a destroyed Unity object within the `if` block body.

Runtime/Scripts/DebugUtils/RGGizmos.cs
analogrelay: Ok, given the above stuff about "bypassing lifetime access checks", I think you should return to the original code path here. The `== null` comparison has special meaning in Unity and I'd prefer we not rock the boat and miss something important.

vontell: Reverted!

abeizer (approved):

rcornwall-reg (approved):
Changes look very good 👍

analogrelay (approved):
LGTM, at least from a C# style and idiom perspective. I'm less familiar with Unity stuff but that looks good to me too.

addisonbgross (approved):

PR #66, opened 2023-10-18
[REG-1340] Support nullable parameters for RGActions

abeizer:
* Fix remote bot connections
* Fix bug that prevented buttons from being added to game state
* Update code generation to handle nullable parameters for actions
  * Can accept types in the form of `Nullable<T>` and `T?`
  * Was yielding invalid deserializer methods for non-primitives
  * Generated RGAction classes expected non-null inputs
* Tested with every BossRoom bot I'm currently aware of 
  * Remote (UnityTestBot) -> abilityBot, abilityValidationBot, coopBot, menuBot, menuValidationBot, menuAndAbilityBot
  * Local -> AbilityBot, MenuAbilityBot 

See https://github.com/Regression-Games/RGBossRoom/pull/39 for more examples in addition to the one below

Example action: 
```csharp
[RGAction]
public void PerformSkill(int skillId, int? targetId, float? xPosition, float? yPosition, float? zPosition)
{
   // action code
}
```

### Before

Generated Deserializer
```csharp 
namespace RegressionGames
{
    using UnityEngine;

    public static class RGSerialization
    {
        public static int? Deserialize_int?(string paramJson)
        {
            return JsonUtility.FromJson<int?>(paramJson);
        }

        public static float? Deserialize_float?(string paramJson)
        {
            return JsonUtility.FromJson<float?>(paramJson);
        }
    }
}
```

Generated Action
```csharp
public override void StartAction(Dictionary<string, object> input)
        {
        
            if (!input.ContainsKey("skillId"))
            {
                RGDebug.LogError("No parameter 'skillId' found");
                return;
            }

            string skillIdInput = input["skillId"].ToString();
            int skillId = default;
            try
            {
                int.TryParse(skillIdInput, out skillId)
            }
            catch (Exception ex)
            {
                RGDebug.LogError("Failed to parse 'skillId'");
                RGDebug.LogError(ex.Message);
            }
            finally
            {
            }

            if (!input.ContainsKey("targetId"))
            {
                RGDebug.LogError("No parameter 'targetId' found");
                return;
            }

            string targetIdInput = input["targetId"].ToString();
            int? targetId = default;
            try
            {
                targetId = RGSerialization.Deserialize_int ? (targetIdInput);
            }
            catch (Exception ex)
            {
                RGDebug.LogError("Failed to parse 'targetId'");
                RGDebug.LogError(ex.Message);
            }
            finally
            {
            }
...
```

### After

Generated Deserializer
```csharp
namespace RegressionGames
{
    using UnityEngine;

    public static class RGSerialization
    {
        public static float? Deserialize_float_Nullable(string paramJson)
        {
            return JsonConvert.DeserializeObject<float?>(paramJson);
        }

        public static int? Deserialize_int_Nullable(string paramJson)
        {
            return JsonConvert.DeserializeObject<int?>(paramJson);
        }
    }
}
```

Generated Action
```csharp
        public override void StartAction(Dictionary<string, object> input)
        {
            int skillId = default;
            if (input.TryGetValue("skillId", out var skillIdInput))
            {
                try
                {
                    int.TryParse(skillIdInput.ToString(), out skillId);
                }
                catch (Exception ex)
                {
                    RGDebug.LogError("Failed to parse 'skillId'");
                    RGDebug.LogError(ex.Message);
                }
            }
            else
            {
                RGDebug.LogError("No parameter 'skillId' found");
                return;
            }

            int? targetId = default;
            if (input.TryGetValue("targetId", out var targetIdInput))
            {
                try
                {
                    targetId = RGSerialization.Deserialize_int_Nullable(targetIdInput.ToString());
                }
                catch (Exception ex)
                {
                    RGDebug.LogError("Failed to parse 'targetId'");
                    RGDebug.LogError(ex.Message);
                }
            }

            // no else block for nullable

...
```


rcornwall-reg (approved):
LG

vontell (approved):

RG-nAmKcAz (changes requested):
Marking request changes so this stops pinging me in slack to review

vontell (approved):

Runtime/Scripts/Types/RGClientConnection_Remote.cs
```diff
@@ -151,8 +151,8 @@ public override async void Connect()
                 // calling us and the main thread must be available for Unity webrequests
                 // that we do in this method to process
                 // IOW.. it will deadlock the system
-                var semaphoneAcquired = _connecting.Wait(1);
-                if (semaphoneAcquired && !Connected())
+                var semaphoreAcquired = _connecting.Wait(1);
```
RG-nAmKcAz, line 154: but how can I call anyone without my semaphone ? ... /sigh i suck at spelling :D

RG-nAmKcAz (approved):

Runtime/Scripts/Types/RGClientConnection_Remote.cs
vontell: But how can you call anyone else's semaphone if there is only one 🤔

analogrelay: If we're going to wait on a semphore here, you'll need to use `WaitAsync`, since this is an `async` method (see `async void Connect()` above). Otherwise you increase the risk of thread-pool starvation.

The core C# async "rules of thumb" for me are:

* Avoid `async void` unless absolutely necessary since they cause hidden background thread transitions (I don't have context yet on why this method is already `async void` though).
* When in an `async` method and calling some other method on a type, *always* prefer methods suffixed with `Async`, if they exist.
* Make sure you `await` any call to an `NnnAsync` method

So, this should be:

```suggestion
                var semaphoreAcquired = await _connecting.WaitAsync(1);
```

analogrelay: In addition, we need to be using a `try...finally` block here. Otherwise there's a risk that an exception will be thrown and the Semaphore will never be released.

When using a semaphore, you need to use a pattern like this:

```csharp
var semaphoreAcquired = await _connecting.WaitAsync(1);
try
{
    // Check semaphoreAcquired and do work that should be protected by semaphore
}
finally
{
    if(semaphoreAcquired)
    {
        _connecting.Release();
    }
}
```

The `finally` block ensures that if any exception is thrown in the body of the `try`, control transfers _first_ to the `finally` block (technically it first goes to any `catch` blocks _in the same `try`_, but we don't need one here) before bubbling up the stack.

I realize all this code was here before, but it's a fairly significant thread-safety risk. We should either fix it here or file an issue to address it soon.

analogrelay (changes requested):
I think we should prioritize fixing some thread-safety issues here, but I'm fine if the resolution in this PR is just to file a bug for it so we can address it later.

Runtime/Scripts/Types/RGClientConnection_Remote.cs
RG-nAmKcAz: The try/finally pattern can't be used here unfortunately.  I realize this method's code looks like a bit of unsafe spaghetti but there is a method to the madness.

We are using a `Semaphore` instead of a `Monitor` because we are having to release this on one of the other threads within this block instead of on the thread that acquired it.

While we are establishing the socket connection to the remote bot, we have both the BeginConnect thread running and a connect timeout thread running.  Whichever of these threads 'wins' is when we need to release the `Semaphore`, but both of these tasks finish long after the `Connect` method this code is in has returned.

analogrelay: I'll clear my request changes for now, since it's not relevant to the change at hand, but I'm going to take a scan through some of this code. It's using fairly irregular C# async patterns which might be limiting how much of the language's built-in `async` support we can use. 

> While we are establishing the socket connection to the remote bot, we have both the BeginConnect thread running and a connect timeout thread running. Whichever of these threads 'wins' is when we need to release the Semaphore, but both of these tasks finish long after the Connect method this code is in has returned.

For example, this is a very common pattern that should be doable in a way that still allows us to use `try...finally`.

But I agree that it's not a priority for now, if things are working well enough to carry forward.

analogrelay (approved):

Runtime/Scripts/Types/RGClientConnection_Remote.cs
RG-nAmKcAz: Something to keep in mind as you review the `async` bits down the line. 

Unity is inherently incompatible with `async` for most of its SDK and runtime.

They (Unity) are very big on coroutines and everything running on their main thread.   A lot (not all) of the weird `async` code you'll find in our project is there to workaround limitations from either where we're being called from in Unity, or where we are calling to, or to workaround Unity functions that MUST run on the main thread to work or avoid deadlocking.

I'm sure there are a few cases where we are async and don't need to be, but probably also quite a few where the code looks weird, but there was an underlying limitation/reason we were working around :(

analogrelay: Yep, it's the same as how `async`/`await` interact with UI threads in desktop apps. Unity runs code under a custom `SynchronizationContext` which provides a mechanism to return to the main thread (which is how desktop app frameworks handle this too) but it can be non-trivial to figure out which context code is running under.

We can definitely restructure most of the stuff I see to use `Task`-returning methods and avoid callbacks. Synchronizing back to a "Main" UI thread is very much doable with that pattern, but I definitely don't think it's worth the churn at the moment. Just something to keep in mind as we iterate on this code.

RG-nAmKcAz: Extra context for your research.

The biggest gotcha with `Task` is that you cannot use `UnityWebRequest` or call something that internally instantiates a Unity `Object` anywhere but on the main thread.

See `RGServiceManager` for an example of some of the shenanigans we did to get the UnityWebRequest to be created on the main thread, but still be run `async` from the callers perspective.

PR #67, opened 2023-10-18
[REG-1327] Add ThirdPersonDemo to SDK as a Sample

rcornwall-reg:
**Notes**
* Because the sample was made with Unity's Universal Render Pipeline (URP), the sample will only import correctly in projects using URP. For the built in render pipeline, we would need to rework the sample's shaders and add it as another sample (i.e. `Third Person Demo (Standard)`)
* Adds `Input System` and `Cinemachine` dependencies to the SDK

**To Test**
1. Create a new project, using URP
2. Import SDK
3. In Package Manager, go to Samples -> Third Person Demo (URP) -> Import
4. Open `Samples->Regression Games Unity Bots->0.01->Third Person Demo (URP)->Demo->Scenes->Playground.unity`. This where unity imported the sample, I don't think there's a way to reduce those nested folders
5. Hit Play to play sample

Screen Shot 2023-10-18 at 3 23 02 PM


Samples~/.DS_Store
abeizer, line 1: You have a bunch of these .DS_Store files. You'll want to .gitignore and untrack these

abeizer (changes requested):

abeizer:
What is the goal for this sample? Will we be adding uses of RGAction and RGState to it? Or will users follow a tutorial to get the sample set up themselves?

Samples~/.DS_Store
RG-nAmKcAz: per discussions before this welcome screen task, this sample is supposed to become THE working out of the box sample when you download our SDK and want to see it in action.  It should have and use each of the code patterns we have for generating state and actions as well as a readme explaining where to find each example type such that when you load it up in unity it is self explanatory.

RG-nAmKcAz: it is also 'the thing' that we should always have up to date with the latest patterns

RG-nAmKcAz:
I see your note about adding a standard render pipeline version.  We probably also want one for HDRP.

package.json
```diff
@@ -12,6 +12,15 @@
...
-    "com.unity.textmeshpro": "3.0.6"
-  }
+    "com.unity.textmeshpro": "3.0.6",
+    "com.unity.inputsystem": "1.7.0",
+    "com.unity.cinemachine": "2.9.7"
+  },
+  "samples": [
+    {
+      "displayName": "Third Person Demo (URP)",
+      "description": "This third person demo is a sample Unity game demoing Regression Game’s Unity SDK",
```
RG-nAmKcAz, line 22: Let's add to this to clarify that your project MUST be using URP for this demo scene to work, otherwise you should pick the demo with the correct render pipeline (which I realize don't exist yet).

RG-nAmKcAz (changes requested):
Is there a PR or task for @vontell for removing the 3rd person demo code repository so we don't have 2 disjointed copies of this around?

Is there a follow up task to fixup this project to use our latest code patterns ?

Samples~/.DS_Store
vontell: @RG-nAmKcAz note this task only includes the part of moving it over, my task (https://linear.app/regression-games/issue/REG-1279/update-building-your-first-bot-tutorial-to-use-new-attributes-and-game) is the one where it will use our patterns and have the tutorial updated.

vontell:
@RG-nAmKcAz yes that work for removing the old sample and updating this one are covered here: https://linear.app/regression-games/issue/REG-1279/update-building-your-first-bot-tutorial-to-use-new-attributes-and-game

package.json
rcornwall-reg: Fixed

package.json
```diff
@@ -12,6 +12,15 @@
...
-    "com.unity.textmeshpro": "3.0.6"
-  }
+    "com.unity.textmeshpro": "3.0.6",
+    "com.unity.inputsystem": "1.7.0",
+    "com.unity.cinemachine": "2.9.7"
+  },
+  "samples": [
+    {
+      "displayName": "Third Person Demo (URP)",
+      "description": "This third person demo is a sample Unity game demoing Regression Game’s Unity SDK. It uses Unity's Universal Render Pipeline (URP). To make sure shaders appear correctly, please make sure your project is setup with URP before downloading sample.",
```
RG-nAmKcAz, line 22: I think you missed the word `the` or `this` at the end of this block `... with URP before downloading [the/this] sample.`

RG-nAmKcAz (approved):

vontell (approved):

abeizer (approved):

PR #71, opened 2023-10-23
[REG-1365] Remove package dependencies from SDK and sample

rcornwall-reg:
RGThirdPersonDemo no longer uses Unity's new input system or Cinemachine camera system.

Changes
* Removed New Input System from sample and SDK
* Removed Cinemachine from sample and SDK

Gameplay functionality that changed: 
* Had to remove the "LockOnCamera" from sample, so camera no longer focuses on enemy that's selected
* Removed camera rotation to follow player, now it's a simple follow target camera
* Delay on the input is slightly longer than previous, so player will continue to move for a few milliseconds after releasing movement key. Still feels pretty solid, almost like he has some inertia


RG-nAmKcAz (approved):

vontell:
Tried it in BossRoom and works great! LGTM

vontell (approved):

PR #72, opened 2023-10-24
[REG-1306] Synchronize Unity Bots With RG Server

RG-nAmKcAz:
- Implements synchronization of RG bots between local and remote servers
- Depends on: https://github.com/Regression-Games/RegressionGames/pull/420

### DIFF
--- a/Editor/Scripts/BotManagement/RGBotSynchronizer.cs
+++ b/Editor/Scripts/BotManagement/RGBotSynchronizer.cs
@@ -195,6 +195,8 @@ namespace RegressionGames.Editor.BotManagement
 
             if (File.Exists(zipDownloadPath))
             {
+                // this shouldn't be reached 99.999% of the time but is here to avoid getting stuck
+                // if someone happened to have a file open in the temp directory and blocked it from cleaning up properly
                 File.Delete(zipDownloadPath);
             }
                     
--- a/Editor/Scripts/BotManagement/RGBotSynchronizer.cs
+++ b/Editor/Scripts/BotManagement/RGBotSynchronizer.cs
@@ -60,7 +60,7 @@ namespace RegressionGames.Editor.BotManagement
          *   - Uploads any zips that need updating
          *   - NOTE: Currently this always pushes from local to remote assuming the local is the preferred record for bots that you already have.
          *   -       if you want the new remote bot, you should delete your local bot before syncing.
-         * - Creates new local bot records for any bots on RG that are missing locally  
+         * - Creates new local bot records and downloads the files for any bots on RG that are missing locally  
          */
         [MenuItem("Regression Games/Synchronize Bots with RG")]
         private static async void SynchronizeBots()
--- a/Editor/Scripts/BotManagement/RGBotSynchronizer.cs
+++ b/Editor/Scripts/BotManagement/RGBotSynchronizer.cs
@@ -1,11 +1,17 @@
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text;
-using RegressionGames;
 using RegressionGames.RGBotLocalRuntime;
 using RegressionGames.Types;
 using UnityEngine;
+using CompressionLevel = System.IO.Compression.CompressionLevel;
 using File = UnityEngine.Windows.File;
 #if UNITY_EDITOR
+using System;
+using System.IO.Compression;
+using System.Security.Cryptography;
+using System.Threading.Tasks;
 using UnityEditor;
 using UnityEditor.Compilation;
 #endif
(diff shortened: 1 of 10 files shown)


Editor/Scripts/BotManagement/RGBotSynchronizer.cs
```diff
@@ -31,24 +48,295 @@ private static void CreateNewBot()
...
+        private async Task CreateLocalBotRecordFromRG(RGBot newRemoteBot)
+        {
+            //Create directory
+            string botFolderPath = _this.CreateBotFolder(newRemoteBot.name, newRemoteBot.id);
+
+            string zipDownloadPath = ZipPathForBot(newRemoteBot.id);
+
+            if (File.Exists(zipDownloadPath))
+            {
+                File.Delete(zipDownloadPath);
```
vontell, line 198: Technically this should never be reached, right? Because if we are creating a local bot from RG, that means that the asset with the same ID didn't exist before?

Editor/Scripts/BotManagement/RGBotSynchronizer.cs
```diff
@@ -31,24 +48,295 @@ private static void CreateNewBot()
...
+            var didUpdate = false;
+            // if zip checksums don't match.. push new zip files
+            // Get remote zip details, check md5sum, decide if need to push
+            var needsToUpload = false;
+
+            await _rgServiceManager.GetBotCodeDetails(
+                botId,
+                (botCodeDetails) =>
+                {
+                    if (md5 != botCodeDetails.md5)
```
vontell, line 258: Forgot to put this in my other PR, do we need a migration for the existing zip bots? I don't really mind much considering no users... but curious about that.

Editor/Scripts/BotManagement/RGBotSynchronizer.cs
```diff
@@ -76,14 +368,32 @@ private string CreateBotFolder(string botName, long botId)
 
         private void CreateNewBotAssets(string folderName, string botName, long botId)
```
vontell, line 369: Should we call our synchronize call immediately after creating a new bot?

vontell (changes requested):

Editor/Scripts/BotManagement/RGBotSynchronizer.cs
RG-nAmKcAz: shouldn't need any migration.. this is computed on the fly in backend from reading the file stream itself, not saved in the DB

Editor/Scripts/BotManagement/RGBotSynchronizer.cs
RG-nAmKcAz: should never be reached.. but.. i had a file open in testing.. and the ziptemp folder failed to delete.. and on the next run one of these files still existed...

Editor/Scripts/BotManagement/RGBotSynchronizer.cs
RG-nAmKcAz: I wouldn't personally.  If we did, I would only consider pushing that 1 bot, not pulling down or updating all the others as that could take quite a bit of time and bandwidth depending on your number of bots.

Editor/Scripts/BotManagement/RGBotSynchronizer.cs
```diff
@@ -31,24 +48,295 @@ private static void CreateNewBot()
...
+         * This is a long one, it...
+         * - Discovers all local bots
+         * - Gets the local Unity bot records that exist on RG 
+         * - Creates any bots on RG that don't already exist
+         * - Creates a zip file of local bots (excluding some key files)
+         * - Gets and compares the MD5 of the local bots to those on RG 
+         *   - Uploads any zips that need updating
+         *   - NOTE: Currently this always pushes from local to remote assuming the local is the preferred record for bots that you already have.
+         *   -       if you want the new remote bot, you should delete your local bot before syncing.
+         * - Creates new local bot records for any bots on RG that are missing locally
```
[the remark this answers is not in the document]
RG-nAmKcAz: slightly reworded last line to include reference to downloading files

Editor/Scripts/BotManagement/RGBotSynchronizer.cs
RG-nAmKcAz: added comment to code

Editor/Scripts/BotManagement/RGBotSynchronizer.cs
RG-nAmKcAz: For context; I say I wouldn't because getting the id reserved doesn't do a whole lot and you won't have your bot code in place right after creating it from the menu.

vontell (approved):

abeizer (approved):

addisonbgross (approved):

PR #75, opened 2023-10-26
[REG-1367] Add 'Third Person Sample' to welcome screen

rcornwall-reg:
**Changes**
* Added an automatic 'Import Sample' button to the welcome screen
* Automatically opens Playground scene after import
* Enables/Disables import button if the project is setup with URP

_Enabled when using URP_
Screen Shot 2023-10-25 at 6 10 24 PM

_Disabled when not using URP_
Screen Shot 2023-10-25 at 6 13 05 PM

### DIFF
--- a/Editor/Scripts/Startup/RegressionPackagePopup.cs
+++ b/Editor/Scripts/Startup/RegressionPackagePopup.cs
@@ -250,6 +250,12 @@ public class RegressionPackagePopup : EditorWindow
         {
             wordWrap = true
         };
+        
+        GUIStyle boldDescriptionStyle = new GUIStyle(EditorStyles.label)
+        {
+            wordWrap = true,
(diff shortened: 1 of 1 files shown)


abeizer (changes requested):
My only comment is that from your sceeenshots, it looks like the description text overflows its box on the right side. Otherwise, LGTM

RG-nAmKcAz (approved):
Approving Pending Abby's overflow comment being addressed.

I would also add that the URP disclaimer may look better as a separate line and possible either in ()'s or italicised.

vontell (approved):

rcornwall-reg:
Screen Shot 2023-10-26 at 9 21 11 AM

abeizer (approved):

PR #77, opened 2023-10-26
[REG-1372] Resolve compile warnings in generated code

RG-nAmKcAz:
Our generated state and action classes were generating compiler warnings

- inconsistent line endings
- duplicate using declarations

Also.. the samplebots for bossroom I had here were generating warnings.  They shouldn't exist in this location anyway and thus have been removed.   (They live on in the bossroom project)


abeizer (approved):

rcornwall-reg (approved):
Much cleaner - LG

PR #78, opened 2023-10-26
[REG-1337] Populate ObjectType in context files for custom assemblies and namespaces

abeizer:
Collect all assemblies in the project and attempt to locate which assembly contains a given action/state script. This is required during the "Extract" step for Agent Builder, where we need to locate each class again. To do this, we call `Type.GetType` which requires the assembly name


Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
```diff
@@ -354,61 +414,85 @@ private static string GetFullTypeNameWithNamespace(ITypeSymbol typeSymbol, Seman
...
-            // Get all Object names from RGStates
-            foreach (var state in statesInfo)
-            {
-                string objectName = state.Object;
-                if (!objectTypeNames.Contains(objectName))
+                    objectTypeMap.TryAdd(objectType, null);
+                }else
                 {
-                    objectTypeNames.Add(objectName);
+                    RGDebug.LogWarning("Type not found: " + key);
```
RG-nAmKcAz, line 473: Seeing this in the log as a developer using our SDK wouldn't tell me much. 

Can we log anything else to help them identify the gameObject they're looking for ?

abeizer: This key is the assembly + full namespace + className of the script we're attempting to look up, which imo is useful information for figuring out _what_ failed, although there's not really anything a user could do to fix this on their end :-/ 
Our current approach is to look up all of the scripts and then associate the name of a GameObject to each of them - unfortunately, this step comes before we attempt to locate any gameObjects from the scene + prefabs, so we have no specific gameObject in context at this point.

I do have a task open [here](https://linear.app/regression-games/issue/REG-1375/actions-and-states-in-game-context-can-be-associated-with-multiple) for allowing multiple objectTypes to be associated to a given action/state script (currently, only one objectType "wins" and makes it into the JSON per script, which is wrong). I'm thinking when we address this task, we'll probably want to reverse this whole process - look up the gameObjects and then associate all of their known action/state scripts to them. We could probably give a more helpful warning about which gameObjects (or at least objectTypes) are affected at that point.

abeizer: Ah, you beat me to a response

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
```diff
@@ -354,61 +414,85 @@ private static string GetFullTypeNameWithNamespace(ITypeSymbol typeSymbol, Seman
             return namespacePrefix + typeSymbol.Name;
         }
 
+
+        private static string GetActionKey(RGActionInfo action, bool includeAssembly)
```
vontell, line 418: What is meant by key here? I see this is returning the namespace, should this actually be GetActionNamespace? A comment on both of these functions would be helpful!

vontell (approved):
Just a comment comment :D lgtm otherwise!

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
rcornwall-reg: @vontell `Type.GetType` requires the fully qualified type of the component and the assembly it's in, as a string with the format: "{namespace}.{typeName}, {assemblyName}" so it's a bit more than the namespace. It's the lookup key for GetType.

Comment would probably be helpful for that :)

rcornwall-reg (approved):
LGTM

Editor/Scripts/CodeGenerators/RGActionCodeGenerator.cs
abeizer: I liked your explanation so heavily borrowed from it ;-)
@vontell let me know if this makes sense now

vontell: Makes sense to me!

vontell (approved):

PR #79, opened 2023-10-26
[REG-1376] Behavior Tree Bots and supporting components

analogrelay:
This adds a new namespace: `RegressionGames.BehaviorTree`. Inside this namespace are types to support Behavior Tree-based bots in the Unity SDK. This includes bots generated by Agent Builder, but also allows users to integrate behavior trees into their own bots if they'd like to use a hybrid approach.
 
A Behavior Tree-based bot inherits from `RGBehaviorTreeBot` **instead** of `RGUserBot`. Then, instead of overriding `ProcessTick`, they override `BuildBehaviorTree` to construct their behavior tree and return the root node.

As part of this change, I've made `RGUserBot.Init` `internal`. I don't think there's reason for a user script to call it, so it probably shouldn't be public, but I can revert that if we know user scripts need to call it. It's also `virtual` now, which allows `RGBehaviorTreeBot` to use it to call the user-provided `BuildBehaviorTree` method. That avoids the weird `ConfigureBotCore` behavior I had in [the earlier sketch](https://github.com/Regression-Games/RGBossRoom/pull/38).

----

I think it's preferable to put these types in the Unity SDK rather than generate them with the Agent Builder bot (which is what we do in TypeScript). If we generate them alongside the bot, we create a few issues:

* Lots of code duplication when there are multiple local bots in a single game.
* Code completion will be polluted with multiple copies of the Behavior Tree types (since most IDEs now are extra helpful and suggest types in namespaces not currently imported)
* We have no ability to make bug-fixes and improvements to Behavior Tree logic

That does mean that if we add new node types to the Agent Builder, we will have to consider the version of our SDK that the bot is targeting. However, we'll need to do that anyway for the code _inside_ of each node, so it's not really more complication than we're already signing ourselves up for.

### DIFF
--- a/Runtime/Scripts/BehaviorTree/Decorators.cs
+++ b/Runtime/Scripts/BehaviorTree/Decorators.cs
@@ -1,0 +1,56 @@
+using System;
+using RegressionGames.RGBotLocalRuntime;
+
+namespace RegressionGames.BehaviorTree
+{
+    /// <summary>
+    /// A node that always returns <see cref="NodeStatus.Success"/> when executed.
+    /// </summary>
+    public class AlwaysSucceed: BehaviorTreeNode
+    {
+        public AlwaysSucceed() : base("Always Succeed")
+        {
+        }
+
+        public override NodeStatus Execute(RG rgObject) => NodeStatus.Success;
+    }
+    
+    /// <summary>
+    /// A node that always returns <see cref="NodeStatus.Failure"/> when executed.
+    /// </summary>
+    public class AlwaysFail: BehaviorTreeNode
+    {
+        public AlwaysFail() : base("Always Fail")
+        {
+        }
+
+        public override NodeStatus Execute(RG rgObject) => NodeStatus.Failure;
+    }
+    
+    /// <summary>
+    /// A node that inverts the result of it's child node.
+    /// </summary>
+    /// <remarks>
+    /// If the child node returns <see cref="NodeStatus.Running"/>, this node returns <see cref="NodeStatus.Running"/>.
+    /// If the child node returns <see cref="NodeStatus.Failure"/>, this node returns <see cref="NodeStatus.Success"/>.
+    /// If the child node returns <see cref="NodeStatus.Success"/>, this node returns <see cref="NodeStatus.Failure"/>.
+    /// </remarks>
+    public class Invert: BehaviorTreeNode
+    {
+        private readonly BehaviorTreeNode _child;
+
+        public Invert(BehaviorTreeNode child) : base("Invert")
+        {
+            _child = child;
+        }
+
+        public override NodeStatus Execute(RG rgObject) =>
+            _child.Execute(rgObject) switch
+            {
+                NodeStatus.Success => NodeStatus.Failure,
+                NodeStatus.Failure => NodeStatus.Success,
+                NodeStatus.Running => NodeStatus.Running,
+                var other => throw new InvalidOperationException($"Unexpected node status: {other}"),
+            };
+    }
+}
--- a/Runtime/Scripts/BehaviorTree/Decorators.cs
+++ b/Runtime/Scripts/BehaviorTree/Decorators.cs
@@ -1,30 +1,57 @@
 using System;
+using Codice.CM.Common;
 using RegressionGames.RGBotLocalRuntime;
 
 namespace RegressionGames.BehaviorTree
 {
     /// <summary>
-    /// A node that always returns <see cref="NodeStatus.Success"/> when executed.
+    /// Base class for decorator nodes.
     /// </summary>
-    public class AlwaysSucceed: BehaviorTreeNode
+    public abstract class DecoratorNode : BehaviorTreeNode
     {
-        public AlwaysSucceed() : base("Always Succeed")
-        {
-        }
+        protected readonly BehaviorTreeNode Child;
 
-        public override NodeStatus Execute(RG rgObject) => NodeStatus.Success;
+        /// <summary>
+        /// Creates a new decorator node with the specified name and child node.
+        /// </summary>
+        /// <param name="name">The name of the node.</param>
+        /// <param name="child">The <see cref="BehaviorTreeNode"/> child of this node.</param>
+        protected DecoratorNode(string name, BehaviorTreeNode child) : base(name)
+        {
+            Child = child;
+        }
     }
     
     /// <summary>
-    /// A node that always returns <see cref="NodeStatus.Failure"/> when executed.
+    /// A node that executes it's child, but ignores the result and always returns <see cref="NodeStatus.Success"/>.
     /// </summary>
-    public class AlwaysFail: BehaviorTreeNode
+    public class AlwaysSucceed: DecoratorNode
     {
-        public AlwaysFail() : base("Always Fail")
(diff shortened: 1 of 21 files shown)


addisonbgross (approved):

addisonbgross:
I approved this but I have 1 question: Would it be worth mentioning that the Decorator type nodes all only allow for 1 child node? This point is in our docs, but maybe it would be nice for the Intellisense to mention it as well

vontell (approved):
This looks great!

No need to do this now, but this is a nice, isolated piece of code that may serve as a good start for unit testing our Unity SDK. However, we should make a separate ticket for that, since I am not sure about the intricacies of unit testing for Unity package...

Runtime/Scripts/BehaviorTree/Decorators.cs
```diff
@@ -0,0 +1,56 @@
+using System;
+using RegressionGames.RGBotLocalRuntime;
+
+namespace RegressionGames.BehaviorTree
+{
+    /// <summary>
+    /// A node that always returns <see cref="NodeStatus.Success"/> when executed.
```
abeizer, line 7: The decorators should clarify that they alter/override their child's status

Runtime/Scripts/BehaviorTree/SequenceNode.cs
```diff
@@ -0,0 +1,33 @@
+using System.Linq;
+using JetBrains.Annotations;
+using RegressionGames;
+using RegressionGames.RGBotLocalRuntime;
+using TreeEditor;
+
+namespace RegressionGames.BehaviorTree
+{
+    /// <summary>
+    /// Represents a node that will run all of it's children in sequence.
```
abeizer, line 10:
```suggestion
    /// Represents a node that will run all of its children in sequence.
```

Runtime/Scripts/BehaviorTree/SelectorNode.cs
```diff
@@ -0,0 +1,29 @@
+using RegressionGames.RGBotLocalRuntime;
+
+namespace RegressionGames.BehaviorTree
+{
+    /// <summary>
+    /// Represents a node that will run each of it's children in sequence.
```
abeizer, line 6:
```suggestion
    /// Represents a node that will run each of its children in sequence.
```

Runtime/Scripts/BehaviorTree/RootNode.cs
```diff
@@ -0,0 +1,26 @@
+using System.Collections.Generic;
+using System.Linq;
+using RegressionGames.RGBotLocalRuntime;
+
+namespace RegressionGames.BehaviorTree
+{
+    /// <summary>
+    /// Represents the root node of a behavior tree.
+    /// The root node carries the data dictionary that is shared between all nodes in the tree.
```
abeizer, line 9:
```suggestion
    /// The root node carries a data dictionary that is shared between all nodes in the tree.
```

abeizer (approved):
Minor wording suggestions for comments

analogrelay:
> I approved this but I have 1 question: Would it be worth mentioning that the Decorator type nodes all only allow for 1 child node? This point is in our docs, but maybe it would be nice for the Intellisense to mention it as well

I think that's well covered by the API itself. There's no mechanism to add more than one child to the decorator nodes. They don't inherit from `ContainerNode` and thus don't have an `AddChild` method. They take just a single inner node in their constructor.

Runtime/Scripts/BehaviorTree/RGBehaviorTreeBot.cs
```diff
@@ -0,0 +1,23 @@
...
+        internal override void Init(long botId, string botName)
+        {
+            base.Init(botId, botName);
+            _rootNode = BuildBehaviorTree();
+        }
+
+        public override void ProcessTick(RG rgObject)
+        {
+            var result = _rootNode.Execute(rgObject);
+            RGDebug.LogVerbose($"Behavior tree completed with status: {result}");
```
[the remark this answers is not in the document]
analogrelay: That's a good question. Returning non-`Success` from the root doesn't necessarily mean a failure. Condition nodes can return `Failure` just to indicate that the condition isn't met.

> Do we log enough information or give them any context to know which node failed and context to help determine why ?

I'll do a pass and see if I can get some per-node logging in place that makes sense. I think there's more we can do here (tracking the result of each node so we can capture the full execution flow, etc.) but I think we can iterate there as we understand the exact debugging scenarios better.

And of course, all of the user's nodes are in their project. So, they have the ability to set breakpoints in the debugger and trace execution that way.

analogrelay:
> No need to do this now, but this is a nice, isolated piece of code that may serve as a good start for unit testing our Unity SDK. However, we should make a separate ticket for that, since I am not sure about the intricacies of unit testing for Unity package...

100% agreed. I'd like to get some infrastructure in place to test out the Unity package in general. Right now I work on this repo primarily by opening RGBossRoom so I have something to test with, but it'd be good if we could have a separate sample project _inside_ the repo (but not part of the package) so that we have a "host" project to use when working on the package.

Runtime/Scripts/BehaviorTree/RGBehaviorTreeBot.cs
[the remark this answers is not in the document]
abeizer: RE: statuses at the root node:

Failure -> For an individual node, this mean simply means the node failed - an action couldn't be completed, a condition was falsey, etc. At the root, this means that the tree couldn't successfully complete _any_ of its branches. Depending on how the user sets up their tree, a tick that resulted in a failure _could_ still effect the state in some way (ex. moving to an enemy but failing to attack it, which causes that sequence to fail).

Running -> You're waiting on some action to complete. It could be one that you started this tick, or one that you started on a previous tick. This is almost the same as "Success", but indicates that you're effectively short-circuiting the tree and returning that status immediately all the way to the root, whereas a "Success" might continue executing other nodes in the tree (in a sequence, for example)

vontell:
> > No need to do this now, but this is a nice, isolated piece of code that may serve as a good start for unit testing our Unity SDK. However, we should make a separate ticket for that, since I am not sure about the intricacies of unit testing for Unity package...
> 
> 100% agreed. I'd like to get some infrastructure in place to test out the Unity package in general. Right now I work on this repo primarily by opening RGBossRoom so I have something to test with, but it'd be good if we could have a separate sample project _inside_ the repo (but not part of the package) so that we have a "host" project to use when working on the package.

You'll be glad to know that @rcornwall-reg recently added a game sample to the SDK! So yes, we will have one to test with that's included as part of the SDK.

Runtime/Scripts/BehaviorTree/Decorators.cs
analogrelay: Ah, I see now that the `AlwaysFail` and `AlwaysSucceed` nodes should still accept one child. I'll make that change and update the docs. Currently they cannot accept any child nodes.

analogrelay:
Updates!
* We log the execution of each step in the tree at `Verbose` level for now. This might be more information than you want, and we'll want to tune the diagnostics and debugging experiences over time
* Added `SetCharacterConfigFromJson` to `RG` so that we can easily load the character config JSON blob that Agent Builder dumps into our bot.

abeizer (approved):

PR #81, opened 2023-10-31
DRAFT: [REG-1359] Fixes to support C# Agent Builder bots

abeizer:
* Selector node must return `Failure` if all of its children failed
* Fix distance comparison in `FindNearestEntity`
* `MathFunctions` is now accessible through the `RG` class. This lets us call `DistanceSq` from Agent Builder


Runtime/Scripts/BehaviorTree/SelectorNode.cs
```diff
@@ -19,7 +20,7 @@ protected override NodeStatus Execute(RG rgObject)
                 }
             }
 
-            return NodeStatus.Success;
+            return NodeStatus.Failure;
```
analogrelay, line 23: 🤦🏻‍♀️ 

I figured we'd identify a bug or two here though. Good catch!

analogrelay (approved):

vontell (approved):

addisonbgross (approved):

RG-nAmKcAz (approved):

rcornwall-reg (approved):

PR #82, opened 2023-10-31
[REG-1363] Support custom RGState in AgentBuilder json & CodeGenerator restructure

RG-nAmKcAz:
- Handles custom written RGState classes being reflected in the state for AgentBuilder
  - (Note: This all started as a way to get the Platformer state classes into the state json)
- Cleans up CodeGenerator to only read/write json when exporting for AgentBuilder instead of during class generation
- Adds progress bars to the UI when extracting and generating
- Reworks how we try to find assembly names for types to avoid having to lookup .cs files from filesystem
  - Makes this work across packages and inside dlls
- Addresses a few quirks around populating the ObjectType from the RGStateEntity
- Adds RG SDK classes to the state/actions jsons, not just those from the user's Assets
- Removes elements from RGStates.json that are not needed for AgentBuilder
- Removes elements from RGActions.json that are not needed for AgentBuilder
- Cleans up naming scheme for generated `RGState_` and `RGAction_` and `RGStateEntity_` and `RGActionRequest_` classes

This PR set out to just handle `RGState` classes written outside of `RGState` attributes.  Apologies to reviewers as its scope changed into a restructure of the code generation process in addition to the original objectives.

Other PRs that go with this one
- (https://github.com/Regression-Games/PlatformerSample/pull/2)
- (https://github.com/Regression-Games/RGBossRoom/pull/41)
- Match3Sample ... no changes needed
- (https://github.com/Regression-Games/RGUnitySample/pull/12)

RGStates.json
```json
{
  "RGStateInfo": [
    {
      "ClassName": "RegressionGames.RGBotConfigs.RGState_EnemyController",
      "ObjectType": "Enemy",
      "State": [
        {
          "StateName": "isAlive",
          "Type": "bool"
        }
      ]
    },
    {
      "ClassName": "RegressionGames.RGBotConfigs.RGState_Button",
      "ObjectType": null,
      "State": [
        {
          "StateName": "interactable",
          "Type": "bool"
        }
      ]
    },
    {
      "ClassName": "RegressionGames.RGBotConfigs.RGStateProviders.RGState_Platformer2DLevel",
      "ObjectType": "Level",
      "State": [
        {
          "StateName": "tileCellSize",
          "Type": "Vector3"
        },
        {
          "StateName": "platformPositions",
          "Type": "RegressionGames.RGBotConfigs.RGStateProviders.RGPlatformer2DPosition[]"
        }
      ]
    },
    {
      "ClassName": "RegressionGames.RGBotConfigs.RGStateProviders.RGState_Platformer2DPlayer",
      "ObjectType": "Player",
      "State": [
        {
          "StateName": "jumpHeight",
          "Type": "float"
        },
        {
          "StateName": "maxJumpHeight",
          "Type": "float"
        },
        {
          "StateName": "velocity",
          "Type": "float"
        },
        {
          "StateName": "maxVelocity",
          "Type": "float"
        },
        {
          "StateName": "safeFallHeight",
          "Type": "float"
        },
        {
          "StateName": "nonFatalFallHeight",
          "Type": "float"
        }
      ]
    },
    {
      "ClassName": "RegressionGames.RGBotConfigs.RGStateProviders.RGState_ZackFrustratedAtRoslyn",
      "ObjectType": null,
      "State": [
        {
          "StateName": "stringField",
          "Type": "string"
        },
        {
          "StateName": "boolField",
          "Type": "bool"
        },
        {
          "StateName": "vector3ArrayField",
          "Type": "Vector3[]"
        }
      ]
    }
  ]
}
```

RGActions.json
```json
{
  "BotActions": [
    {
      "ClassName": "RegressionGames.RGAction_setZero",
      "ObjectType": null,
      "ActionName": "setZero",
      "Parameters": [
        {
          "Name": "newZero",
          "Type": "Vector2",
          "Nullable": false
        }
      ]
    },
    {
      "ClassName": "RegressionGames.RGAction_setZeroNullable",
      "ObjectType": null,
      "ActionName": "setZeroNullable",
      "Parameters": [
        {
          "Name": "newZero",
          "Type": "Vector2?",
          "Nullable": true
        }
      ]
    },
    {
      "ClassName": "RegressionGames.RGAction_SetBotMove",
      "ObjectType": "Player",
      "ActionName": "SetBotMove",
      "Parameters": [
        {
          "Name": "x",
          "Type": "float",
          "Nullable": false
        }
      ]
    },
    {
      "ClassName": "RegressionGames.RGAction_SetBotJump",
      "ObjectType": "Player",
      "ActionName": "SetBotJump",
      "Parameters": [
        {
          "Name": "onOff",
          "Type": "bool",
          "Nullable": false
        }
      ]
    }
  ]
}
```


RG-nAmKcAz:
Done/ready-for-review.. links for other projects changed coming soon...

abeizer:
Your examples for state + action json both include `null` objectTypes. Is this intentional? I wouldn't expect to see these output because if nothing is using them then Agent Builder is better off not knowing about them

RG-nAmKcAz:
> Your examples for state + action json both include `null` objectTypes. Is this intentional? I wouldn't expect to see these output because if nothing is using them then Agent Builder is better off not knowing about them

Currently yes that is intentional, we can purge those from the output if desired though fairly easily.

abeizer (approved):
Class + content generation works for BossRoom, and bots run without issue (see https://github.com/Regression-Games/RGBossRoom/pull/42). Really only good things to say here and notes on some things you and I talked about that may need to be recorded in Linear.

The filename change, while small, is suuuuper helpful in the component search on prefabs. Wow.

I adore these progress bars. They make it very easy to tell when the editor is actively generating content versus waiting on some background task, or already having completed generation. I didn't realize this annoyed me until it was no longer an issue. It's also neat to have some insight into what steps our SDK is taking as a user.

As noted in a previous comment, it'd be nice to remove entries with a null/empty ObjectType BUT in the interest of getting this cycle wrapped up I think this can be handled as a separate task. At this point in time, I can't say for certain how unused entries will affect GPT, although I'm fairly confident it will ignore them. It's possible null could be interpreted as a globally-shared state, but even then GPT has been very good about using appropriate states for the given prompt.

I'm not sure whether a task for this already exists, I couldn't find anything searching in linear, but we'd chatted about collecting states into one JSON entry per ObjectType. This would be awesome as a future... although may end up being covered under [REG-1375](https://linear.app/regression-games/issue/REG-1375/actions-and-states-in-game-context-can-be-associated-with-multiple) (in which case, ignore me).

abeizer (changes requested):
Ah I missed this issue because it's only with states with different names than their variable/method (in the case of BossRoom we only do this with methods): it seems we're extracting the _actual_ name rather than the override. 

Ex 
```json
       {
          "StateName": "HasQueuedActions", // instead of "hasQueuedActions"
          "Type": "bool"
        },
```

EDIT: my debugging skills are bad today

We're pulling the name from the generated RGStateEntity class, which uses the actual name
```csharp
        public bool HasQueuedActions => ...

```

We need to update the `GeneratePropertyDeclarationForNumbers` and `GeneratePropertyDeclarationForNormalTypes` to set the first occurrence of `PropertyDeclaration` to 
```csharp
               return PropertyDeclaration(
                    IdentifierName(memberInfo.Type),
                    Identifier(memberInfo.StateName) // instead of FieldName
                )
```

RG-nAmKcAz:
> Ah I missed this issue because it's only with states with different names than their variable/method (in the case of BossRoom we only do this with methods): it seems we're extracting the _actual_ name rather than the override.
> 
> Ex
> 
> ```json
>        {
>           "StateName": "HasQueuedActions", // instead of "hasQueuedActions"
>           "Type": "bool"
>         },
> ```
> 
> EDIT: my debugging skills are bad today
> 
> We're pulling the name from the generated RGStateEntity class, which uses the actual name
> 
> ```cs
>         public bool HasQueuedActions => ...
> ```
> 
> We need to update the `GeneratePropertyDeclarationForNumbers` and `GeneratePropertyDeclarationForNormalTypes` to set the first occurrence of `PropertyDeclaration` to
> 
> ```cs
>                return PropertyDeclaration(
>                     IdentifierName(memberInfo.Type),
>                     Identifier(memberInfo.StateName) // instead of FieldName
>                 )
> ```

Good catch.. now resolved.. see `[REG-1363] testing for unmatched RGState names` commit on this PR and the one in (https://github.com/Regression-Games/PlatformerSample/pull/2)

abeizer (approved):

Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
```diff
@@ -9,107 +9,181 @@
...
+using RegressionGames.StateActionTypes;
 using UnityEditor;
 using UnityEngine;
 using Object = UnityEngine.Object;
 
 namespace RegressionGames.Editor.CodeGenerators
 {
-    public class RGActionCodeGenerator
+    /**
+     * This calls takes a multi pass approach for states, but a single pass approach for actions.
```
vontell, line 20: Typo in comment, this class?

vontell (approved):
Lots of changes! Approving since I don't see any issues, but will also leave it to the folks who have been involved in this area of the code as well (@rcornwall-reg @abeizer)

RG-nAmKcAz:
> Overriden state names that include a space throws a compile error in the generated code:
> 
> ```
> [RGState("My State")] public int? myNullableState;
> ```
> 
> generates:
> 
> ```
> public int? My State => (int?)this.GetValueOrDefault("My State");
> ```
> 
> I think it just needs a `Replace(' ', '_')` somewhere in the generator. Actions that override names and include spaces work great though.

resolved

Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
RG-nAmKcAz: fixed

rcornwall-reg (commented):
I'm still some compiler errors in the generated scripts. Here's the full tagged attribute test I'm running. Can you paste it into a project and see if the generated files turn out correct for you? @RG-nAmKcAz 

```
        [RGAction]
        public void TestActionOne(Vector3 vectorParam, float floatParam)
        {
            Debug.Log("Test");
        }

        [RGAction("Test Action Two")]
        public void TestActionTwo(int? nullableIntParam)
        {
            Debug.Log("Test");
        }

        [RGState] public float floatState;
        [RGState] public Vector3 vectorState;
        [RGState("My State")] public int? myNullableState;

        [RGState]
        public int GetIntState()
        {
            return 6;
        }
        ```

RG-nAmKcAz:
> I'm still some compiler errors in the generated scripts. Here's the full tagged attribute test I'm running. Can you paste it into a project and see if the generated files turn out correct for you? @RG-nAmKcAz
> 
> ```
>         [RGAction]
>         public void TestActionOne(Vector3 vectorParam, float floatParam)
>         {
>             Debug.Log("Test");
>         }
> 
>         [RGAction("Test Action Two")]
>         public void TestActionTwo(int? nullableIntParam)
>         {
>             Debug.Log("Test");
>         }
> 
>         [RGState] public float floatState;
>         [RGState] public Vector3 vectorState;
>         [RGState("My State")] public int? myNullableState;
> 
>         [RGState]
>         public int GetIntState()
>         {
>             return 6;
>         }
>         ```
> ```

Fixing, but both of these (i fixed states, not actions in my last push) seem to be pre-existing bugs.  Good tests none the less.. fixing up

> I'm still some compiler errors in the generated scripts. Here's the full tagged attribute test I'm running. Can you paste it into a project and see if the generated files turn out correct for you? @RG-nAmKcAz
> 
> ```
>         [RGAction]
>         public void TestActionOne(Vector3 vectorParam, float floatParam)
>         {
>             Debug.Log("Test");
>         }
> 
>         [RGAction("Test Action Two")]
>         public void TestActionTwo(int? nullableIntParam)
>         {
>             Debug.Log("Test");
>         }
> 
>         [RGState] public float floatState;
>         [RGState] public Vector3 vectorState;
>         [RGState("My State")] public int? myNullableState;
> 
>         [RGState]
>         public int GetIntState()
>         {
>             return 6;
>         }
>         ```
> ```

I got no compile errors with your sample.  Are you sure you have my latest commit pulled down ?

rcornwall-reg (approved):

PR #84, opened 2023-11-01
[REG-1382] Include core state fields just from RGEntity

RG-nAmKcAz:
- Removes the need to put the basic `RGState` on a GameObject with `RGEntity`
  - The 'core' state fields like id/position/etc are now automatic if your object is an `RGEntity`
  - `RGState` class changed to `abstract` to enforce this and make it easy/obvious to find and remove from objects in the editor
- Tries to make the combined `RGStateEntity` in the state for an object with a custom `RGState` subclass implementation  be the correct class type for easier C# coding
- Adds an option to `RGEntity` to get the 'everything' state for that object with the click of a button (verbosely documented with tooltip in the editor)
- Re-orders the `RGEntity` fields in the editor so that the object type entry is first above the 3 checkboxes
- Adds help url links to our most visible editor classes to link the (?) button to our docs website
- Re-Imagines how we represent buttons and actions
  - RGEntity is now all that is needed on a button.  It automatically handles interactable and actions
  - RGAction_ClickButton now lives ONLY on the RGOverlayMenu and handles directing all button clicks to the correct/requested target

See also...
- (https://github.com/Regression-Games/PlatformerSample/pull/3)
- (https://github.com/Regression-Games/RGBossRoom/pull/43)

image


RG-nAmKcAz:
@vontell This is the PR that moves `RGState` to `abstract` and thus needs the Sample updates for the `Enemy` prefab

vontell:
@RG-nAmKcAz will do! You are fine with merging this and then me updating that part, correct? In the meantime, if you wanted to, you could just remove that script from the enemy prefab since it isn't needed at this PR anyways.

Runtime/Scripts/RGBotConfigs/RGEntity.cs
```diff
@@ -6,18 +6,23 @@
 
 namespace RegressionGames.RGBotConfigs
 {
+    [HelpURL("https://docs.regression.gg/studios/unity/unity-sdk/overview")]
```
vontell, line 9: Woah! I wanna use this more often

RG-nAmKcAz:
> @RG-nAmKcAz will do! You are fine with merging this and then me updating that part, correct? In the meantime, if you wanted to, you could just remove that script from the enemy prefab since it isn't needed at this PR anyways.

I think we can wait for your PR to fix it.  The sample runs as is, it just doesn't work for RG Bots, which it didn't anyway without your upcoming changes.

vontell (approved):

Runtime/Scripts/RGBotConfigs/RGState.cs
```diff
@@ -68,6 +59,73 @@ protected virtual RGStateEntity CreateStateEntity()
...
+            // find all Components and get their values
+            var components = entity.gameObject.GetComponents<Component>();
+            foreach (var component in components)
+            {
+                // skip 'expensive' components, only get colliders and MonoBehaviours
+                if (component is Collider or Collider2D or MonoBehaviour and not RGState and not RGAction and not RGEntity and not RGAgent)
+                {
+                    var type = component.GetType();
+                    var dictionary = new Dictionary<string, object>();
+                    foreach (PropertyInfo prop in type.GetProperties())
```
rcornwall-reg, line 101: Would it be possible to find these properties _without_ using reflection? Or did we ever test if reflection in runtime builds is fully supported on the platforms our SDK supports?

https://docs.unity3d.com/Manual/ScriptingRestrictions.html

RG-nAmKcAz: For any arbitrary user written class, it is unlikely we could find a non-reflection solution that didn't cause them extra work (thus defeating the purpose of adding this feature in the first place).

I would be happy though to enhance the tooltip disclaimer to indicate that this option is only safe in situations where `AOT` is not in use and give them the link you pasted (https://docs.unity3d.com/Manual/ScriptingRestrictions.html)

rcornwall-reg: From the link, `AOT` looks like its supported when the backend scripting is `IL2CPP` and not when it's `Mono`. We can grab the scripting backend through an Editor script like:

```
using UnityEditor
public static string GetScriptingBackend() 
{
   BuildTargetGroup buildTarget = EditorUserBuildSettings.selectedBuildTargetGroup;
   ScriptingImplementation scriptingBackend = PlayerSettings.GetScriptingBackend(buildTargetGroup);
   return scriptingBackend.ToString(); // will return Mono or IL2CPP
}
```
We could implement a custom inspector for RGState and disable the checkbox if they're on a Mono scripting backend. 

I'm not seeing this as a blocking change though, so if you could at least add to the tooltip that'd look good to me

Runtime/Scripts/RGBotServerListener.cs
```diff
@@ -515,41 +516,41 @@ private void FixedUpdate()
...
-                    if (rgEntity == null) continue;
+                // get the state behaviors on this game object
+                var rgStates = statefulRGEntity.gameObject.GetComponents<IRGState>();
                 
-                    if (true.Equals(gameObjectState.isPlayer))
+                // build up the state for this entity from all the different state classes
+                foreach (var rgState in rgStates)
+                {
+                    var gameObjectStateEntity = rgState.GetGameObjectState();
+                    if (true.Equals(gameObjectStateEntity.isPlayer))
```
abeizer, line 545: `isPlayer` check here should be from the core (RGEntity) state, and we don't need to repeat it for every RGstate script

Runtime/Scripts/RGBotServerListener.cs
```diff
@@ -559,41 +560,48 @@ private void FixedUpdate()
...
-                        }
-                        combinedGameObjectState[x.Key] = x.Value;
-                    });
-                    fullGameState[key] = combinedGameObjectState;
-                }
-                else
-                {
-                    fullGameState[key] = gameObjectState;
+                            // Note that core state files like "position", "rotation", "isPlayer", etc are ignored.
+                            // They are part of the core state and can be overridden, but we don't log a warning for it.
```
abeizer, line 587: (Future) If it's possible to override these, then we should consider not allowing users to do that. These are keys reserved for very specific concepts - if the user wants to "override" them then they can always define a new key.

abeizer (changes requested):

Runtime/Scripts/RGBotServerListener.cs
RG-nAmKcAz: actually.. in my platformer bot, i found that overriding position was a very logical thing to do.  The player model's position was centered somewhere around his chest/neck, but that was completely useless for sprite tile mapping.  

I re-mapped his position to be relative to the ground at his feet.

Thus why I was quite intentional in putting this code in such that you can override those fields.

I can see an argument for blocking everything but position/rotation from being overridden though.

RG-nAmKcAz: adding code now to block all but those 2 as i rework this section anyway

RG-nAmKcAz:
@abeizer
BotServerListener field merging and override code has been corrected/improved

@rcornwall-reg 
Prompt updated with NOTE about AOT environments

Also, fixed a missing piece of code in the generators for states that wasn't mapping the RGState to the generated RGStateEntity type for it correctly.

Runtime/Scripts/RGBotServerListener.cs
RG-nAmKcAz: code updated / resolved

Runtime/Scripts/RGBotConfigs/RGEntity.cs
RG-nAmKcAz: Definitely... as discussed in our 1-1 .. we need to update our docs site to have a little less top down tutorial style and a little more raw information/direct link style to make good use of this for Unity components

abeizer:
I believe `RGState_Button` needs to be updated - buttons are spamming me with warnings about state conflicts because the RGEntity keys are still defined on it

```csharp
        public IRGStateEntity GetGameObjectState()
        {
            var state = new RGStateEntity_Button()
            {
                ["id"] = transform.GetInstanceID(),
                ["type"] = objectType,
                ["isPlayer"] = false,
                ["isRuntimeObject"] = false,
                ["position"] = Vector3.zero,
                ["rotation"] = Vector3.zero
            };
            
            CanvasGroup cg = this.gameObject.GetComponentInParent<CanvasGroup>();
            state["interactable"] = (cg == null || cg.interactable) && this.gameObject.GetComponent<Button>().interactable;
            
            return state;
        }
```

RG-nAmKcAz:
> I believe `RGState_Button` needs to be updated - buttons are spamming me with warnings about state conflicts because the RGEntity keys are still defined on it
> 
> ```cs
>         public IRGStateEntity GetGameObjectState()
>         {
>             var state = new RGStateEntity_Button()
>             {
>                 ["id"] = transform.GetInstanceID(),
>                 ["type"] = objectType,
>                 ["isPlayer"] = false,
>                 ["isRuntimeObject"] = false,
>                 ["position"] = Vector3.zero,
>                 ["rotation"] = Vector3.zero
>             };
>             
>             CanvasGroup cg = this.gameObject.GetComponentInParent<CanvasGroup>();
>             state["interactable"] = (cg == null || cg.interactable) && this.gameObject.GetComponent<Button>().interactable;
>             
>             return state;
>         }
> ```

Very good catch.. and a Pandora's box as the buttons themselves used to set the objectType.... will come up with a solution.

rcornwall-reg (approved):

rcornwall-reg:
Approving pending `RGState_Button` abby mentioned

RG-nAmKcAz:
Just an update.. The button thing was like a pandora's box inside a pandora's box.... Will re-ask for review on that commit once I ever get it cleanly working :/

RG-nAmKcAz:
- The last 2 commits (https://github.com/Regression-Games/RGUnityBots/pull/84/commits/4e74b0d30fc0e13ad2f4e566a5839d7f2a75748b) & (https://github.com/Regression-Games/RGUnityBots/pull/84/commits/85f0c93605e8e0921ac492253c18f8bce281b210) re-imagine how we represent buttons and their click actions
  - RGEntity is now all that is needed on a button.  It automatically handles interactable and actions
  - RGAction_ClickButton now lives ONLY on the RGOverlayMenu and handles directing all button clicks to the correct/requested target
- See also updated bossroom (https://github.com/Regression-Games/RGBossRoom/pull/43)

RG-nAmKcAz:
@abeizer @rcornwall-reg while you should feel free to review the above changes....

I found in further testing that our button story for agent builder is not good.  I'm going to submit a small fix to the generator to avoid a bug I found, and have created follow up task. ([REG-1400](https://linear.app/regression-games/issue/REG-1400/make-ui-buttons-work-in-agent-builder))

abeizer (approved):

PR #87, opened 2023-11-02
[REG-1397] Make SetCharacterConfigFromJson public so user bots can call it

analogrelay:
I thought I made `RG.SetCharacterConfigFromJson` a `public` method but I guess not 😩 . Anyway, it needs to be public so that user bots can call it. Our generated Agent Builder code includes a call to `SetCharacterConfigFromJson`:

```csharp
    public class BotEntryPoint : RGBehaviorTreeBot
    {
        protected override bool GetIsSpawnable() => true;
        protected override RGBotLifecycle GetBotLifecycle() => RGBotLifecycle.MANAGED;

        private const string CharacterConfig = @"{}";

        public override void ConfigureBot(RG rgObject)
        {
            rgObject.SetCharacterConfigFromJson(CharacterConfig);
        }

        protected override RootNode BuildBehaviorTree()
        {
            var rootNode = new RootNode();
            rootNode.AddChild(new SequenceNode());

            return rootNode;
        }
    }
```

Also, Rider went ahead and cleaned up some trailing spaces 🤣 . If you turn "Ignore Whitespace" on in the files view, you'll see the only real change.

### DIFF
--- a/Runtime/Scripts/RGBotLocalRuntime/RG.cs
+++ b/Runtime/Scripts/RGBotLocalRuntime/RG.cs
@@ -12,11 +12,11 @@ namespace RegressionGames.RGBotLocalRuntime
     public class RG
     {
         public bool Completed { get; private set; } = false;
-        
-        private RGTickInfoData _tickInfo; 
-        
+
+        private RGTickInfoData _tickInfo;
+
         private readonly ConcurrentQueue<RGActionRequest> _actionQueue = new();
-        
+
         private readonly ConcurrentQueue<RGValidationResult> _validationResults = new();
 
         public Dictionary<string, object> CharacterConfig = new();
@@ -97,7 +97,7 @@ namespace RegressionGames.RGBotLocalRuntime
             if (result.Count > 1)
             {
                 var pos = position ?? GetMyPlayers()[0].position;
-                
+
                 // sort by distance
                 result.Sort((e1, e2) =>
                 {
@@ -141,7 +141,7 @@ namespace RegressionGames.RGBotLocalRuntime
             {
                 return new List<IRGStateEntity>();
             }
-            
+
             // filter down to objectType Matches
             var result = gameState.Values.Where(value =>
             {
@@ -170,7 +170,7 @@ namespace RegressionGames.RGBotLocalRuntime
 
             // filter down to isPlayer
             var result = gameState.Values.Where(value => value.isPlayer);
-            
+
             // filter down to clientId Matches
             if (clientId != null)
             {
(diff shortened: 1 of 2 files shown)


rcornwall-reg (approved):
I wonder if there's a way we can create & all share the same IDE auto formatter to prevent odd whitespaces like this

RG-nAmKcAz (changes requested):
Because we currently have this public in `RG.cs`, the Set methods for it are somewhat redundant.
`public Dictionary<string, object> CharacterConfig = new();`

Option 1:  We can make that be `internal` and have BOTH of the set methods be `public`.  NOTE: This will involve updating the existing Sample bots in RGBossRoom and any other projects with local C# unity bots.

Option 2: We remove both set methods altogether.

vontell (approved):

analogrelay:
I agree the internal Set method that takes a Dictionary is redundant, but the one that parses it from JSON is still useful since it internalizes our parsing logic and keeps the bot code from being coupled to Newtonsoft.JSON. I can remove that other internal Set method though.

analogrelay:
> I wonder if there's a way we can create & all share the same IDE auto formatter to prevent odd whitespaces like this

Rider supports having a settings file that's checked in to the repo and overrides your local user settings. I should be able to configure it with just our basic style settings (so everyone can keep their other settings how they like them). I'll check in to that.

RG-nAmKcAz:
> I agree the internal Set method that takes a Dictionary is redundant, but the one that parses it from JSON is still useful since it internalizes our parsing logic and keeps the bot code from being coupled to Newtonsoft.JSON. I can remove that other internal Set method though.

I guess I was suggesting that we put the `JsonConvert.DeserializeObject<Dictionary<string, object>>(characterConfigJson)` code into the AgentBuilder generated code itself and remove that Set from string method also.

analogrelay:
> that we put the JsonConvert.DeserializeObject<Dictionary<string, object>>(characterConfigJson) code into the AgentBuilder generated code itself and remove that Set from string method also.

I think we should isolate Agent Builder generated code from dependencies other than our SDK. If we remove Newtonsoft.JSON someday and a user updates our SDK in their game, all the generated code would break.

analogrelay:
The generated code is long-lived and while we may make breaking changes that require regenerating, it's best to avoid that when we can. I think keeping the SetFromJson is a good compromise to keep as much compatibility as possible

analogrelay:
I've made an update to remove the unnecessary `SetCharacterConfig(Dictionary<string, object>)` method. Leaving `SetCharacterConfigFromJson(string)` because I do still think there's value in it.

RG-nAmKcAz (approved):

PR #90, opened 2023-11-06
[REG-1404] Fix Windows Sample Import

rcornwall-reg:
**Changes**
* Updated Docs URL
* Fixed paths to work on both Windows and Mac

**To Test**
1. Create new URP Project on Mac
2. Import SDK
3. Hit "Continue as Guest"
4. Hit "Import Sample" and verify it opens
5. Create new URP Project on PC
6. Import SDK
7. Hit "Continue as Guest"
8. Hit "Import Sample" and verify it opens

### DIFF
--- a/Editor/Scripts/Startup/RegressionPackagePopup.cs
+++ b/Editor/Scripts/Startup/RegressionPackagePopup.cs
@@ -364,7 +364,7 @@ public class RegressionPackagePopup : EditorWindow
     {
         string packageName = "gg.regression.unity.bots";
         string sampleDirectoryName = "Samples~";
-        string assetsDirectoryName = "Assets";
+        string assetsDirectoryName = Application.dataPath;
         string destinationDirectoryName = sampleName;
 
         // Construct the path to the sample within the package
--- a/Editor/Scripts/Startup/RegressionPackagePopup.cs
+++ b/Editor/Scripts/Startup/RegressionPackagePopup.cs
@@ -363,20 +363,29 @@ public class RegressionPackagePopup : EditorWindow
     private void ImportSample(string sampleName)
     {
         string packageName = "gg.regression.unity.bots";
-        string samplePath = "Samples~/" + sampleName;
-        string destinationPath = "Assets/" + sampleName;
+        string sampleDirectoryName = "Samples~";
+        string assetsDirectoryName = "Assets";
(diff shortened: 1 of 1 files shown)


Editor/Scripts/Startup/RegressionPackagePopup.cs
```diff
@@ -360,30 +360,33 @@ private bool IsURPEnabled()
...
         string packageName = "gg.regression.unity.bots";
-        string samplePath = "Samples~/ThirdPersonDemoURP";
-        string destinationPath = "Assets/ThirdPersonDemoURP";
+        string samplePath = "Samples~/" + sampleName;
+        string destinationPath = "Assets/" + sampleName;
 
         // Construct the path to the sample within the package
-        string packagePath = Path.Combine("Packages", packageName, samplePath);
-
+        string packagePath = "Packages/" + packageName + '/' + samplePath;
```
RG-nAmKcAz, line 370: I think we should stick with using `Path.Combine` instead of assumed `/` pathing.

the reason it wasn't working before is that the first argument must be an absolute path.

something like this

```csharp
string samplePath = Path.Combine(FileUtil.GetPhysicalPath("Samples~"), "ThirdPersonDemoURP");
string destinationPath = Path.Combine(FileUtil.GetPhysicalPath("Assets"), "ThirdPersonDemoURP");
string packagePath = Path.Combine(FileUtil.GetPhysicalPath("Packages"), packageName, samplePath);

//...


EditorSceneManager.OpenScene(Path.Combine(destinationPath, "Demo", "Scenes", "Playground.unity"));
```

RG-nAmKcAz (changes requested):

vontell (approved):
Approving assuming that Zack's changes are approved. Approving so you aren't blocked once he approves.

abeizer (approved):
Approving pending Zack's comments

rcornwall-reg:
Ready for re review ✅ 
Tested on Mac and Windows ✅ 

* Used `Path.Combine(...).Replace("\\", "/");` to maintain requirement of forward slashes in directory paths for [Unity's File Utils](https://docs.unity3d.com/ScriptReference/FileUtil.CopyFileOrDirectory.html)
* `EditorSceneManager` uses full absolute name from `Application.dataPath`

RG-nAmKcAz (approved):

PR #91, opened 2023-11-07
[REG-1401] fix fallback namespace code

RG-nAmKcAz:
- In my sick/tired stupor, I missed pushing my last commit :(


vontell (approved):

analogrelay (approved):
Oof, this is a sneaky one. It happens :). It does bring up that we should at least figure out a way to run the compiler in a PR check :).

PR #92, opened 2023-11-07
[REG-1373] Populate state and action json from all scenes

RG-nAmKcAz:
- Handles looking up objectTypes from all scenes, not just the current active ones
  - Includes logic for restoring their editor scenes back to the same open/active scenes they started with, including keeping scenes original loaded/unloaded states.
- Includes prompting to warn about this as well as handle unsaved scene changes.

image

Screenshot 2023-11-07 at 11 16 06 AM

### DIFF
--- a/Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
+++ b/Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
@@ -153,11 +153,8 @@ namespace RegressionGames.Editor.CodeGenerators
                             "Cancel"
                         ))
                     {
-                        foreach (var dirtyScene in dirtyScenes)
-                        {
-                            EditorSceneManager.SaveScenes(dirtyScenes.ToArray());
-                            ExtractGameContextHelper();
-                        }
+                        EditorSceneManager.SaveScenes(dirtyScenes.ToArray());
+                        ExtractGameContextHelper();
                     }
                 }
                 else
--- a/Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
+++ b/Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
@@ -9,7 +9,6 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using RegressionGames.RGBotConfigs;
-using RegressionGames.StateActionTypes;
 using UnityEditor;
 using UnityEditor.SceneManagement;
 using UnityEngine;
--- a/Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
+++ b/Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
@@ -138,8 +138,10 @@ namespace RegressionGames.Editor.CodeGenerators
         private static void ExtractGameContext()
         {
             if (EditorUtility.DisplayDialog(
-                    "Extract Game Context",
-                    "Warning: This operation will load and unload every scene in your project's build configuration while gathering data.  This operation can take a long time to complete depending on the size of your project.",
+                    "Extract Game Context\r\nWarning",
+                    "This operation will load and unload every scene in your project's build configuration while gathering data." +
+                    "\r\n\r\nOnly Scenes that are enabled in your build configuration will be evaluated." +
+                    "\r\n\r\nThis operation can take a long time to complete depending on the size of your project.",
                     "Continue",
                     "Cancel"))
             {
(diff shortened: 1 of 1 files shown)


Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
```diff
@@ -118,8 +119,76 @@ private static void GenerateRGScripts()
...
+                            "Unsaved Changes Detected",
+                            "One or more open Scenes have unsaved changes.",
+                            "Save Scenes and Continue Extract",
+                            "Cancel"
+                        ))
+                    {
+                        foreach (var dirtyScene in dirtyScenes)
+                        {
+                            EditorSceneManager.SaveScenes(dirtyScenes.ToArray());
+                            ExtractGameContextHelper();
```
abeizer, line 159: Do we need to do call `ExtractGameContextHelper` for every scene, or can we do this once after all scenes are saved?

RG-nAmKcAz: yes, we need to call it for every scene

I start by calling for whatever (scene/scenes) they have open, then call it one by one for all other scenes as we open and close them

then.. restore their original scene state as closely as we can (ordering gets slightly changed if they had unloaded scenes in their view)

RG-nAmKcAz: oh... haha.. i forgot to take out the loop.. fixing

RG-nAmKcAz: fixed

Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
```diff
@@ -577,22 +627,71 @@ private static void PopulateObjectTypes(List<RGStatesInfo> statesInfos, List<RGA
...
+                }
+                return allLoadedScenePaths.Contains(b) ? 1 : 0;
+            });
+            
+            // get all the objects in the currently open scenes.. this minimizes the amount of scene loading we have to do
+            LookupEntitiesForCurrentScenes(objectTypeMap);
+            
+            // Get for all the other scenes in the build 
+            EditorBuildSettingsScene[] scenesInBuild = EditorBuildSettings.scenes;
+            foreach (var editorScene in scenesInBuild)
```
rcornwall-reg, line 660: If we're only searching scenes that have been added toe the build settings, we're introducing an edge case where scenes exist in the project but have not been added to build settings yet. The order of `Create Scene`->`Generate Scripts`->`Add scene to build settings` wouldn't work. Just to be 100% sure we're not missing scenes I would either 
* regenerate every time a scene is added to build settings
* search the actual directory for `.unity` files and get paths that way
* search directory for `.unity` files and compare the names to what `EditorBuildSettings.scene` and show a warning of "this scene won't be included in generated data" etc

abeizer (approved):

rcornwall-reg (changes requested):
Looking good

**[Suggestion][Non-Blocking]** 
In the future I think there _might_ be a way to do this without having to open and close unity scenes. It'd have to be verified though.
1. Search directory for `.unity` files
2. Read file data and extract GameObject data
3. Compare GUID of gameObject components with GUID of RGEntity to find all entities
4. Call the map & object type logic

RG-nAmKcAz:
@rcornwall-reg Re: Suggestion, I'm not sure how we would find objects in scene or evaluate components of game objects outside of loading the scene ?  If there is an elegant way to do it we should evaluate it, but if we'd be trusting unity file formats not to change and doing file content based actions to infer data that sounds dangerous.

Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
RG-nAmKcAz: If their scene isn't in their build settings, I wouldn't want us to be interrogating it for data as it won't be part of their runtime anyway and there may be a very good reason it is excluded; it may not even be in a working state.

I have added a sentence to the warning prompt to clarify what is included.

image

rcornwall-reg:
> @rcornwall-reg Re: Suggestion, I'm not sure how we would find objects in scene or evaluate components of game objects outside of loading the scene ? If there is an elegant way to do it we should evaluate it, but if we'd be trusting unity file formats not to change and doing file content based actions to infer data that sounds dangerous.

They have it documented here: [https://docs.unity3d.com/Manual/FormatDescription.html](https://docs.unity3d.com/Manual/FormatDescription.html) So it's a bit more stable then "let's just look and see". But the scene file format might change from LTS version to LTS version, so that'd be something we'd have to test thoroughly.

rcornwall-reg (approved):

PR #94, opened 2023-11-09
[REG-1419] Allow decorator nodes to have no children

analogrelay:
Found this while working on some remote bot syncing code. We require that a decorator node has a child in it's constructor. However, that's not required by the Agent Builder design surface. It also makes sense to have decorators without children. The `AlwaysSucceed` and `AlwaysFail` decorators can do their job regardless of whether or not they have children.

Matching RGService PR: https://github.com/Regression-Games/RegressionGames/pull/431

### DIFF
--- a/Runtime/Scripts/BehaviorTree/Decorators.cs
+++ b/Runtime/Scripts/BehaviorTree/Decorators.cs
@@ -12,6 +12,14 @@ namespace RegressionGames.BehaviorTree
         protected readonly BehaviorTreeNode Child;
 
         /// <summary>
+        /// Creates a new decorator node with the specified name and no child node.
+        /// </summary>
+        /// <param name="name">The name of the node.</param>
+        protected DecoratorNode(string name) : base(name)
+        {
+        }
+
+        /// <summary>
         /// Creates a new decorator node with the specified name and child node.
         /// </summary>
         /// <param name="name">The name of the node.</param>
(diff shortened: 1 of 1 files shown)


abeizer:
> I can't really think of a situation where you'd want to use always succeed or fail without a child, but technically having one versus not having one doesn't affect its functionality so I'm willing to give a begrudging okay. Invert I think really does need a child, because its whole deal is inverting the child's status. The UI has no validation right now, and I think that's the correct next step for invert - we need validation for other things too (nodes that need children have children, no orphaned nodes, no circular connections, all code is valid at least syntactically, etc.)

Created a linear task for client validation: https://linear.app/regression-games/issue/REG-1421/agent-builder-perform-validation-on-save

analogrelay:
Yep, I agree it's kinda silly, but I think we should try and be somewhat forgiving here until the Agent Builder editor has some more validation.

analogrelay:
> Invert I think really does need a child, because its whole deal is inverting the child's status

Fair enough (you beat me to writing the comment ;)). I'll just make it throw if it has no child for now. If we add client validation, that will be a clearer experience.

analogrelay:
I was already leaning slightly back towards throwing in Invert anyway :).

abeizer (approved):
Approve pending invert node changes 🦆

RG-nAmKcAz (approved):

vontell (approved):

PR #96, opened 2023-11-09
[REG-1398] Prompt to select local/remote when synchronizing bots

analogrelay:
Loom demo: https://www.loom.com/share/db6cfb4246e940f099939794eaa0b8d5

This changes our sync logic a bit when there's a bot that exists both locally and remote. First, we ZIP up the local bot and then compare the MD5 checksum. If they match, there's nothing to do so we move on. If they don't match, we grab timestamps for the local and remote files and present a dialog to the user asking which to keep. Based on the user's response, we either pull the bot code down (if they selected "Remote") or push the bot code up (if they selected "Local").

This is a fairly chatty experience right now, you'll be prompted for any change. I think there's some iteration to do here, but this at least removes the current blocker on using Agent Builder bots along with the sync feature (since the current logic always uploads local bots to the remote server regardless of which is newer).

Alas, there are more trailing whitespace cleanups here from Rider, so I recommend the "Ignore whitespace" option in GitHub when reviewing!


rcornwall-reg (approved):
Thanks for the loom video, looks very useful

analogrelay:
I had a realization that we could probably use the `botCreationTool` field on the Bot response itself to simplify this even further. I may open an alternate PR with that to consider. The general idea is that if `botCreationTool == AGENT_BUILDER`, there's no reason to ever sync from Local up to Remote. The behavior will never quite be right, until/unless we make major changes to the Agent Builder. The Agent Builder is driven off the `agentBuilderWorkspace.json` file. If you modify the _code_ but don't update that file, and then push it to RG, you won't see your changes in the agent builder. And when you save the bot again on the server, you'll blast away any generated code.

So, my thinking is that we can actually just _always pull_ agent builder bots from Remote to Local and never try to upload them.

Runtime/Scripts/Types/RGBot.cs
```diff
@@ -21,6 +21,17 @@ public override string ToString()
...
-}
 
+    /// <summary>
+    /// Represents a bot record that comes from the Regression Games service.
+    /// </summary>
+    /// <remarks>
+    /// Bots that come from the Regression Games service include some additional metadata that local bots do not have.
+    /// </remarks>
+    [Serializable]
+    public class RGRemoteBot : RGBot
```
RG-nAmKcAz, line 32: RGBot is in the `Types` folder/namespace... These `Types` represent the same entities that we use in RGService.  Long story short, we don't need RGRemoteBot vs RGBot as RGBot is already meant to represent those remote records.

Editor/Scripts/BotManagement/RGBotSynchronizer.cs
```diff
@@ -363,26 +410,33 @@ private void CreateParentFolders()
         private string CreateBotFolder(string botName, long botId)
         {
             CreateParentFolders();
-            var folderString = $"{BOTS_PATH}/{botName}_{botId}".Replace('-','n');
-            AssetDatabase.CreateFolder(BOTS_PATH, $"{botName}_{botId}".Replace('-','n'));
+            var botFolderName = $"{botName}_{botId}".Replace('-', 'n');
+            var folderString = $"{BOTS_PATH}/{botFolderName}";
+            if (AssetDatabase.IsValidFolder(folderString))
```
RG-nAmKcAz, line 415: I should have put a code comment here I guess when I first wrote this, but this auto delete code was intentionally excluded.

If they have a directory that matches this new bot being synced, but that directory does not have a bot asset record in it that would have allowed us to do the proper checks...  I chose not to wipe out whatever they had in that directory blindly and instead let the process throw an error so they could resolve the situation themselves and save whatever was in that directory from getting deleted.

RG-nAmKcAz (changes requested):
This looks like it will work, but I also like your suggestion of making it 'automatic' for AgentBuilder bots so they don't get prompted un-necessarily.

vontell (changes requested):

vontell:
Also, @analogrelay I do like your idea of having special logic for agent builders, would be a simple way to not have the dialog in those cases.

Runtime/Scripts/Types/RGBot.cs
analogrelay: We do also use it for local bots so I was a little hesitant to just combine them. It's probably not a big deal though. Happy to just merge these types and keep calling them RGBot for now. It may become useful to separate the local and remote metadata for a synced bot but we can do that when we have a case that necessitates it.

RG-nAmKcAz: let's definitely have it as just `RGBot` / `RGBotList`; these types are meant to represent the same record as RGService's version of `Bot` with the same fields/etc whether they are being used for a Local bot or Remote

Editor/Scripts/BotManagement/RGBotSynchronizer.cs
```diff
@@ -230,61 +244,94 @@ private async Task<List<RGBot>> GetBotsFromRG()
...
-                    // not found.. need to push for first time
-                    needsToUpload = true;
-                }
-            )!;
+            var localTime = localLastUpdated is null
+                ? "<<unknown>>" // Very unlikely, indicates the local files don't exist.
+                : localLastUpdated.Value.ToLocalTime().ToString("ddd MMM d, yyyy HH:mm:ss tt");
+            var remoteTime = botCodeDetails.modifiedDate is null
+                ? "<<unknown>>" // Very unlikely, indicates the server is misbehaving.
+                : botCodeDetails.modifiedDate.Value.ToLocalTime().ToString("ddd MMM d, yyyy HH:mm:ss tt");
```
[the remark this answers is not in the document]
analogrelay: This is definitely a case where I'd consider throwing, but since we're in the user's Unity editor I hesitated. If we had some telemetry, I'd have it log an "error"-level message but then recover by showing `<<unknown>>`.

Editor/Scripts/BotManagement/RGBotSynchronizer.cs
analogrelay: Hmm, there's probably more we have to do to handle that scenario, since `AssetDatabase.CreateFolder` won't error out if the directory does not exist. Instead, it [appends a sequence number to the end](https://docs.unity3d.com/ScriptReference/AssetDatabase.CreateFolder.html). That does preserve the existing code, but our logic for identifying Bot code on disk is highly dependent upon the path so I don't think the new bot will be properly synced with the server. It might be best to just throw an error in that case, but I'm happy to do a little bit of quick testing and file a separate ticket for handling that.

For now, I'll have a `replace` argument that flows in here from RGBotSynchronizer to make it more explicit when we **want** auto-deleting here. We _do_ want it when we're synchronizing something we know to be a remote bot, otherwise old files don't get cleaned up.

analogrelay:
Updates:
* `RGBotSynchronizer` now considers `RGBot.CodeIsServerOwned` (which is derived from `botCreationTool`) before deciding which direction to sync. If the bot code is considered "server-owned" (as it is for Agent Builder bots), then local changes are overwritten with remote code every time.
* `RGBotAsset` now has a `ChecksumAtLastSync` property. It captures the checksum of the ZIP retrieved from the server when the bot was last downloaded from the server. This isn't used for conflict resolution at this time, but should now be properly updated every time we sync a bot. Capturing this data now allows us to add conflict resolution logic later and have it light up for existing bots. For example, with an Agent Builder bot, we'd be able to detect if the user has changed the local code and warn them that we're going to overwrite those changes with changes from the server.
* `RGRemoteBot` is removed. I didn't realize at the time that `RGBotAsset` existed for us to put local bot metadata. I definitely agree it makes sense for us to have `RGBot` mirror the API response from the server.

Editor/Scripts/BotManagement/RGBotSynchronizer.cs
analogrelay: Yeah, it looks like the existing behavior for conflicting directories is a little unexpected, but that's a separate issue. I filed [REG-1430](https://linear.app/regression-games/issue/REG-1430/unity-bot-sync-behavior-is-unexpected-if-directory-exists-and-is-not-a) from my investigation.

For this PR, I'll leave that part of the behavior and make auto-delete conditional on the `replace` parameter, as I've done in the latest iteration. That will preserve the current behavior for directories that already exist but aren't bots.

Editor/Scripts/BotManagement/RGBotSynchronizer.cs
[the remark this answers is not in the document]
RG-nAmKcAz: to answer @vontell 's question.  BotCodeDetails is the metadata about the zip file itself.  Yes, it gets updated on any zip change on the server.

Runtime/Scripts/Types/RGRemoteBotList.cs
```diff
@@ -8,4 +8,3 @@ public class RGBotList
         public RGBot[] bots;
```
RG-nAmKcAz, line 8: It looks like the types got fixed back, but the files still show as being renamed

RG-nAmKcAz (changes requested):

Runtime/Scripts/Types/RGRemoteBotList.cs
analogrelay: Oop, yeah, good catch. Updating now.

RG-nAmKcAz (approved):

vontell (approved):

PR #97, opened 2023-11-14
Release 0.0.15

vontell:
We don't really have a process for release candidates for the Unity SDK, so I made this v0.0.15 branch for now. Maybe we can consider this process for next release.


RG-nAmKcAz (approved):

addisonbgross (approved):

analogrelay (approved):
I think a branch is definitely the right way to "release" the package for now. You can reference the package via a branch reference, so users can pin to it and we can hotfix it if needed.

PR #98, opened 2023-11-14
[REG-1423] Fix issues where editor libraries were preventing project builds

RG-nAmKcAz:
- Add `#if UNITY_EDITOR` to files as needed to properly exclude Editor libraries from production builds
  - Unbreaks the ability for users to actually build their projects
  
- We still can't run our bots in their production builds, but we'll fix that as a follow up task as it gets more complicated

### DIFF
--- a/Runtime/Scripts/RGBotLocalRuntime/RGBotAssetsManager.cs
+++ b/Runtime/Scripts/RGBotLocalRuntime/RGBotAssetsManager.cs
@@ -73,7 +73,7 @@ namespace RegressionGames.RGBotLocalRuntime
                 }
             }
 #else
-            //TODO: can't use asset database in real runtime
+            //TODO (REG-1424): can't use asset database in real runtime
 #endif
         }
     }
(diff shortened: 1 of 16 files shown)


Runtime/Scripts/RGBotLocalRuntime/RGBotAssetsManager.cs
```diff
@@ -71,6 +72,9 @@ public void RefreshAvailableBots()
                     RGDebug.LogWarning($"Bot at path `{botDirectory}` could not be loaded: {ex}");
                 }
             }
+#else
+            //TODO: can't use asset database in real runtime
```
vontell, line 76: Assuming a ticket will be made for this follow up exploration

vontell (approved):

abeizer (approved):

Runtime/Scripts/RGBotLocalRuntime/RGBotAssetsManager.cs
RG-nAmKcAz: it was, and was updated.. you just beat me to it :D

PR #102, opened 2023-11-21
[REG-1461] Fix bot creation and loading on Windows

RG-nAmKcAz:
- Fixes namespace generation bug introduced in (https://github.com/Regression-Games/RGUnityBots/pull/100)
- Fixes asset pathing for both Windows and Mac for bot creation and loading

### DIFF
--- a/Runtime/Scripts/RGBotLocalRuntime/RGBotAssetsManager.cs
+++ b/Runtime/Scripts/RGBotLocalRuntime/RGBotAssetsManager.cs
@@ -58,13 +58,8 @@ namespace RegressionGames.RGBotLocalRuntime
             foreach (var botGuid in botGuids)
             {
                 var botAssetPath = AssetDatabase.GUIDToAssetPath(botGuid);
-                var botDirectory = botAssetPath;
-                var lastIndex = botAssetPath.LastIndexOf(Path.DirectorySeparatorChar);
-                
-                if (lastIndex != -1) 
-                {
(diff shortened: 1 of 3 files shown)


Runtime/Scripts/RGBotLocalRuntime/RGBotAssetsManager.cs
```diff
@@ -58,13 +58,8 @@ public void RefreshAvailableBots()
...
                 var botAssetPath = AssetDatabase.GUIDToAssetPath(botGuid);
-                var botDirectory = botAssetPath;
-                var lastIndex = botAssetPath.LastIndexOf(Path.DirectorySeparatorChar);
-                
-                if (lastIndex != -1) 
-                {
-                    botDirectory = botAssetPath.Substring(0, lastIndex);        
-                }
+                // unity assets always use '/' regardless of Operating System
+                var botDirectory = botAssetPath.Substring(0, botAssetPath.LastIndexOf('/'));
```
vontell, line 62: Ah, so was the original issue that we were using the Path.DirectorySeparatorChar instead of just '/'?

vontell (approved):

analogrelay (approved):

Runtime/Scripts/RGBotLocalRuntime/RGBotAssetsManager.cs
RG-nAmKcAz: yep

PR #105, opened 2023-11-28
[REG-1478] [REG-1464] Auto generate RG SDK classes

RG-nAmKcAz:
- Auto generates RG SDK attribute classes on any code change (whenever the editor would normally recompile scripts).
  - Does not run on SDK internals changes
  - Does not run for changes to files in the generated scripts folder
- Removes the manual option for generating classes
- Updates the error handling in the code generator to better indicate a user issue (WARNING) vs an RG SDK bug (ERROR)
- Displays a prompt to the user and a log message when code generation or extracting fail
  - Addresses REG-1464
---

### DIFF
--- a/Editor/Scripts/CodeGenerators/BuildAutoProcessor.cs
+++ b/Editor/Scripts/CodeGenerators/BuildAutoProcessor.cs
@@ -1,0 +1,48 @@
+using System.Linq;
+using RegressionGames;
+#if UNITY_EDITOR
+using RegressionGames.Editor.CodeGenerators;
+using UnityEditor.Build;
+using UnityEditor.Build.Reporting;
+namespace RegressionGames.Editor.CodeGenerators
+{
+    class BuildAutoPreProcessor : IPreprocessBuildWithReport
+    {
+        public int callbackOrder { get { return 0; } }
+        
+        public void OnPreprocessBuild(BuildReport report)
+        {
+            BuildAssetPostProcessor.TriggerGeneration();
+        }
+    }
+   
+    class BuildAssetPostProcessor : UnityEditor.AssetPostprocessor
+    {
+
+        static bool _enabled = true;
+        
+        public static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAssets, string[] movedAssets,
+            string[] movedFromAssetPaths, bool didDomainReload)
+        {
+            // domain reload after build resets enabled to true
+            // and this gets called once with the original script changes, once with the generated changes, and then one more time with an empty list
+            if (_enabled && importedAssets.FirstOrDefault(ia => ia.EndsWith(".cs")) != null)
+            {
+                TriggerGeneration();
+            }
+        }
+
+        public static void TriggerGeneration()
+        {
+            // domain reload after build resets enabled to true
+            if (_enabled)
+            {
+                _enabled = false;
+                RGDebug.LogInfo("! Generating RG code due to asset change or build");
+                RGCodeGenerator.GenerateRGScripts();
+                
+            }
+        }
+    }
+}
+#endif
--- a/Editor/Scripts/CodeGenerators/BuildAutoProcessor.cs
+++ b/Editor/Scripts/CodeGenerators/BuildAutoProcessor.cs
@@ -26,7 +26,7 @@ namespace RegressionGames.Editor.CodeGenerators
         {
             // domain reload after build resets enabled to true
             // and this gets called once with the original script changes, once with the generated changes, and then one more time with an empty list
-            if (_enabled && importedAssets.FirstOrDefault(ia => ia.EndsWith(".cs")) != null)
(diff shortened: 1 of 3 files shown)


vontell:
Just passing by - love this idea, this would be a HUGE improvement to efficiency and our documentation!

analogrelay:
This looks good so far! If you feel inclined to, you could probably easily add some logic to pop up a dialog when there are errors and close out [REG-1464](https://linear.app/regression-games/issue/REG-1464/show-a-dialog-when-code-generation-fails) while you're doing this. It's totally fine if you'd prefer to defer that to a separate PR though.

Editor/Scripts/CodeGenerators/BuildAutoProcessor.cs
```diff
@@ -0,0 +1,47 @@
...
+                TriggerGeneration();
+            }
+        }
+
+        public static void TriggerGeneration()
+        {
+            // domain reload after build resets enabled to true
+            if (_enabled)
+            {
+                _enabled = false;
```
vontell, line 40: Why does this change to false here?

Editor/Scripts/CodeGenerators/BuildAutoProcessor.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+
+        static bool _enabled = true;
+        
+        public static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAssets, string[] movedAssets,
+            string[] movedFromAssetPaths, bool didDomainReload)
+        {
+            // domain reload after build resets enabled to true
+            // and this gets called once with the original script changes, once with the generated changes, and then one more time with an empty list
+            if (_enabled && importedAssets.FirstOrDefault(ia => ia.EndsWith(".cs") && !ia.StartsWith("Packages/gg.regression.unity.bots") && !ia.Contains("RegressionGames/Runtime/GeneratedScripts")) != null)
```
vontell, line 29: I remember that you provided feedback that the concept in this PR caused the generation to happen too often and was a nuisance - is that still an issue? If I am working on random files in my project unrelated to projects, will generation occur?

vontell (changes requested):

Editor/Scripts/CodeGenerators/BuildAutoProcessor.cs
RG-nAmKcAz: It is far more reasonable with these new filters added.  I did a PoC of scanning the changes file(s) sources for the presence of [RGState] or [RGAction] strings, but ...
- That didn't scale up well with many files changed
- It didn't catch removals

IOW.. for now, this is as good as it gets.  We can filter out more in the future if necessary.

analogrelay: I think that makes sense. The App Domain will have to reload any time Unity recompiles the scripts and that will indeed reset `_enabled` to `true`.

Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
```diff
@@ -365,25 +419,25 @@ private static List<RGStateAttributesInfo> SearchForBotStateAttributes()
                         {
                             if (!fieldDeclaration.Modifiers.Any(SyntaxKind.PublicKeyword))
                             {
-                                RGDebug.LogError($"Error: Field '{fieldDeclaration.Declaration.Variables.First().Identifier.ValueText}' in class '{className}' is not public.");
+                                RecordError($"Error: Field '{fieldDeclaration.Declaration.Variables.First().Identifier.ValueText}' in class '{className}' is not public.");
```
analogrelay, line 422: (Non-blocking, since I'm happy to leave it to you to decide if it's worth it and I don't think it's worth a separate ticket for now)

Maybe give `RecordError`/`RecordWarning` an optional `SyntaxNode` parameter, and if it's provided, you can use `node.Span` to get the position in the file and `node.SyntaxTree.FilePath` to get the path to the file. Then you could include `[filename]:[line number]` in the log message. That would get us as close as possible to "properly" integrating as compiler diagnostics without having to do all the work required to do that (which is absolutely out of scope now).

analogrelay (approved):

vontell (approved):

rcornwall-reg (changes requested):
Functionally it's looking good, but there was a reason we avoided doing this in our initial implementation and that was fear of clogging the developer's compile time with generating potentially hundreds of scripts.

1. Have we profiled this and determined the generate time is small enough to do this every compile?
2. We were discussing having an "Auto Generate Scripts" checkbox in RGSettings that would disable this functionality, and just have it be checked by default. Could that be a follow up task?

vontell:
> Functionally it's looking good, but there was a reason we avoided doing this in our initial implementation and that was fear of clogging the developer's compile time with generating potentially hundreds of scripts.
> 
> 1. Have we profiled this and determined the generate time is small enough to do this every compile?
> 2. We were discussing having an "Auto Generate Scripts" checkbox in RGSettings that would disable this functionality, and just have it be checked by default. Could that be a follow up task?

@rcornwall-reg you may have not seen this because the comment was resolved and hidden, but we discuss some of this here: https://github.com/Regression-Games/RGUnityBots/pull/105#discussion_r1415820327

vontell:
That being said, if we do find this is too detrimental, we can always take another look at it!

PR #107, opened 2023-11-28
[REG-1425] Handle replay data / screenshot / validation disk buffering

RG-nAmKcAz:
- Asynchronously writes replay data, validations, screenshots, etc to disk on each tick to avoid filling up memory with the data
  - Tracks validation summary in a running fashion for each client in memory
- Uploads the data from the files directly without fully reading into memory first

---

### DIFF
--- a/Runtime/Scripts/DataCollection/RGDataCollection.cs
+++ b/Runtime/Scripts/DataCollection/RGDataCollection.cs
@@ -6,6 +6,7 @@ using System.IO.Compression;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Newtonsoft.Json;
 using RegressionGames.StateActionTypes;
 using RegressionGames.Types;
 using UnityEngine;
@@ -39,11 +40,12 @@ namespace RegressionGames.DataCollection
         private readonly string _sessionName = Guid.NewGuid().ToString();
         private readonly Dictionary<long, RGBot> _clientIdToBots = new();
         private readonly Dictionary<long, DateTime> _clientIdStartTimes = new();
-        private readonly ConcurrentDictionary<long, List<RGStateActionReplayData>> _clientIdToReplayData = new();
+        private readonly ConcurrentDictionary<long, List<Task>> _clientIdToReplayDataTasks = new();
+        private readonly ConcurrentDictionary<long, List<Task>> _clientIdToValidationDataTasks = new();
         private readonly string _rootPath = Application.persistentDataPath;
         private readonly ConcurrentQueue<long> _screenshotTicksRequested = new();
         private readonly ConcurrentDictionary<long, long> _screenshotTicksCaptured = new();
-        
+        private readonly ConcurrentDictionary<long, RGValidationSummary> _clientIdToValidationSummaries = new();
 
         /**
          * Registers a bot instance under a specific client id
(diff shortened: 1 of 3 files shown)


Runtime/Scripts/RGBotLocalRuntime/RGBotRuntimeManager.cs
```diff
@@ -93,7 +93,7 @@ public long StartBot(long botId)
                 RGClientConnection connection = RGBotServerListener.GetInstance()
                     .AddClientConnectionForBotInstance(botInstance.id, RGClientConnectionType.LOCAL);
                 
-                // Also attack the bot so we can upload later
+                // Also attach the bot so we can upload later
```
analogrelay, line 96: ⚔️

Runtime/Scripts/DataCollection/RGDataCollection.cs
```diff
@@ -106,26 +108,60 @@ public void SaveReplayDataInfo(long clientId, RGStateActionReplayData replayData
             RGDebug.LogVerbose($"DataCollection[{clientId}] - Saving replay data for tick {replayData.tickInfo.tick}");
             
             // Add the new replay data to a new or existing mapping in our client replay data dictionary
-            var replayDatas = _clientIdToReplayData.GetOrAdd(clientId, new List<RGStateActionReplayData>());
-            replayDatas.Add(replayData);
+            var replayDataTasks = _clientIdToReplayDataTasks.GetOrAdd(clientId, v => new List<Task>());
+            var validationDataTasks = _clientIdToValidationDataTasks.GetOrAdd(clientId, v => new List<Task>());
```
analogrelay, line 112: Are we guaranteed that this isn't being called simultaneously for the _same `clientId` value_? The `List<Task>` you're working with isn't thread-safe, but that should be alright as long as we're only processing one `SaveReplayDataInfo` for each `clientId` at a time.

Runtime/Scripts/DataCollection/RGDataCollection.cs
```diff
@@ -106,26 +108,60 @@ public void SaveReplayDataInfo(long clientId, RGStateActionReplayData replayData
...
+                }
+
+                // Combine the JSON strings with newline characters
+                string jsonLinesString = string.Join("\n", jsonLines) + "\n";
+                
+                //Write out the validations to the JSONL file
+                var validationFilePath =
+                    GetSessionDirectory($"validationData/{clientId}/rgbot_validations.jsonl");
+                var validationTask = File.AppendAllTextAsync(validationFilePath, jsonLinesString);
+                validationDataTasks.Add(validationTask);
```
analogrelay, line 148: This doesn't provide a guarantee of ordering. If tick 1 calls `SaveReplayData` and queues a write to `rgbot_validations.jsonl` and tick 2 does the same, there's no guarantee that tick 1's write finishes before tick 2's write starts. The `AppendAllTextAsync` method opens the file and locks it exclusively, so a simultaneous call to `AppendAllTextAsync` for the same file would throw because it can't acquire a lock on the file. If the writes can complete before the next tick you'd be OK, but we shouldn't really rely on it.

If you want to ensure ordering, what you need is a "Task Queue". Basically, you build up a chain of `.ContinueWith` calls that ensure that task `N` doesn't **start** until task `N-1` completes . It's written something like this:

```csharp
private object _lock = new object();
private Task _activeTask = Task.CompletedTask;

public void QueueSomething()
{
    async Task DoTheWork()
    {
        // This actually does the work.
        // Because of the queue logic, it's guaranteed to complete fully before it gets called again.
    }

    // If you're confident QueueSomething can't be called simultaneously, you can omit the lock
    // Problems arise if two threads are messing with _activeTask at once.
    lock(_lock)
    {
        // Note that we _replace_ "_activeTask"!
        // This means that _activeTask now represents the completion of _all_ queued work so far
        // So, when the next call to QueueSomething comes in, it's not going to _start_ it's work until the entire queue up to that point completes.
        _activeTask = _activeTask.ContinueWith(_ => DoTheWork());
    }
}
```

That's a rough version. It'll have a bunch of allocations of `Task`s though, so it might not be the best approach.

Another approach would be to enqueue the write request with all the validations to a static `ConcurrentQueue` somewhere and have a single background thread grab write requests in order and flush them to disk.

analogrelay: Another solution would be to buffer the validations in separate files for each tick. Then you'd end up in the same place as replay data, which doesn't have conflicting writes to the same file, but does still produce a lot of async work to be tracked and waited for.

analogrelay (changes requested):
Looks mostly good, but I do see a race condition that may result in errors appending validations to the file.

Runtime/Scripts/DataCollection/RGDataCollection.cs
RG-nAmKcAz: Good callout, but yes we are guaranteed not to call this in parallel for the same clientId.  This is called from each client botrunner thread in a manner that blocks said thread until this completes.

Runtime/Scripts/DataCollection/RGDataCollection.cs
RG-nAmKcAz: eesh.. it looks like our `unity.js` version of this code probably has a similar issue

For this C# case, I'm going to go with a simple solution for now.

The code will wait for the prior write to finish on the next call.  Normally that async write will complete before the next tick comes in, so this shouldn't hold up our client processing.  If we find that we're having issues holding up the client tick processing with high tick rates or other cases we can pursue a solution more in line with your proposal above.

Runtime/Scripts/DataCollection/RGDataCollection.cs
```diff
@@ -101,32 +103,70 @@ public void ProcessScreenshotRequests()
...
+                string jsonLinesString = string.Join("\n", jsonLines) + "\n";
+                
+                // Wait for any prior validation write task to finish
+                if (_clientIdToValidationDataTask.TryRemove(clientId, out var validationDataTask))
+                {
+                    await validationDataTask;
+                }
+
+                // Write out the validations to the JSONL file; note.. the prior write mush finish before the next tick
+                // otherwise we can't write to the file safely
```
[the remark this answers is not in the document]
vontell: Ah I see @analogrelay may have left a similar review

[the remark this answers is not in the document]
RG-nAmKcAz: see about 6 lines above your question where it waits for the prior one to finish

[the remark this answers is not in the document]
vontell: But doesn't await just wait for the task to finish, nothing to say that the tasks are still running in parallel? Maybe this is a C# thing I'm not familiar with (im thinking mostly in Python), but don't the tasks get started right when you create them? In other words, by the time we get to that await line, you may be waiting for one task to finish, but other tasks can still be running at the same time?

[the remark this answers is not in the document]
vontell: Oh... I'm reading the wrong task. I thought this was a parallel task to write into JSONL for each validation, ignore what I am saying!

vontell (approved):

addisonbgross (approved):
some smooth async work 🧊

RG-nAmKcAz:
@analogrelay Notifying you that this is waiting on your second review/approval

PR #108, opened 2023-12-04
[RG-1489] Use 'RGDebug.LogException' as required

rcornwall-reg:
**Changes**
Cases where we had:
```
catch (Exception e) {
   RGDebug.LogError($"Some Error Message - {e});
}
```
have been updated to:
```
catch (Exception e) {
   RGDebug.LogException(e, "Some Error Message");
}
```
and `RGDebug.LogException` has been updated to include an option string message that will log as a warning

This maintains clickable callbacks

---

### DIFF
--- a/Editor/Scripts/Startup/RegressionPackagePopup.cs
+++ b/Editor/Scripts/Startup/RegressionPackagePopup.cs
@@ -396,7 +396,8 @@ public class RegressionPackagePopup : EditorWindow
             }
             catch (Exception e)
             {
-                RGDebug.LogError("Failed to import sample: " + e.Message);
+                RGDebug.LogError("Failed to import sample");
+                RGDebug.LogException(e);
             }
         }
         else
--- a/Editor/Scripts/Startup/RegressionPackagePopup.cs
+++ b/Editor/Scripts/Startup/RegressionPackagePopup.cs
@@ -396,8 +396,7 @@ public class RegressionPackagePopup : EditorWindow
             }
             catch (Exception e)
             {
-                RGDebug.LogError("Failed to import sample");
-                RGDebug.LogException(e);
+                RGDebug.LogException(e, "Failed to import sample");
             }
         }
         else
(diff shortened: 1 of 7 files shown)


analogrelay (approved):
LGTM! This will create two log messages for each error but that should be fine. If it becomes annoying later, we can find other solutions then.

rcornwall-reg:
> LGTM! This will create two log messages for each error but that should be fine. If it becomes annoying later, we can find other solutions then.

I was hoping for a 
```
} catch (Exception e) {
  string customMessage = "Custom Error Message" + e.Message;
  e.Message = customMessage;
  RGDebug.LogException(e);
}
```
but alas, no 'Message' setters were found

analogrelay:
Yeah, you should treat `Exception`s as immutable. We could wrap the incoming exception in a new one (with the old one passed in the `innerException` constructor arg), but that might mess with stack traces. The stack trace for an `Exception` is kinda magic and only appears if you `throw` it.

What would be ideal is if `Debug.LogException` had a variant that took a `string` message _and_ an `Exception`, but alas, it does not.

RG-nAmKcAz (changes requested):
Let's either drop out the RGDebug.LogError lines or change them to RGDebug.LogWarning

When running Unity with Error Pause enabled, my project pauses twice for each of these conditions which is super annoying.

Editor/Scripts/Startup/RegressionPackagePopup.cs
```diff
@@ -396,7 +396,8 @@ private void ImportSample(string sampleName)
             }
             catch (Exception e)
             {
-                RGDebug.LogError("Failed to import sample: " + e.Message);
+                RGDebug.LogError("Failed to import sample");
+                RGDebug.LogException(e);
```
vontell, line 400: Since this is our own function, can't we have LogException in RGDebug take a message? Then we'd only have to do
```csharp
RGDebug.LogException("My message", e);
```
(this is a non-blocking idea)

vontell (approved):
Approving assuming Zack's comment is resolved

Editor/Scripts/Startup/RegressionPackagePopup.cs
rcornwall-reg: Added 👍  It's a good idea to prevent duplicate 'RGDebug' statements throughout the code. I added an option 'message' param in 'LogException'

RG-nAmKcAz (approved):

PR #109, opened 2023-12-04
[REG-1473] Collect logs during bot execution and upload

analogrelay:
This adds support to the Unity SDK to collect logs during bot execution. There are a few limitations in this first version, which lead to some open questions.

Log collection works as follows:
1. Prior to any bot starting, the `RGDataCollection` constructor hooks in to `Application.logMessageReceivedThreaded`, which is called when a log message is emitted **on any thread**.
2. When a bot starts, `RGDataCollection.RegisterBot` is called. This adds a `ConcurrentQueue<LogDataPoint>` to the `ConcurrentDictionary<long, ConcurrentQueue<LogDataPoint>>` tracking which bots are running and listening to logs. This starts collection of logs for this bot.
3. Each tick of the bot, we perform an atomic swap of the active log message queue with a new empty queue. This allows us to flush that set of messages to disk while simultaneously giving the collector a fresh queue to write to. The `ConcurrentQueue` collection does not provide an atomic way to do this swap internally, so we need to do it ourselves.
4. When a bot stops, we flush the last set of messages to disk. Then we upload the logs and the on-disk copy gets cleaned up.

The code to collect log messages is always running, but unless a bot is running, it will basically no-op. The presence of a queue in the `_clientIdToLogs` dictionary is what tells our collection logic to assign a log message that arrives to that bot instance.

In order to simplify some things, **and** enable the atomic swap in (3) above, I bundled all the dictionaries of `clientId -> Something` we had into a single dictionary of `clientId -> BotInstanceState` where `BotInstanceState` is a class that wraps up all the individual pieces we have. This allows us to only have to mess with the concurrent dictionary once as long as we are careful to access `BotInstanceState` in thread-safe ways.

And once again, Rider is aggressively removing whitespace. I recommend the "Hide whitespace" option in the PR review settings:

image

These are the settings I have in Rider that cause the reformat and I highly recommend we use them, as they ensure that we clean up trailing spaces and lines:

image

---

### DIFF
--- a/Runtime/Scripts/DataCollection/RGDataCollection.cs
+++ b/Runtime/Scripts/DataCollection/RGDataCollection.cs
@@ -11,7 +11,9 @@ using JetBrains.Annotations;
 using Newtonsoft.Json;
 using RegressionGames.StateActionTypes;
 using RegressionGames.Types;
+using Unity.Profiling;
 using UnityEngine;
+using UnityEngine.Profiling;
 using CompressionLevel = System.IO.Compression.CompressionLevel;
 using Object = UnityEngine.Object;
 using Random = UnityEngine.Random;
@@ -39,14 +41,16 @@ namespace RegressionGames.DataCollection
      */
     public class RGDataCollection
     {
+        private static readonly ProfilerMarker LogCollectionMarker = new(ProfilerCategory.Ai, "RegressionGames.DataCollection.LogCollector");
+        private static readonly ProfilerMarker SaveTickDataMarker = new(ProfilerCategory.Ai, "RegressionGames.DataCollection.SaveTickData");
 
         private readonly string _sessionName = Guid.NewGuid().ToString();
-        private readonly ConcurrentDictionary<long, BotInstanceState> _clientIdToState = new();
+        private readonly ConcurrentDictionary<long, BotInstanceDataCollectionState> _clientIdToState = new();
         private readonly string _rootPath = Application.persistentDataPath;
         private readonly ConcurrentQueue<long> _screenshotTicksRequested = new();
         private readonly ConcurrentDictionary<long, long> _screenshotTicksCaptured = new();
 
-        private class BotInstanceState
+        private class BotInstanceDataCollectionState
         {
             public long clientId;
             public RGBot bot;
@@ -57,17 +61,31 @@ namespace RegressionGames.DataCollection
             public Task logFlushTask;
             public RGValidationSummary validationSummary = new(0,0,0);
 
-            public BotInstanceState(RGBot bot, DateTime startTime)
+            public BotInstanceDataCollectionState(RGBot bot, DateTime startTime)
             {
                 this.bot = bot;
                 this.startTime = startTime;
             }
-        }
 
-        public RGDataCollection()
-        {
-            // Start collecting logs
-            Application.logMessageReceivedThreaded += LogMessageReceived;
+            public void StartCapturingLogs()
+            {
+                Application.logMessageReceivedThreaded += LogMessageReceived;
+            }
+
+            public void StopCapturingLogs()
+            {
+                Application.logMessageReceivedThreaded -= LogMessageReceived;
+            }
+
+            private void LogMessageReceived(string message, string stacktrace, LogType type)
+            {
+                Profiler.BeginSample("RegressionGames.DataCollection.LogCollector");
+                // PERF: DateTimeOffset.Now is suprisingly costly, use DateTimeOffset.UtcNow unless we have to know the local timezone.
+                // It's still good to use a DateTimeOffset here though because it ensures that when we serialize the data, it is clearly marked as a UTC time (with the `Z` suffix)
+                var dataPoint = new LogDataPoint(DateTimeOffset.UtcNow, type, message, stacktrace);
+                logs.Enqueue(dataPoint);
+                Profiler.EndSample();
+            }
         }
 
         /**
@@ -77,7 +95,9 @@ namespace RegressionGames.DataCollection
         public void RegisterBot(long clientId, RGBot bot)
         {
             RGDebug.LogVerbose($"DataCollection[{clientId}] - Registering client for bot {bot.id}");
-            _clientIdToState[clientId] = new(bot, DateTime.Now);
+            var state = new BotInstanceDataCollectionState(bot, DateTime.Now);
+            _clientIdToState[clientId] = state;
+            state.StartCapturingLogs();
         }
 
         /**
@@ -136,6 +156,8 @@ namespace RegressionGames.DataCollection
                 RGDebug.LogError("Attempted to save replay data for a stopped bot");
                 return;
             }
+
+            using var _ = SaveTickDataMarker.Auto();
 
             // Add the new replay data to a new or existing mapping in our client replay data dictionary
             var filePath =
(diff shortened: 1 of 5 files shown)


Runtime/Scripts/DataCollection/RGDataCollection.cs
```diff
@@ -124,135 +152,150 @@ public async Task SaveReplayDataInfo(long clientId, RGStateActionReplayData repl
...
-            replayDataTasks.RemoveAll(v => v.IsCompleted);
+            state.replayDataTasks.RemoveAll(v => v.IsCompleted);
 
         }
 
         public async Task SaveBotInstanceHistory(long clientId)
         {
-            try
+            // Removing the state from the dictionary will prevent any further data from being saved
+            var state = _clientIdToState.TryRemove(clientId, out var s) ? s : null;
```
analogrelay, line 192: This is intentionally done outside the `try` because there's no reason it _should_ throw and the `Cleanup` in the `finally` block assumes it's been done already.

Runtime/Scripts/DataCollection/RGDataCollection.cs
```diff
@@ -124,135 +152,150 @@ public async Task SaveReplayDataInfo(long clientId, RGStateActionReplayData repl
...
+            {
                 RGDebug.LogVerbose($"DataCollection[{clientId}] - Starting to save bot instance history");
 
-                var bot = _clientIdToBots[clientId];
+                var serviceManager = RGServiceManager.GetInstance();
+                if (serviceManager == null)
+                {
+                    // No point in continuing if there's no service manager
+                    return;
+                }
```
analogrelay, line 209: This saves a bunch of `?.` calls

Runtime/Scripts/DataCollection/RGDataCollection.cs
```diff
@@ -298,14 +341,22 @@ await RGServiceManager.GetInstance()
             {
                 Cleanup(clientId);
                 // If there are no more bots, cleanup everything else
-                if (_clientIdToBots.Count == 0)
+                if (_clientIdToState.IsEmpty)
```
analogrelay, line 344: IsEmpty is _slightly_ optimized in the case where the dictionary is not empty. `Count` has to acquire all the locks in the dictionary, whereas `IsEmpty` can do a fast lock-free check to see if there are any elements and return immediately if there are.

Runtime/Scripts/DataCollection/RGDataCollection.cs
```diff
@@ -332,18 +383,47 @@ public void Cleanup()
...
+
+            // Note: Enumerating a concurrent dictionary is _safe_ in that it won't fail,
+            // but it doesn't take a snapshot of the dictionary, so entries may be added/removed _during_ enumeration.
+            // That should be fine for us.
+            // The main edge case here is a bot stopping _during_ this enumeration,
+            // which would cause us to add a log entry to the queue after it's been flushed to the server.
+            // That's fine, it just means we'll miss that log entry. The queue will still get garbage collected.
+            foreach (var (_, state) in _clientIdToState)
+            {
+                state.logs.Enqueue(dataPoint);
```
RG-nAmKcAz, line 422: Can you do some testing in the unity profiler and check how many ms this is taking.

When micro-optimizing the Platformer bot libraries I found that in cases like this is was actually more efficient to lock on the clientId specific object outside the loop and use a non concurrent collection type.  The overhead of the concurrent collection for all the adds took much more frametime than the lock operation.

Where I saw that the worst was on ConcurrentDictionary, but that's why I ask if you could check this in the profiler as even a single ms reduction in frame time is huge.  Our SDK should strive to take as close to zero time as possible so that the game developers don't have to worry about framerate because of us.

RG-nAmKcAz (approved):

vontell (approved):
Assuming that Zack's comment gets addressed, LGTM!

Runtime/Scripts/DataCollection/RGDataCollection.cs
analogrelay: Yeah, that's a good point. ConcurrentDictionary is actually quite a lock-heavy data structure. I'll take a look in the profiler. I also have an idea for how to remove the dictionary in this case, by having each bot set up it's own handler on `logMessageReceivedThreaded` (since events can have multiple handlers, which can be added/removed at will) and have that handler capture the "state" object in a closure.

Runtime/Scripts/DataCollection/RGDataCollection.cs
```diff
@@ -32,20 +37,56 @@ namespace RegressionGames.DataCollection
...
+            {
+                Application.logMessageReceivedThreaded += LogMessageReceived;
+            }
+
+            public void StopCapturingLogs()
+            {
+                Application.logMessageReceivedThreaded -= LogMessageReceived;
+            }
+
+            private void LogMessageReceived(string message, string stacktrace, LogType type)
```
analogrelay, line 80: Update from the previous iteration here.

Previously, we used a single `logMessageReceivedThreaded` handler and iterated through all the active bots to stash the log message in the queue. I did some profiling and when logging 100 messages per tick, **and** enabling "Deep Tracing" (aka Instrumentation, where each function call is measured and counted), I did see some impact from this. So, I tried a refactoring that I had been considering before profiling and it did yield a _slightly_ noticeable improvement. It's worth noting that I had to really dig deep to find this call in the profile at all, so I'm not sure it'll be a big deal either way, but I went with the "faster" method anyway.

Basically, we add a separate handler for `logMessageReceivedThreaded` for each bot. This means that when a log message arrives, the built-in .NET event dispatching logic runs and calls each handler in sequence. Since this handler holds a reference to the `BotInstanceDataCollectionState` instance on which it is defined, it has immediate access to the `logs` queue, without having to look it up in the lock-heavy ConcurrentDictionary structure. This keeps thread contention to a minimum.

The trade-off is that we have to allocate a `LogDataPoint` for each bot in this approach. I didn't see a measurable impact from that, but we can always optimize this further later. We have a lot more allocations throughout the data collection process that could be optimized away, so there's plenty of fodder for a future optimization pass through here to speed things up more.

Runtime/Scripts/DataCollection/RGDataCollection.cs
```diff
@@ -32,20 +37,56 @@ namespace RegressionGames.DataCollection
...
+            {
+                Profiler.BeginSample("RegressionGames.DataCollection.LogCollector");
```
vontell, line 82: This sampling needs to be removed, correct?

vontell (changes requested):

Runtime/Scripts/DataCollection/RGDataCollection.cs
analogrelay: It doesn't have to be. It's compiled out of non-development builds and it can be useful to allow users to see the impact we have

vontell (approved):

analogrelay:
No meaningful changes in the latest commit, just merging `main` to resolve a trivial merge conflict after #108 landed.

analogrelay:
Going to do a final validation against dev now that the API has landed, then merge.

analogrelay:
☝🏻 The most recent commit removed a change that snuck in from work I was doing to optimize JSON serialization per tick. I ended up splitting that work off from this branch, but incompletely (Really looking forward to getting a PR build in place 😩 ).

analogrelay:
Validated end-to-end against dev 🎉 . Going to merge this now since the only change since approval was a merge and a typo fix (removed reference to `RGSerializer` type I was looking at to optimize some serialization stuff, coming in a later PR :))

PR #110, opened 2023-12-05
[REG-1508] Support RGState on class properties

RG-nAmKcAz:
- Allows using [RGState] on class properties like 
```csharp
        [RGState]
        public InventorySystem Inventory => m_Inventory;
```

---

### DIFF
--- a/Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
+++ b/Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
@@ -387,36 +387,78 @@ namespace RegressionGames.Editor.CodeGenerators
                                 hasError = true;
                             }
                         }
+                        else if (member is DelegateDeclarationSyntax delegateDeclaration)
+                        {
+                            if (!delegateDeclaration.Modifiers.Any(SyntaxKind.PublicKeyword))
+                            {
+                                RGDebug.LogError($"Error: Delegate '{delegateDeclaration.Identifier.ValueText}' in class '{className}' is not public.");
+                                hasError = true;
+                            }
+                            else if (delegateDeclaration.ParameterList.Parameters.Count > 0)
+                            {
+                                RGDebug.LogError($"Error: Delegate '{delegateDeclaration.Identifier.ValueText}' in class '{className}' has parameters, which is not allowed.");
+                                hasError = true;
+                            }
+                            else if (delegateDeclaration.ReturnType is PredefinedTypeSyntax predefinedType && predefinedType.Keyword.IsKind(SyntaxKind.VoidKeyword))
+                            {
+                                RGDebug.LogError($"Error: Delegate '{delegateDeclaration.Identifier.ValueText}' in class '{className}' has a void return type, which is not allowed.");
+                                hasError = true;
+                            }
+                        }
+                        else if (member is PropertyDeclarationSyntax propertyDeclaration)
+                        {
+                            if (!propertyDeclaration.Modifiers.Any(SyntaxKind.PublicKeyword))
+                            {
+                                RGDebug.LogError(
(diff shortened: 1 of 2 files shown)


vontell (approved):

Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
```diff
@@ -394,9 +394,30 @@ private static List<RGStateAttributesInfo> SearchForBotStateAttributes()
...
+                                fieldName = ((FieldDeclarationSyntax)member).Declaration.Variables.First().Identifier.ValueText;
+                                type = RemoveGlobalPrefix(semanticModel.GetTypeInfo(((FieldDeclarationSyntax) member).Declaration.Type)
+                                    .Type.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat));
+                                break;
+                            case PropertyDeclarationSyntax:
+                                fieldName = ((PropertyDeclarationSyntax)member).Identifier.ValueText;
+                                type = RemoveGlobalPrefix(semanticModel
+                                    .GetTypeInfo(((PropertyDeclarationSyntax)member).Type)
+                                    .Type.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat));
+                                break;
```
analogrelay, line 419: A few comments here.

1. The `switch` statement can handle the cast for you while it's performing the type-check and save having to cast `member` in the body of each block. You just have to give a name to the variable that should be bound to the casted value. Given `case MethodDeclarationSyntax method`, the `method` variable holds the casted value and is in scope for duration of the `case` body.
2. The `default` clause being at the top is a little jarring to me, I don't expect it there. Functionally it works, but I completely missed it in my first review because I don't expect to see it anywhere but in the final clause.
3. If we only support a `MethodDeclarationSyntax`, `FieldDeclarationSyntax`, or `PropertyDeclarationSyntax`, then I'd expect the `default` clause to fail somehow. In fact, if control ever does enter the top block via the `default` label, the cast to `MethodDeclarationSyntax` will fail and throw.

Here's my suggested update that factors these issues in (I _think_ I updated the syntax correctly but didn't do this in a proper C# IDE so I might have missed a `(` somewhere):

```suggestion
                            case MethodDeclarationSyntax method:
                                fieldName = method.Identifier.ValueText;
                                type = RemoveGlobalPrefix(semanticModel
                                    .GetTypeInfo(method.ReturnType)
                                    .Type.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat));
                                break;
                            case FieldDeclarationSyntax field:
                                fieldName = field.Declaration.Variables.First().Identifier.ValueText;
                                type = RemoveGlobalPrefix(semanticModel.GetTypeInfo(field.Declaration.Type)
                                    .Type.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat));
                                break;
                            case PropertyDeclarationSyntax property:
                                fieldName = property.Identifier.ValueText;
                                type = RemoveGlobalPrefix(semanticModel
                                    .GetTypeInfo(property.Type)
                                    .Type.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat));
                                break;
                            default:
                                RGDebug.LogWarning($"RGState applied to invalid declaration: {member}");
                                continue;
```

analogrelay (changes requested):

vontell:
Huh... sorry about that (re bitbuilder), it shouldn't be replying here I think

rcornwall-reg (approved):
Approving given Ashley's suggestions are addressed

Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
analogrelay: I'll take this offline, so consider this non-blocking, but I don't quite understand the delegate support.

analogrelay (approved):

PR #115, opened 2023-12-06
[REG-1488] Prompt for new bot name and decouple local bots from their IDs

analogrelay:
This PR decouples a local bot from it's ID as much as possible. The ID remains in the `BotRecord.asset` file, but does not need to be in the folder path or namespace. If you _do_ have an existing bot where the ID is in the folder path and namespace, it will sync just fine, but newly created bots and newly synced bots will not have the ID.

### An overview of the changes

* Bot launching now happens by locating the `BotRecord.asset` file, finding the `BotEntryPoint.cs` next to it and loading that script by path. We don't need to know the namespace of the BotEntryPoint type anymore
* Creating a new bot prompts you for the name _and path_ (defaulting to our existing Bots directory) and we create the bot where you ask us to
* Once the bot is created, we prompt you to open the BotEntryPoint.cs file if you want
* When syncing the bot, we mark `BotRecord.asset` dirty so that the ID gets persisted to disk immediately
* When syncing bots that don't already exist locally, we put them in our normal Bots folder. You can move them and the sync will continue to work in their new location. If a bot already exists at the default bot path, we let Unity deduplicate the folder name (`MyBot 1`, `MyBot 2`, etc.).
* We aren't even coupled to the `BotRecord.asset` file name, we now have the option of using a different file name if we'd like, though I haven't done that here.

### Loom Demo

[Bot Asset Management Changes - Watch Video](https://www.loom.com/share/439787ad18a94097a868299c7020b6a2)

---


Editor/Scripts/BotManagement/RGBotSynchronizer.cs
```diff
@@ -38,13 +38,47 @@ private static string ZipPathForBot(long botId)
         [MenuItem("Regression Games/Create New Bot")]
         private static void CreateNewBot()
         {
-            var botId = RGSettings.GetOrCreateSettings().GetNextBotId();
+            // Create our bot folder if it doesn't exist
+            RGEditorUtils.CreateAllAssetFolders(BOTS_PATH);
+
+            // prompt the user to choose a folder
+            var newBotPath = EditorUtility.SaveFilePanelInProject(
```
RG-nAmKcAz, line 45: If I already have `MyCoolBot` and I try to create a second `MyCoolBot` using the menu, where does it handle it ?  Does this file panel itself stop you, or do we have to handle it in code somewhere ?

analogrelay: That experience is a little less ideal than I'd like, but it's functional. I was hoping we could check for that case and provide an error or alternate behavior. In my testing, what happens in Unity pops a dialog:

image

This seems fine, but unfortunately _both_ buttons do the same thing, which is to return to the Save dialog. That's not terrible because at least it doesn't actually overwrite any bot code, and it puts control back in the user's hands. It's kinda annoying that the "Replace" button doesn't actually dismiss the dialog though, we could handle that case in our code if Unity told us it was happening. I think it's because this is the _Save File_ dialog but the conflict is with a folder. We're kinda abusing this dialog a _tiny_ bit because we're prompting to save a file but actually creating a folder. The other dialogs don't work for this though, the only one related to folders allows you to pick a folder to save something in, but doesn't give you control over the name.

So, I think the current behavior is OK, if not ideal.

Editor/Scripts/BotManagement/RGBotSynchronizer.cs
```diff
@@ -390,33 +426,16 @@ private string ZipLocalBot(long botId, string sourcePath)
...
          * <returns>String path of the folder</returns>
          */
-        private string CreateBotFolder(string botName, long botId, bool replace)
+        private string CreateDefaultBotFolder(string botName, bool replace)
         {
-            CreateParentFolders();
-            var botFolderName = $"{botName}_{botId}".Replace('-', 'n');
+            RGEditorUtils.CreateAllAssetFolders(BOTS_PATH);
+
+            var botFolderName = $"{botName}".Replace('-', 'n');
```
vontell, line 438: Is this needed anymore since the bot names won't include the ID? This was only to handle negative ID numbers

vontell (approved):
Just a small comment, otherwise LGTM!

Editor/Scripts/BotManagement/RGBotSynchronizer.cs
vontell: 👍 whenever you fix it, feel free to merge!

analogrelay:
The only change here post-approval is a rebase on `main` to pick up #121 so I'm going to go ahead and merge when the CI is green.

PR #116, opened 2023-12-06
[REG-1463] Generate 'partial' class to attach RGState and RGAction to host behavior

analogrelay:
This updates our Unity code gen to add partial classes that create a `RequireComponent` dependency from the "host" behavior (the `MonoBehavior` on which the `[RGState]` or `[RGAction]` attribute is placed) to the generated `RGState_[HostBehaviorName]` and `RGAction_[HostBehaviorName]` components. This signals to the Unity editor, when reimporting assets, to add the `RGState_...`/`RGAction_...` components to any `GameObject` that also has the host behavior attached. This mostly allows us to "auto-attach" the behaviors.

From my testing, this is fairly consistent. Unity [does caution that it doesn't check dependencies in all scenarios](https://docs.unity3d.com/ScriptReference/RequireComponent.html), but it does it whenever `GameObject.AddComponent` is called.

In order to generate this auto-attach class, we need the original host behavior to be marked as a `partial` class. I added code to check for that and report an error if that's not the case. WIth the changes in #105, this also causes a dialog to pop up to draw your attention to the log:

image

NOTE: One additional change I made is to restore the "Generate Scripts" menu item. After #105 merged, auto-generation is mostly working but there are a few scenarios where I've noticed it doesn't work:

1. If you change anything in the RegressionGames packages, it doesn't re-generate. This _mostly_ affects us
2. If you delete or otherwise mess with the generated code (perhaps accidentally).

Both cases are fairly rare (case 1 is mostly just for us), but it seems reasonable to me to bring back the _ability_ to manually regenerate the scripts, even if it's only needed in edge cases.

---


Editor/Scripts/CodeGenerators/GenerateRGStateClasses.cs
```diff
@@ -97,12 +103,13 @@ public static void Generate(List<RGStateAttributesInfo> rgStateAttributesInfos)
                                     GenerateStateEntityFields(rgStateAttributeInfo.State)
                                 ),
                             classDeclaration
-
                         );
 
                     // Add the namespace declaration to the compilation unit
                     compilationUnit = compilationUnit.AddMembers(namespaceDeclaration);
 
+                    compilationUnit = compilationUnit.AddMembers(CodeGeneratorUtils.CreatePartialAttacherClass(rgStateAttributeInfo.NameSpace, componentType, $"RegressionGames.RGBotConfigs.{className}"));
```
RG-nAmKcAz, line 111: I think we should consider this having a `RequireComponent` for the behavior we found the RGState or RGAction attribute in as well.  Thus if they remove that behavior they will get warned that our scripts require it.

RG-nAmKcAz (changes requested):

rcornwall-reg (approved):

RG-nAmKcAz (approved):

analogrelay:
Resolved a gnarly enough conflict that I'd like to get re-reviews (particularly from @RG-nAmKcAz since that's where the conflict arose and I want to make sure I didn't blast away his changes). I'll do another smoke-test to make sure things are working.

And I see you already approved 🤣 . Thanks @RG-nAmKcAz !

RG-nAmKcAz:
> Resolved a gnarly enough conflict that I'd like to get re-reviews (particularly from @RG-nAmKcAz since that's where the conflict arose and I want to make sure I didn't blast away his changes). I'll do another smoke-test to make sure things are working.
> 
> And I see you already approved 🤣 . Thanks @RG-nAmKcAz !

For clarity, I'm not hand inspecting the code to see if you merged it correctly.  I would expect the dev themselves to do that as necessary.  I didn't see anything obvious though at a high level glance.

analogrelay:
> For clarity, I'm not hand inspecting the code to see if you merged it correctly. I would expect the dev themselves to do that as necessary. I didn't see anything obvious though at a high level glance.

Yep, that's my expectation too, but I think it's always valuable for the conflicting commit author to take a scan over it. Thanks for looking!

analogrelay:
There was one small style change lost in the conflict resolution. I've restored that: https://github.com/Regression-Games/RGUnityBots/pull/116/commits/16e8995f513e22ab571db7e72eb0654656565f93

PR #119, opened 2023-12-07
[REG-1524] Don't allow RGState on delegates anymore

analogrelay:
More details in the linear ticket, but I can't quite see why we support Delegates for the `RGState` attribute, so I'm opening this PR to remove it. Happy to adjust if I'm missing something here.

We allow you to apply `RGState` to a delegate type:

```csharp
public class MyBehavior: MonoBehavior
{
    [RGState]
    public delegate string Foo();
}
```

What I don't understand is what we expect that to do. Delegates are types, and have no value at runtime. They are named function signatures. The following C#:

```csharp
public delegate bool Predicate<T>(T value)
```

Is equivalent to this Kotlin:

```kotlin
typealias Predicate<T> = (T) -> Boolean
```

In fact, the code we generate ends up trying to call `myBehavior.Foo()`, where `myBehavior` is an instance of `MyBehavior`. That isn't valid C# since `Foo` is a _type_. So, there's nothing we can invoke at runtime to get this value.

Given that, I'm opening this to remove support for generating code for delegates, to prevent users from getting compilation errors from our generated code. We can always restore it if there _is_ a need to generate this code later.

---

### DIFF
--- a/src/gg.regression.unity.bots/Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
+++ b/src/gg.regression.unity.bots/Editor/Scripts/CodeGenerators/RGCodeGenerator.cs
@@ -34,7 +34,7 @@ namespace RegressionGames.Editor.CodeGenerators
      * 1. Find [RGAction] attributes and generate classes for them; captures the generated class name
      * 2. /\ This same information is used as the available Actions for AgentBuilder json.
      */
-    
+
     public class RGCodeGenerator
     {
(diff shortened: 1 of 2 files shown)


RG-nAmKcAz (approved):

rcornwall-reg (approved):

analogrelay:
The only change here is to rebase on `main` after #121 landed. Once this goes green I'm going to merge it since there are no relevant changes since approval.

PR #125, opened 2023-12-13
Revert "[REG-1463] Generate 'partial' class to attach RGState and RGAction to host behavior"

analogrelay:
Reverts Regression-Games/RGUnityBots#116

Draft, because I used GitHub to get this started and I need to pull it down local and make sure it does indeed fix the problem.


analogrelay:
This is mostly a straight revert, but I did bring one change from REG-1463 back: The menu item to forcibly regenerate scripts. I've found that really useful for the edge cases where automatic gen doesn't work (which are 99% limited to _our_ local dev scenarios, where we're messing with the generator itself). It's also useful for users to have the ability to force regeneration IMO.

RG-nAmKcAz (approved):

src/gg.regression.unity.bots/Editor/Scripts/CodeGenerators/CodeGeneratorUtils.cs
```diff
@@ -1,24 +1,15 @@
...
 namespace RegressionGames
 {
 #if UNITY_EDITOR
-    public static class CodeGeneratorUtils
+    public class CodeGeneratorUtils
     {
-        public static readonly string HeaderComment = "/// <auto-generated>" + Environment.NewLine +
-                                                      "/// This file has been automatically generated. Do not modify." + Environment.NewLine +
-                                                      "/// </auto-generated>" + Environment.NewLine + Environment.NewLine;
-
```
analogrelay, line 21: This was probably what messed up newlines too 😨

vontell (approved):

PR #126, opened 2023-12-13
[REG-1530] Automatic code formatting for C# code

analogrelay:
This PR adds a new PR check and a new script:

* `script/fmt` can be run from your local machine to reformat all the C# code. By default, it expects you _already have the sln/csproj files_. This is because regenerating them requires Unity and if you have the project open in Unity it will fail. You can specify the `--resync` argument to have the script re-generate these files before attempting to reformat.

* A new PR check to validate the code meets the style guidelines specified by `dotnet-format`. In general, this should match with defaults in Rider, it's a fairly conservative set of rules mostly focused on whitespace and indentation.

This pattern (a `fmt` script and a PR check) is what I've used most often, but I know we also have a pre-commit hook in other repos. If folks are interested in that approach, I'm open to it but I'm concerned about the requirement to resync sln/csproj files before reformatting and that you can't run Unity from the command line when a project is already open. It does mean that if you don't remember to run `script/fmt`, **and** don't have Rider or other editors set up with auto formatting, you might get a build failure from the fmt check and have to run `script/fmt` locally and push a new commit. That can be annoying too, so I'm very open to ideas.

This PR doesn't fix any of the style issues. That's because including those would overwhelm the other changes in this PR and make it difficult to review. Instead, I've opened a separate PR _on top of this branch_ (#127) to show those changes. When both PRs have the required approvals I will merge the PRs together and then merge them to `main` in a single merge.

---


RG-nAmKcAz:
I want this in place.

But, I am unlikely to remember to or want to run the manual format script any better than I remember to tell the IDE to format the code.

Whatever our solution is, it should be as automatic as possible. (build time, pre-commit time, somewhere like this)

analogrelay:
Yep, that's the main problem I've hit in the past. I can investigate a pre-commit hook, since that's something we've done elsewhere.

addisonbgross:
I think the pre-commit check would work well! Since we're all using Rider for the most part, it should be simple enough to enable a plugin for it (if needed). I'd hate to have to do another push just to fix some formatting. Having said that: I rarely work in this repo

analogrelay:
Ok, we have a pre-commit hook. When you run `script/bootstrap`, we install a symlink to `script/pre-commit` as a Git pre-commit hook. That hook will run `script/fmt` any time you commit.

I have _left_ the `fmt.yml` GitHub check in place, since pre-commit hooks aren't reliable, but I'm OK dropping it if folks are happy enough without it. I'd like to avoid extraneous format changes in PRs, so I'm trying to be as confident as possible that when a PR is merged, it contains well-formatted code and a later PR won't just randomly fix-up that formatting. But I **don't** want this to be annoying, so if it becomes annoying, we can yeet it.

So to wrap up, here's what we have:

* `script/fmt` can be run any time to reformat your code automatically. `script/fmt --check` can be run to check but not fix formatting
* `script/pre-commit` contains all precommit checks and fixes we want, right now just `fmt`
* `script/bootstrap` can and should be run regularly to "update" your environment. It's idempotent and safe to run any time
* On commit, we'll reformat code and commit changes to _files already added to the pending commit_. We _will_ reformat files that aren't staged, but _will not_ commit them. This helps with staged committing (committing some of your files but not all of them).
* PR builds will check that the code in the branch has clean formatting. In rare cases, this build may fail if the precommit hook failed or wasn't run. If that happens, a `script/fmt` in your branch should fix you up.

analogrelay:
Once again, this PR is _expected_ to fail the C# formatting check. When it is approved, I will merge #127, which contains the actual formatting changes, **into** this PR. That will make this PR noisier (since it will be filled with all the formatting fix-ups necessary to pass the build), so I wanted approval on the main workflow first.

RG-nAmKcAz:
Can we make the pre-commit hook itself re-run `script/bootstrap` ?

It would be nice if this kept itself up to date as much as possible.

analogrelay:
> Can we make the pre-commit hook itself re-run script/bootstrap ?

I suppose it could. My concern would be that `script/bootstrap` [is often in charge of things like updating dependencies](https://github.com/github/scripts-to-rule-them-all?tab=readme-ov-file#scriptbootstrap) and I'm not sure I like that happening on every commit. Currently it'd be safe to run on precommit since all it does is very that you have Unity installed and then link the precommit script. I don't think there's value in re-linking the precommit script from the precommit script since if the precommit script is running then it's by-definition "up to date".

abeizer (approved):
Not a wiz with this stuff so I'd wait for more approvals, but nothing jumps out at me as "omg change this"

vontell (approved):

addisonbgross (approved):

analogrelay:
Changes since approval:

1. Merged `main`
2. Ran `script/fmt`

I did that instead of merging #127 because there had been a lot of changes that conflicted with the style/whitespace changes (*cough* @RG-nAmKcAz *cough* 🤣 ).

Given that, I'm going to merge this when the build goes green because it still honors the spirit of the approvals IMO.

RG-nAmKcAz:
@analogrelay Maybe you covered this already, but is there a quick way (template/format-rules/etc) to setup Rider to format in this same way automatically ?

analogrelay:
@RG-nAmKcAz yeah, if you go to "Tools > Actions on Save" in the Settings dialog, you can configure things to happen on save. You have a lot of control to customize the reformatting "profile" to use, I'll do a little looking to see what makes sense. A "Full Cleanup" is probably going to work, but might do more than we usually do.

image

PR #128, opened 2023-12-13
[REG-1531] Fix compiler warnings and enable warnings as errors

analogrelay:
To keep the build tidy, I've enabled warnings as errors in this PR. I have a commit locally coming shortly which also fixes all the warnings, but I want to confirm the build fails because of the warnings-now-as-errors.

Most warnings are just unused fields and variables. There was one more complicated category to fix: Unnecessary `async` methods. Methods or lambdas marked `async` but with no `await` calls add unnecessary state machine logic to the generated code. The compiler can't just remove this logic because it has _some_ impact on the behavior. Most notably, it has to wrap exceptions that are thrown in the code into faulted `Task` objects. However, when the code is completely synchronous, this is a waste of time. So the compiler warns us about it. Almost all the places where we had `async` but no `await` were in the `onSuccess` and `onFailure` parameters to `SendWebRequest` and related helpers. So, I added overloads of those functions that take _non-async_ versions of `onSuccess` and `onFailure` (just wrapping them up to pass them to the main helper).

---


Directory.Build.props
```diff
@@ -0,0 +1,11 @@
+<!--
+    This file is automatically imported in to every 'csproj' in this directory.
+    This allows us to set settings that apply _when building in Rider_.
+    NOTE: These settings do NOT apply when building in Unity!
+    The src/RGUnityBots/Assets/csc.rsp file applies in Unity.
+-->
+<Project>
+    <PropertyGroup>
+        <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
RG-nAmKcAz, line 9: I don't agree with this as a concept.

Warnings are warnings
Errors are errors

If we have selected warnings that are causing issues in the compiled runtime or performance problems we should elevate those specific warnings to error level.

RG-nAmKcAz (changes requested):

Directory.Build.props
analogrelay: I can live with that if that's the general consensus. I know that in most cases, and in the C# compiler in particular, most warnings represent cases where the behavior is specified and deterministic but surprising to the user. So, most projects I've worked on elevate all warnings to errors and then suppress warnings that aren't relevant (rather than the other way around). This is similar to how a lot of more modern languages are forgoing warnings entirely in place of on-by-default lints that can be disabled or suppressed in certain cases.

But honestly I don't think it's worth much energy debating it so I'm happy to back it out.

RG-nAmKcAz: I wouldn't give up on it, I just wanted to spark a team discussion before merging. @vontell @addisonbgross @abeizer @rcornwall-reg 

Some of the things it calls out are quite good to fix (as noted by your changes)

RG-nAmKcAz: footnote.. I'm approving either way pending the results of the discussion since the other change was made

RG-nAmKcAz (approved):

Directory.Build.props
vontell: Yeah I am also unsure of this - mostly because this goes against the grain of how all of our other build processes work as well (in other builds, warnings are displayed and some correct during pre-commit hooks, but doesn't stop the build).

Would it make sense to instead of this as a pre-commit hook? This would exactly match the flow we have for `RegressionGames`, which doesn't let you commit if the linter finds errors, for instance.

If we did it that way vs this way, then instead of failing in a build, we can fail on warnings before a warning ever even gets committed. Thoughts?

More concretely - for this approach, I would need to commit and push and wait for a build to kick off, rather than having the instant feedback locally of a warning. I imagine cases where this can become a timely iteration loop / adds cycles to the PR prep, whereas a pre-commit would be more immediate.

analogrelay: My main concern is that if we don't believe these warnings are worth fixing, we should avoid filling the build output and PR check results with them. In most cases, these warnings point out surprising code patterns that should be avoided. If we don't fail the build for them, they will continue to accumulate and become harder to control for later. If we believe these warnings are not things we want to fix, then I'm fine with suppressing them. I have suppressed C# warnings in the past because they aren't as conducive to iterating mid-development.

As a concrete example, this is what Unity's log looks like because of all the warnings we have. I find it actually makes it _more_ difficult to find the compilation error among all the warnings:

image

GitHub also "helpfully" includes them in the PR files view:

image

We can (and indeed I have) fix all these now, but my concern is more warnings accumulating over time.

> Would it make sense to instead of this as a pre-commit hook? This would exactly match the flow we have for RegressionGames, which doesn't let you commit if the linter finds errors, for instance.

We can absolutely do that, but much like we do with RegressionGames, we have a precommit hook that builds _and_ we do that check on the server because precommit hooks are completely advisory and do not prevent a user from pushing, or even committing, code that violates them. I definitely believe we can trust our engineers not to bypass precommit hooks, but it still feels like a good idea to have the layers or protection here.

> More concretely - for this approach, I would need to commit and push and wait for a build to kick off, rather than having the instant feedback locally of a warning.

With the settings I have above, you see the warnings as errors in Unity and Rider. They appear exactly the same way compiler errors appear. This change actually _shortens_ that feedback loop.

Essentially, my primary concern is in removing a "grey area" of having a set of compiler diagnostics we _don't suppress_ but also _don't enforce_. Warnings are probably one of my least favorite compiler features, if I'm being honest, because they create this gray area of "You should fix this, but maybe don't?"

analogrelay: Sounds like there isn't consensus here, so I expect I'll just close this.

RG-nAmKcAz: while we debate how to enforce this .. i don't want to lose the work you did to fix the existing warnings... maybe we can merge that bit as its own PR ?

analogrelay: Yep, the most recent commit did exactly that. I'll ping to get some final approvals on this so we can merge it

vontell (approved):

RG-nAmKcAz (changes requested):
I small comment on most recent commit

analogrelay:
I think the comment itself is missing @RG-nAmKcAz ? At least I don't see it.

RG-nAmKcAz:
apparently you can't comment on lines that aren't changed.. even when github actions is showing warnings about them...

RGTickInfoActionManager line 116


do we need the pragma around this line also ?

githubactions is throwing warnings when viewing the changed files

analogrelay:
Yep, good catch. Should be fixed now.

RG-nAmKcAz (approved):

vontell (approved):

PR #129, opened 2023-12-15
In-loop bot experiment

analogrelay:
*Not for merging, just illustration*

Here's some pre-holiday experimentation for y'all to peruse at your leisure.

This is my experimentation branch that I used to write [REG-1538 In-Loop Bots](https://linear.app/regression-games/issue/REG-1538/in-loop-bots). The premise here is that we provide the option to strip away a lot of our machinery around managing local bots in their own background thread. Instead, a bot can provide a "Bot Controller" as part of it's `BotRecord.asset` file, and we'll just instantiate that in the scene and let it do the work.

Key files:

* `RGBotController` - This is a base class that you must derive from in your bot. Think of that derived class as a replacement for `BotEntryPoint.cs`.  It's just a regular `MonoBehavior` that's given a reference to the `BotInstance` when it's created (for state tracking).
* `RGBotAsset` gains an `RGBotController BotController` field used to track a reference to the Bot Controller component. This will usually be given a Prefab that has an `RGBotController` component somewhere on it. The entire prefab gets spawned when the bot is started, which allows the user to provide as many relevant GameObjects and behaviors as they'd like. _We_ just want to make sure one of those components is an `RGBotController` we can communicate with.
* `RGBotRuntimeManager` gets a lot simpler when a Bot Controller is in play. We just instantiate the controller, initialize it, create a "connection" for it, and then store all that away.
* `RGClientConnection_Local` also gets a lot simpler when a Bot Controller is in play. It basically just no-ops. It doesn't need to send tick info to the bot since the bot is already running in the loop.

Most of the rest of the changes are just incidental (things like suppressing cleanup of replay data so I can easily look at it locally). If/when we decide to proceed down this path, I will clean this up _and open a separate PR_. I don't intend to merge _this_ specific PR, it's just a convenient way to look at code.

Loom Demo:

https://www.loom.com/share/ada89f81bb09404681dd158a80921113


src/gg.regression.unity.bots/Runtime/Scripts/RGBotLocalRuntime/RGBotAsset.cs
```diff
@@ -20,5 +21,12 @@ public class RGBotAsset: ScriptableObject
         /// It may later use that to make conflict resolution decisions.
         /// </remarks>
         public string ChecksumAtLastSync;
+
+        /// <summary>
+        /// An object that will be spawned in the scene when this bot is spawned. This object can be used to retrieve state and trigger dynamic ticks.
+        /// </summary>
+        [FormerlySerializedAs("botDelegate")]
```
RG-nAmKcAz, line 28: looks like `FormerlySerializedAs` snuck into the PR

RG-nAmKcAz:
I did something in the same vein as this for Super Marxist Twins, but I stuck with using our state exposure logic.

I ended up using our 'state' logic to expose the state to my bot, but the process was very manual compared to getting the full state into my bot automagically.  The limited state was for sure more efficient and precise, but it had a big feeling of .. why bother with RG at all, I could have just done this as a custom written AI MonoBehaviour.

It's interesting/notable to me that we are both playing with the same concept, so there must be something here.  I just don't feel like either one of our code samples quite hits the nail on the head.


```csharp
        private void OnEnable()
        {
            _botEntryPoint = ScriptableObject.CreateInstance<BotEntryPoint>();

            //somewhere in the scene
            _rgStateLevel = FindObjectOfType<RGState_Platformer2DLevel>();

            // on same gameObject
            _rgStatePlayer = GetComponent<RGState_Platformer2DPlayer>();
            _rgActionSetBotJump = GetComponent<RGAction_SetBotJump>();
            _rgActionSetBotMove = GetComponent<RGAction_SetBotMove>();

        }
        
        private void Update()
        {
            // This is being controlled externally
            // This would be used to simulate first player being played
            if (runUpdateLoop)
            {
                // TODO: Move this lookup to some function that can happen after loading the level but not every update
                var victoryZone = FindObjectsOfType<RGEntity>().FirstOrDefault(v => v.objectType == "Victory");
                
                if (victoryZone != null)
                {
                    ProcessMovementToTarget(victoryZone.transform);
                }
            }
        }

        public void ProcessMovementToTarget(Transform target)
        {
            var levelState = _rgStateLevel.GetGameObjectState() as RGStateEntity_Platformer2DLevel;
            var playerState = _rgStatePlayer.GetGameObjectState() as RGStateEntity_Platformer2DPlayer;
            
            // have to do this every frame because of destructible blocks
            // if we had a nice method to call after the level loaded, we could optimize here by finding them all once including disabled
            // then just only send through the non-disabled ones to the bot code on each update
            var platformsOrBlocks = FindObjectsOfType<RGState_Platformer2DPlatformOrBlock>().Select(v => 
                v.GetGameObjectState() as RGStateEntity_Platformer2DPlatformOrBlock
            );

            //This calls the actual bot entry point method that is the 'same' thing that would be called by ProcessTick if running in our separate thread.
            _botEntryPoint.ProcessState(levelState, platformsOrBlocks, playerState, target == null ? null : target.position, _rgActionSetBotJump, _rgActionSetBotMove, renderDebugGizmos);
        }
```

analogrelay:
> why bother with RG at all, I could have just done this as a custom written AI MonoBehaviour.

I came to the same conclusion too, but I think we shouldn't try to run from that idea because it's telling us something about where our focus should be. If we find that the Bot "Framework" in the SDK isn't adding much value, that means we should refocus our goals on what actually _does_ provide value. I think we might find that the SDK's purpose is _not_ to make a complicated framework for writing your bot code, and that's **ok**. Instead, we shift our focuses to a few smaller areas that may not need much "framework" at all but still give you value _through the RG service_:

1. Easy collection of state during a session (via `[RGState]` and perhaps imperative mechanisms within the game loop itself to capture state), both for bot decision making and for sending off to RG for storage and later analysis.
2. Easy injection of hooks for remote automation (RGAction), to allow for LLM and other cloud-powered automation.
3. A simple, but optional, bot "lifecycle" in which we can create/destroy your bot and manage it's lifecycle. But it isn't _necessary_ to use this to still take advantage of 1 & 2 if you want.

I think it's _good_ to realize we can strip back a lot of our SDK structure and simplify down to a model where the bot is basically just a Game Object you control that has access to APIs we provide. If you want remote control, we have a way to handle dispatching state to the remote bot and processing incoming actions. If you want state recording, we have tools to allow that. If you _do_ want a nice bot "API" that ties that together (i.e. the `BotController` stuff I mention above) then we've got it, but you can also just make calls to RG APIs yourself in your own components within the game.

analogrelay:
I'm going to fiddle with this some more and make another iteration that looks closer to what I'm thinking ☝🏻 , and I'll notify folks when that iteration is pushed.

PR #130, opened 2023-12-18
Release 0.0.16

vontell:


RG-nAmKcAz (approved):

README.md
```diff
@@ -29,7 +29,7 @@ In the Unity Editor, inside of **Window** > **Package Manager** > **+** >
 **Add package from git URL**, enter this URL (and restart your IDE after adding):
 
 ```
-https://github.com/Regression-Games/RGUnityBots.git#v0.0.15
+https://github.com/Regression-Games/RGUnityBots.git?path=src/gg.regression.unity.bots#v0.0.16
```
analogrelay, line 32: I just noticed our _actual_ tag for v0.0.15 appears to have been `0.0.15` (no `v` prefix). Which means the previous link wouldn't have worked either.

It's fine either way (`v` or no `v`) but the value after `#` here _must_ match the name of a Tag or Branch in the repo for the URL to work:

image

analogrelay (approved):
Approving. I'm confident @vontell can do the needful on my one and only semi-blocking comment :)

README.md
vontell: Ah slight correction from me - it looks like we _do_ have proper tags for these. I make a `v####` tag for the release which is how its resolving. https://github.com/Regression-Games/RGUnityBots/releases

vontell: slightly weird because I failed to make a `release-###` branch in the past, I just did the number, which is resolved now

addisonbgross (approved):

README.md
analogrelay: Ah right I forgot those have tags too. I don't really get why they don't appear in the tags list on the repo but 🤷‍♀️

PR #140, opened 2024-03-01
[REG-1636] Replays state recordings using the new unity input system only

RG-nAmKcAz:
- Replays state recordings using the new unity input system only (any code using Input.Get.. will NOT trigger)
- Fixes an annoyance where we log 2 errors when your RG credentials aren't setup
- Changes the state recording code and file format since the previous PR.. No, it will not be backwards compatible at this point, just make a new recording for your testing.
- At this time, this is fairly stable all things considered, but does NOT support hover over interactions very well.
- Includes 3rd party code for runtime SimpleFileBrowser (https://github.com/yasirkula/UnitySimpleFileBrowser)
- Adds 2 new behaviour scripts, 1 for excluding an object and its children from the state (used for RGOverlay), 1 for including an object in the state... useful for empty parent gameobjects that you may want state to roll up under

Demo Video: 

https://github.com/Regression-Games/RGUnityBots/assets/112959031/dae37f13-81ba-4f25-85e8-edeb3ab6029a

---


abeizer (approved):
The demo video is neat, I'm excited to consume this work for validations. I see a few TODOs that may need linear IDs, but I don't have any blocking comments. Just general coolness.

Side-note: Never have I imagined a file picker being this large. Wow.

RG-nAmKcAz:
@abeizer Re: File Picker... free runtime file picker asset so I didn't have to do the work... we can enhance the appearance later

abeizer:
@RG-nAmKcAz yes, I saw. That wasn't a criticism of your PR - just funny ;-)

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ReplayDataPlaybackController.cs
```diff
@@ -0,0 +1,499 @@
...
+            }
+        }
+
+        private InputDevice GetMouse(bool forceListener = false)
+        {
+            var mouse = InputSystem.devices.FirstOrDefault(a => a.name == "RGVirtualMouse");
+            var created = false;
+            if (mouse == null)
+            {
+                mouse = InputSystem.AddDevice<Mouse>("RGVirtualMouse");
```
vontell, line 293: Very relieved we can do stuff like this with the new input system 😬

vontell (approved):
Looking good! Excited to try this out. Note that we had a question about states of monobehaviours as well, sent in the validations slack channel.

PR #142, opened 2024-03-06
[REG-1639] New bot system

rcornwall-reg:
**Walkthrough**
https://www.loom.com/share/fa7751c7684e476196c65876cc523685?sid=50f26aa8-1c3d-4b52-88b4-391e89b4dfb4

**Bots are Monobehaviors**
Bot behavior scripts are just MonoBehaviors that inherit from `IRGBot`. No method implementations required

**Bot's can be spawned:**
1. Runtime:
`RGBots.SpawnBot<TestBot>();`
`RGBots.SpawnBot<TestBot>(prefab);`
`RGBots.SpawnBot("TestBot", prefab);`
`RGBots.SpawnBot("TestBot");`
2. Overlay: Choose the prefab and bot behavior from the dropdown, then click "Start"
3. Manually: Any gameObject in the scene with a bot script will be found by overlay

**New Overlay**
New `RGOverlayCanvasV2` 

**Runtime Support**
Because of their MonoBehavior nature, all bots work at runtime. Bot management and creation also fully supported at Runtime due to new `RGBotList` scriptable object

---


src/gg.regression.unity.bots/Runtime/Scripts/BotManagement/BotSpawnManager.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using RegressionGames;
+using UnityEngine;
+using RegressionGames;
```
vontell, line 6: I think it wants you to fix this 😬

src/gg.regression.unity.bots/Editor/Scripts/BotManagement/IRGBotFinder.cs
```diff
@@ -0,0 +1,71 @@
...
+public class IRGBotFinder : Editor
+{
+    [DidReloadScripts]
+    private static void OnScriptsReloaded()
+    {
+        // Only run this method in the Unity editor, not at runtime
+        if (!EditorApplication.isPlayingOrWillChangePlaymode || EditorApplication.isPlaying)
+        {
+            return;
+        }
```
vontell, line 16: Is there actually a downside of allowing this at play time? Wouldn't this have the benefit of letting us see new bots on the fly (assuming your game even does well with script reloading in the first place)

src/gg.regression.unity.bots/Editor/Scripts/BotManagement/IRGBotFinder.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        // Find the IRGBotList ScriptableObject in your project
+        IRGBotList botList = LoadBotList();
+
+        if (botList == null)
+        {
+            Debug.LogWarning("IRGBotList ScriptableObject not found");
+            return;
+        }
```
vontell, line 25: Instead, should this be created for them? This was actually one question I had from the video - how does this get into their project?

src/gg.regression.unity.bots/Editor/Scripts/BotManagement/IRGBotFinder.cs
```diff
@@ -0,0 +1,71 @@
+using UnityEditor;
```
vontell, line 1: I notice this file is in the Editor folder, but wouldn't we also want this for builds that QA developers are using on a device?

src/gg.regression.unity.bots/Editor/Scripts/BotManagement/IRGBotFinder.cs
```diff
@@ -0,0 +1,71 @@
...
+        // Mark the ScriptableObject as dirty to ensure it saves
+        EditorUtility.SetDirty(botList);
+        AssetDatabase.SaveAssets();
+        AssetDatabase.Refresh();
+    }
+
+    private static IRGBotList LoadBotList()
+    {
+        string packagePath = "Packages/gg.regression.unity.bots/Resources/RGBotList.asset";
+        string localPath = "Assets/Runtime/Resources/RGBotList.asset";
```
vontell, line 61: So is this specific project using the asset inside of the package? Curious about the tradeoffs of creating this asset specifically in their project rather than inside the package.

src/gg.regression.unity.bots/Runtime/Scripts/BotManagement/BotSpawnManager.cs
```diff
@@ -0,0 +1,41 @@
...
+using RegressionGames.Types;
+
+public class BotSpawnManager : RGBotSpawnManager
```
vontell, line 9: What exactly is this file for? This seems related to the old approach.

src/gg.regression.unity.bots/Runtime/Scripts/IRGBot.cs
```diff
@@ -0,0 +1,5 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public interface IRGBot { }
```
vontell, line 5: I wonder if we should make this an abstract class... maybe we can always change it later, but I imagine there may be some built-in, implemented functions we may want to add at some point.

src/gg.regression.unity.bots/Runtime/Scripts/RGBots.cs
```diff
@@ -0,0 +1,95 @@
...
+        var behaviorName = botEntry?.botName ?? "Empty";
+        if (botEntry != null)
+        {
+            botGameObject.AddComponent<T>();
+        }
+        LinkToOverlay(botGameObject, behaviorName);
+        return botGameObject;
+    }
+
+    public static GameObject SpawnBot<T>(GameObject prefab) where T : Component
```
vontell, line 47: Just realized, for these, should we also be providing a position vector? 

Also, before we get to merging, these functions specifically should have comments since they will be user-facing.

vontell (changes requested):
Just a few initial comments, coming along well!

src/gg.regression.unity.bots/Runtime/Scripts/BotManagement/IRGBotEntry.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Collections.Generic;
+using UnityEditor;
+using UnityEngine;
+
+
+[Serializable]
+public class IRGBotEntry
+{
+    public string botName;
+    public MonoScript monoScript;
```
RG-nAmKcAz, line 12: This file prevents the project from building outside the editor.  We should not reference/use anything related to `UnityEditor` in the runtime files.

src/gg.regression.unity.bots/Runtime/Scripts/IRGBot.cs
RG-nAmKcAz: I would vote against that.  You will already be inheriting from MonoBehaviour, keeping this as an interface would seem more appropriate

src/gg.regression.unity.bots/Runtime/Scripts/RGBotManager.cs
```diff
@@ -0,0 +1,234 @@
+using System;
```
RG-nAmKcAz, line 1: Let's move these UI overlay files into their own folder.

src/gg.regression.unity.bots/Runtime/Scripts/RGBotEntry.cs
```diff
@@ -0,0 +1,42 @@
+using System.Collections;
```
RG-nAmKcAz, line 1: Let's move these UI overlay files into their own folder.

src/gg.regression.unity.bots/Editor/Scripts/BotManagement/IRGBotFinder.cs
RG-nAmKcAz: +1

RG-nAmKcAz (changes requested):
Conceptually I like this, seems clean and simple.  Some comments to address before merge.

- cleanup unused using declarations in all files
- multiple files are missing namespace declarations

There is at least one case where we are referencing editor libraries in runtime scripts, will need to remove that.


Also, is there a doc update PR to go with this, or updated sample project examples ?  If not, those need tasks created

RG-nAmKcAz:
@rcornwall-reg , when you merge with main, you need to sync up the overlayv2 with the old overlay.  There were new behaviours added to it in my state recording prs

RG-nAmKcAz:
@vontell I think this should eliminate the options for 'synchronize bots with RG'.  That doesn't really have meaning anymore does it ?

vontell:
> @vontell I think this should eliminate the options for 'synchronize bots with RG'. That doesn't really have meaning anymore does it ?

I believe that would still be used for Agent Builder? But really we shouldn't have a "sync" option, it would be a pull option?

RG-nAmKcAz:
> > @vontell I think this should eliminate the options for 'synchronize bots with RG'. That doesn't really have meaning anymore does it ?
> 
> I believe that would still be used for Agent Builder? But really we shouldn't have a "sync" option, it would be a pull option? because we don't care much now about syncing back and forth, just pulling down any bots that they haven't defined locally. But we could remove it until we fix agent builder, im fine with that too

There needs to be a task to track that work created before this is merged then.

RG-nAmKcAz:
@rcornwall-reg 

I'm using this branch as the basis of my state removal work.  As part of that, I wrote new bossroom bots for this system.

Issue I'm hitting... I can't load a bot.  RGBots.SpawnBot is failing trying to GetType for the botEntry.qualifiedName

the qualified name for my bot got persisted by the code as 
`TestBot, Assembly-CSharp, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null`

which is not parseable input for GetType

RG-nAmKcAz:
Random observation... 'Empty' stays in the bot list even when I have a bot, we should hide 'empty' when we have bots.

src/gg.regression.unity.bots/Runtime/Scripts/RGBots.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        GameObject botGameObject = GameObject.Instantiate(prefab);
```
RG-nAmKcAz, line 49: We should name this so it isn't just 'New Game Object' in the scene

src/gg.regression.unity.bots/Runtime/Scripts/IRGBot.cs
rcornwall-reg: @vontell I could see either way, I don't think it will really affect the dev experience so more of an architecture question

`public class MyTestBot: MonoBehavior, IRGBot`
`public class MyTestBot: MonoBehavior, RGBot`

src/gg.regression.unity.bots/Runtime/Scripts/RGBots.cs
rcornwall-reg: 👍

src/gg.regression.unity.bots/Editor/Scripts/RGOverlay/IRGBotFinder.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            // Mark the ScriptableObject as dirty to ensure it saves
+            EditorUtility.SetDirty(botList);
+            AssetDatabase.SaveAssets();
+            AssetDatabase.Refresh();
+        }
+
+        private static IRGBotList LoadBotList()
+        {
+            string packagePath = "Packages/gg.regression.unity.bots/Runtime/Resources/RGBotList.asset";
```
RG-nAmKcAz, line 58: I thought the discussion yesterday concluded with us copying this default into their project and editing the one in their project ?

src/gg.regression.unity.bots/Runtime/Scripts/RGBots.cs
RG-nAmKcAz: This comment doesn't seem to be fixed in the current changes.

RG-nAmKcAz (changes requested):

RG-nAmKcAz:
> > > @vontell I think this should eliminate the options for 'synchronize bots with RG'. That doesn't really have meaning anymore does it ?
> > 
> > 
> > I believe that would still be used for Agent Builder? But really we shouldn't have a "sync" option, it would be a pull option? because we don't care much now about syncing back and forth, just pulling down any bots that they haven't defined locally. But we could remove it until we fix agent builder, im fine with that too
> 
> There needs to be a task to track that work created before this is merged then.

Reminder that this task(s) needs to be created before merging

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGBotManager.cs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    void BehaviorSelected(TMP_Dropdown dropdown)
+    {
+        int index = dropdown.value;
+
+        // Check if "Empty" is selected
+        if (index == 0)
+        {
+            _selectedBehavior = null;
```
RG-nAmKcAz, line 217: Trying this out, you can't start the bot at the first slot in the list.

RG-nAmKcAz (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGBotManager.cs
```diff
@@ -0,0 +1,236 @@
...
+        RGDebug.LogVerbose("Showing RG Overlay Menu");
+        selectionPanel.SetActive(true);
+    }
+
+    public void OnOverlayClosed()
+    {
+        selectionPanel.SetActive(false);
+    }
+
+    public void AddBot()
```
RG-nAmKcAz, line 112: This code makes the assumption that `_selectedBehavior` has been assigned by `BehaviourSelected`.  But that doesn't get called unless the user changes the dropdown list.  So if they leave it on the first entry, the `_selectedBehavior` is null and the bot fails to start.

RG-nAmKcAz (changes requested):

src/gg.regression.unity.bots/Editor/Scripts/BotManagement/IRGBotFinder.cs
vontell: @rcornwall-reg note that in our PR process, never resolve comments yourself - the reviewers will mark the comments as resolved. Instead, just put a comment saying that you fixed it.

src/gg.regression.unity.bots/Editor/Scripts/RGOverlay/IRGBotFinder.cs
rcornwall-reg: After the package path was fixed, it's working as expected with the scriptable object in the package. It should never have to be seen or updated by a developer so if there's no technical reason to copy it over, I think we should leave it. What do you think? @RG-nAmKcAz @vontell

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGBotManager.cs
rcornwall-reg: Fixed

src/gg.regression.unity.bots/Runtime/Scripts/RGBots.cs
rcornwall-reg: Fixed. Current logic is:
1. Name of the bot will be the name of the prefab spawn
2. If no prefab, name of the bot will be name of the behavior selected
3. If no prefab and no behavior, uses `New Game Object`

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGBotManager.cs
rcornwall-reg: Fixed

src/gg.regression.unity.bots/Editor/Scripts/RGOverlay/IRGBotFinder.cs
RG-nAmKcAz: Quite simply, you can't modify that file in our package unless you're an RG developer loading it from your local filesystem.

When using the package from a tarball or from git that file isn't writeable.

image

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGBotManager.cs
```diff
@@ -0,0 +1,237 @@
...
+
+    void BehaviorSelected(TMP_Dropdown dropdown)
+    {
+        int index = dropdown.value;
+
+        // Check if "Empty" is selected
+        _selectedBehavior = null;
+        if (_botEntries.Length > 0)
+        {
+            _selectedBehavior = _botEntries[index - 1].botName;
```
RG-nAmKcAz, line 222: Robb, this still doesn't work.  It is selecting the wrong index.

src/gg.regression.unity.bots/Editor/Scripts/RGOverlay/IRGBotFinder.cs
rcornwall-reg: Implemented 👍

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGBotManager.cs
rcornwall-reg: Should be fixed in the latest

RG-nAmKcAz (approved):
@rcornwall-reg My existing findings seem to be fixed.

I ran into an issue with my specific bot code that i'll need to work through before confirming if there are any other bugs

vontell (approved):

PR #144, opened 2024-03-11
[REG-1659] Working network variable json conversion, a bit slow though

RG-nAmKcAz:
- Includes bossroom code in state (even thought it is 'Unity')
- Passes through values of Unity.Netcode things like NetworkObjects and NetworkVariables
- Improves overall state recording json performance... unfortunately that is more than offset by the volume of network state sent by bossroom now. Getting the state doesn't take too long, but writing to Json for all the behaviour fields with nested objects certainly does

---


abeizer (approved):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/JsonConverters/ImageJsonConverter.cs
```diff
@@ -16,18 +16,13 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
-                serializer.Serialize(writer, val.color, typeof(Color));
-                writer.WritePropertyName("material");
-                writer.WriteValue(val.material?.name);
-                writer.WritePropertyName("raycastTarget");
-                writer.WriteValue(val.raycastTarget);
-                writer.WritePropertyName("preserveAspect");
-                writer.WriteValue(val.preserveAspect);
-                writer.WriteEndObject();
+                // raw is way faster than using the libraries
+                writer.WriteRawValue("{\"sourceImage\":" + JsonConvert.ToString(val.sprite?.name)
```
vontell, line 20: Just out of curiosity, does C# also have a StringBuilder concept like in Java? I wonder if these string constructions are creating lots of intermediate strings.

vontell (approved):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/JsonConverters/ImageJsonConverter.cs
RG-nAmKcAz: they do, but + is faster for anything less than about 1000 appends in a code block given how the compiler optimizes + to be a concat operation

RG-nAmKcAz: if you want a detailed rundown on the performance details i'd be happy to share, but suffice to say, nothing so easy/shallow as this is going to solve the issue / I've 99.99% sure already considered and/or tested it

RG-nAmKcAz:
Notes of other things considered outside of code level micro-optimizations
1. Pulled in System.Text.Json dlls and tried that as it is known to perform better than Newtonsoft JSON...  It did perform betterish, BUT has no way to handle exceptions on fields during serialization (like deprecation) and thus is useless for serializing arbitrary monobeheaviours as unity has inherently deprecated fields all over the place that throw exceptions in serialization.

RG-nAmKcAz:
have brought frame time on a recorded frame down from 60+ms to < 20ms on the boss room main menu when running deep profiling by
0. altering newtonsoft json config to avoid repeated type lookups
1. micro optimizing the json conversions to skip most of Newtonsoft when possible
2. writing custom serializer for decimal/float/double types to cut time to write float values in half (float to string was 6+ms of the original 60ms on its own and did not scale down with other changes made)

PR #147, opened 2024-03-14
[REG-1667] Fix overlay resizing

rcornwall-reg:
Updated `OverlayCanvasV2` prefab and `RGBotEntry` prefab to correctly resize on all screen sizes

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Prefabs/RGOverlayCanvasV2.prefab
+++ b/src/gg.regression.unity.bots/Runtime/Prefabs/RGOverlayCanvasV2.prefab
@@ -792,7 +792,7 @@ GameObject:
   m_Icon: {fileID: 0}
   m_NavMeshLayer: 0
   m_StaticEditorFlags: 0
-  m_IsActive: 0
+  m_IsActive: 1
 --- !u!224 &1094668923711103164
 RectTransform:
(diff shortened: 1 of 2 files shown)


src/gg.regression.unity.bots/Runtime/Prefabs/RGOverlayCanvasV2.prefab
```diff
@@ -792,7 +792,7 @@ GameObject:
   m_Icon: {fileID: 0}
   m_NavMeshLayer: 0
   m_StaticEditorFlags: 0
-  m_IsActive: 0
+  m_IsActive: 1
```
RG-nAmKcAz, line 795: This seems to now be active by default... I don't think that was intentional

RG-nAmKcAz (changes requested):
Layout seems to work well, but overlay should be off by default

src/gg.regression.unity.bots/Runtime/Prefabs/RGOverlayCanvasV2.prefab
rcornwall-reg: Correct - fixed in latest

RG-nAmKcAz (approved):

vontell (approved):

PR #148, opened 2024-03-16
Gameplay Session Uploading from the SDK

vontell:
This PR adds the capability to upload screen recordings as gameplay sessions in the backend.

See loom video (no audio, see video description): https://www.loom.com/share/470bb5bda5ab4e79ac9d7037857ce1c0?sid=9bcc1898-0be0-4d9a-be37-12f9a811f464

* Adds endpoints in RGServiceManager for creating and uploading resources
* Modifies the screen recorder logic to separate out screenshots and data for ease of uploading later
* Adds a text indicator to the overlay so that you know when something is being uploaded, since this can be a long-running task
* Confirmed to work in a build as well 🎉 

Request for reviewers:
* I am no expert yet on async code in Unity... I don't even get how this is working since I thought I was spawning off an async task on a background thread which then runs API code, which shouldn't be allowed... if there are any modifications to my logic for async stuff, let me know! Looking to learn

Known limitations:
* We don't have much retry logic for uploads
* This has not been tested with Azure upload yet, only local (I am attempting this in dev now which should have all the recent needed code...)

---


RG-nAmKcAz:
@vontell Github shows a lot of errors on the changes page, is this actually ready for review ?

vontell:
@RG-nAmKcAz it is not ready yet, it's marked as a draft. I think it tagged you two because of the codeowners file... looking now at how to turn that off for draft PRs.

RG-nAmKcAz:
@vontell The setting I had turned on to make sure code-owners couldn't bypass main branch protections is what forced that.  I'll look a bit this AM, but honestly I think the code owners file should just be removed

RG-nAmKcAz:
@vontell The 'other' reason I could push to main was being an admin for this specific repository.  I have turned on the setting to prevent admins from doing that.  If we ever have a case where we truly need to , we can temporarily disable that setting

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -40,18 +40,21 @@ public class ScreenRecorder : MonoBehaviour
...
         private int _frameCountSinceLastTick;
 
-        private string _currentVideoDirectory;
+        private string _currentGameplaySessionDirectoryPrefix;
 
         private CancellationTokenSource _tokenSource;
 
         private static ScreenRecorder _this;
+        
+        private static RGServiceManager _rgServiceManager = new();
```
vontell, line 49: @RG-nAmKcAz I tried to grab an existing RGServiceManager in Awake, but it was not created by the time this object was awake. Would you prefer this approach, or one that allows this behaviour to create the singleton?

vontell:
@abeizer @addisonbgross if you pull this branch and the most recent RGBossRoom code, and locally start your servers and add an API key to your bossroom project, you can start to see gameplay sessions in the backend

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
RG-nAmKcAz: NEVER.. (mostly).. create a monobehaviour using 'new'.. there are very rare exceptions but this case isn't one of them.

We have discussed the topic of unity event functions in the past, but accessing other objects in Awake is unstable as there is no guarantee the other objects are Awake yet.  If you have something you want to happen any time the object wakes back up, you can use OnEnable, or if you want it to happen once and only once, you can use Start.

https://docs.unity3d.com/Manual/ExecutionOrder.html

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -150,18 +228,19 @@ public void StartRecording()
...
+                    _currentGameplaySessionDirectoryPrefix =
                         $"{stateRecordingsDirectory}/{Application.productName}/run_{_videoNumber++}";
-                } while (Directory.Exists(_currentVideoDirectory) || File.Exists(_currentVideoDirectory + ".zip"));
+                } while (Directory.Exists(_currentGameplaySessionDirectoryPrefix + "_data") || File.Exists(_currentGameplaySessionDirectoryPrefix + "_data.zip"));
 
-                if (!Directory.Exists(_currentVideoDirectory))
+                if (!Directory.Exists(_currentGameplaySessionDirectoryPrefix))
                 {
-                    Directory.CreateDirectory(_currentVideoDirectory);
+                    Directory.CreateDirectory(_currentGameplaySessionDirectoryPrefix + "_data");
```
RG-nAmKcAz, line 237: This isn't properly checking that the inner folders exist

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -150,18 +228,19 @@ public void StartRecording()
                 // find the first index number we haven't used yet
                 do
                 {
-                    _currentVideoDirectory =
+                    _currentGameplaySessionDirectoryPrefix =
                         $"{stateRecordingsDirectory}/{Application.productName}/run_{_videoNumber++}";
-                } while (Directory.Exists(_currentVideoDirectory) || File.Exists(_currentVideoDirectory + ".zip"));
+                } while (Directory.Exists(_currentGameplaySessionDirectoryPrefix + "_data") || File.Exists(_currentGameplaySessionDirectoryPrefix + "_data.zip"));
```
RG-nAmKcAz, line 233: since these are separated, we need to check both _data and _screenshots in our loop to avoid conflicts

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -100,18 +103,92 @@ private void OnDestroy()
...
+         * We separate out the async saving code since OnDestroy can't be async, and we need to wait for saving
+         * before doing API calls
+         */
+        private async void HandleRecordingSaving()
+        {
+            _isRecording = false;
+            var endTime = DateTime.Now;
+            KeyboardInputActionObserver.GetInstance()?.StopRecording();
+            MouseInputActionObserver.GetInstance()?.StopRecording();
+            var gameplaySessionDataDirectory = _currentGameplaySessionDirectoryPrefix + "_data";
```
RG-nAmKcAz, line 121: we are using _data and _screenshots strings in multiple places
we should either make these constants, or better yet, save off _ currentGameplaySession(Session|Data)DirectoryPrefix variable

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -366,7 +445,7 @@ private void ProcessFrames()
                     ImageConversion.EncodeNativeArrayToJPG(frameData, graphicsFormat, (uint)width, (uint)height);
 
                 // write out the image to file
-                var path = $"{directoryPath}/{frameNumber}".PadLeft(9, '0') + ".jpg";
+                var path = $"{directoryPath}_screenshots/{frameNumber}".PadLeft(9, '0') + ".jpg";
```
RG-nAmKcAz, line 448: I would make these subfolders , so that we don't have 2x the director listings in the top level folder.. 1 directory listing per run, inside.. screenshots and data

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -100,18 +103,92 @@ private void OnDestroy()
...
+        private async void HandleRecordingSaving()
+        {
+            _isRecording = false;
+            var endTime = DateTime.Now;
+            KeyboardInputActionObserver.GetInstance()?.StopRecording();
+            MouseInputActionObserver.GetInstance()?.StopRecording();
+            var gameplaySessionDataDirectory = _currentGameplaySessionDirectoryPrefix + "_data";
+            var gameplaySessionScreenshotDirectory = _currentGameplaySessionDirectoryPrefix + "_screenshots";
+            var thumbnailPath = _currentGameplaySessionDirectoryPrefix + "_thumbnail.png";
+            await Task.Run(() =>
```
RG-nAmKcAz, line 124: Running this on another thread but awaiting it seems pointless , why not just run it where we are

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -100,18 +103,92 @@ private void OnDestroy()
...
+                RGDebug.LogInfo($"Finished zipping replay to file: {gameplaySessionDataDirectory}.zip");
+                
+                // Then save the screenshots separately
+                RGDebug.LogInfo($"Zipping screenshot recording replay to file: {gameplaySessionScreenshotDirectory}.zip");
+                ZipFile.CreateFromDirectory(gameplaySessionScreenshotDirectory, gameplaySessionScreenshotDirectory + ".zip");
+                RGDebug.LogInfo($"Finished zipping replay to file: {gameplaySessionScreenshotDirectory}.zip");
+                
+                // Finally, we also save a thumbnail, by choosing the middle file in the screenshots
+                var screenshotFiles = Directory.GetFiles(gameplaySessionScreenshotDirectory);
+                var middleFile = screenshotFiles[screenshotFiles.Length / 2]; // this gets floored automatically
```
RG-nAmKcAz, line 139: This seems very arbitrary and likely to get some random(possibly even blank) frame.  I would suggest taking the first frame instead.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -100,18 +103,92 @@ private void OnDestroy()
...
+                File.Copy(middleFile, thumbnailPath);
+                Directory.Delete(gameplaySessionScreenshotDirectory, true);
+            });
+            CreateAndUploadGameplaySession(endTime);
```
RG-nAmKcAz, line 143: call this as

`_ = CreateAndUploadGameplaySession(endTime);`

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -100,18 +103,92 @@ private void OnDestroy()
...
+                RGBotManager.GetInstance().ShowUploadingIndicator(true);
+                HandleRecordingSaving();
+            }
+        }
+
+        /**
+         * We separate out the async saving code since OnDestroy can't be async, and we need to wait for saving
+         * before doing API calls
+         */
+        private async void HandleRecordingSaving()
```
RG-nAmKcAz, line 115: This shouldn't be async.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -100,18 +103,92 @@ private void OnDestroy()
...
+                // Finally, we also save a thumbnail, by choosing the middle file in the screenshots
+                var screenshotFiles = Directory.GetFiles(gameplaySessionScreenshotDirectory);
+                var middleFile = screenshotFiles[screenshotFiles.Length / 2]; // this gets floored automatically
+                File.Copy(middleFile, thumbnailPath);
+                Directory.Delete(gameplaySessionScreenshotDirectory, true);
+            });
+            CreateAndUploadGameplaySession(endTime);
+        }
+
+        private async void CreateAndUploadGameplaySession(DateTime endTime)
```
RG-nAmKcAz, line 146: `async void` is a code smell, this should return a Task

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -100,18 +103,92 @@ private void OnDestroy()
...
+                await _rgServiceManager.UploadGameplaySessionThumbnail(gameplaySessionId,
+                    _currentGameplaySessionDirectoryPrefix + "_thumbnail.png",
+                    () => { RGDebug.LogInfo($"Uploaded gameplay session thumbnail"); },
+                    () => { RGDebug.LogError($"Failed to upload gameplay session thumbnail"); });
+            }
+            catch (Exception e)
+            {
+                Debug.LogError(e);
             }
+            RGBotManager.GetInstance().ShowUploadingIndicator(false);
```
RG-nAmKcAz, line 190: This will ultimately call .`SetActive` off of the main thread in Unity... dragons ahead...

you can either make the function toggle a state variable that is checked in Update() of RGBotManager to enable/disable the indicator on the next update, or you can deal with launching this in a coroutine with `yield return null;` to make it do it on the main thread in the next frame, or some other cleverness to do this action on the next main thread update call

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -100,18 +103,92 @@ private void OnDestroy()
...
-                });
-                _isRecording = false;
+                    return;
+                }
+                
+                // Upload the gameplay session data
+                await _rgServiceManager.UploadGameplaySessionData(gameplaySessionId,
+                    _currentGameplaySessionDirectoryPrefix + "_data.zip",
+                    () => { RGDebug.LogInfo($"Uploaded gameplay session data"); },
+                    () => { RGDebug.LogError($"Failed to upload gameplay session data"); });
```
RG-nAmKcAz, line 172: We're already logging a more detailed error for this in RGServiceManager, we shouldn't log it twice

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -100,18 +103,92 @@ private void OnDestroy()
...
+                
+                // Next, upload the gameplay session screenshots
+                await _rgServiceManager.UploadGameplaySessionScreenshots(gameplaySessionId,
+                    _currentGameplaySessionDirectoryPrefix + "_screenshots.zip",
+                    () => { RGDebug.LogInfo($"Uploaded gameplay session screenshots"); },
+                    () => { RGDebug.LogError($"Failed to upload gameplay session screenshots"); });
```
RG-nAmKcAz, line 178: We're already logging a more detailed error for this in RGServiceManager, we shouldn't log it twice

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -100,18 +103,92 @@ private void OnDestroy()
...
+                
+                // Finally, upload the thumbnail
+                await _rgServiceManager.UploadGameplaySessionThumbnail(gameplaySessionId,
+                    _currentGameplaySessionDirectoryPrefix + "_thumbnail.png",
+                    () => { RGDebug.LogInfo($"Uploaded gameplay session thumbnail"); },
+                    () => { RGDebug.LogError($"Failed to upload gameplay session thumbnail"); });
```
RG-nAmKcAz, line 184: We're already logging a more detailed error for this in RGServiceManager, we shouldn't log it twice

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -100,18 +103,92 @@ private void OnDestroy()
...
-                    RGDebug.LogInfo($"Finished zipping replay to file: {_currentVideoDirectory}");
-                });
-                _isRecording = false;
+                    return;
+                }
+                
+                // Upload the gameplay session data
+                await _rgServiceManager.UploadGameplaySessionData(gameplaySessionId,
+                    _currentGameplaySessionDirectoryPrefix + "_data.zip",
+                    () => { RGDebug.LogInfo($"Uploaded gameplay session data"); },
```
RG-nAmKcAz, line 171: We should include more information if we are going to log this message, like the path

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -100,18 +103,92 @@ private void OnDestroy()
...
+                    () => { RGDebug.LogError($"Failed to upload gameplay session data"); });
+                
+                // Next, upload the gameplay session screenshots
+                await _rgServiceManager.UploadGameplaySessionScreenshots(gameplaySessionId,
+                    _currentGameplaySessionDirectoryPrefix + "_screenshots.zip",
+                    () => { RGDebug.LogInfo($"Uploaded gameplay session screenshots"); },
```
RG-nAmKcAz, line 177: We should include more information if we are going to log this message, like the path

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -100,18 +103,92 @@ private void OnDestroy()
...
+                    () => { RGDebug.LogError($"Failed to upload gameplay session screenshots"); });
+                
+                // Finally, upload the thumbnail
+                await _rgServiceManager.UploadGameplaySessionThumbnail(gameplaySessionId,
+                    _currentGameplaySessionDirectoryPrefix + "_thumbnail.png",
+                    () => { RGDebug.LogInfo($"Uploaded gameplay session thumbnail"); },
```
RG-nAmKcAz, line 183: We should include more information if we are going to log this message, like the path

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -100,18 +103,92 @@ private void OnDestroy()
...
+                    () => { RGDebug.LogError($"Failed to upload gameplay session thumbnail"); });
+            }
+            catch (Exception e)
+            {
+                Debug.LogError(e);
```
RG-nAmKcAz, line 188: Why are we hiding this exception ?  Use RGDebug and LogException, or take some code action here

RG-nAmKcAz (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
vontell: Fixed!

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
vontell: This will create all directories needed up to that folder, so it handles the parents as well.
> Creates all directories and subdirectories in the specified path unless they already exist.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
vontell: Now because of the other changes, they are no longer separated and are inside one folder, which now already exists since it holds the zip files, so I've moved this back to the original check.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
vontell: Added!

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
vontell: (discussed in Discord, this seems correct)

vontell (commented):
Addressed all comments I believe

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
RG-nAmKcAz: yes.. BUT.. checking that the parent folder exists does not confirm that the inner folders also exist, and we don't call to create the inner if the parent exists

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
RG-nAmKcAz: don't like it, but we can change later

RG-nAmKcAz (changes requested):
Directory creation check still needs a fix

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
rcornwall-reg: Could `RGEditorUtils.CreateAllAssetFolders(targetDirPath);` be the solution here?

RG-nAmKcAz: unfortunately not.. this needs to work at runtime and those utils utilize mostly assetdatabase apis (editor only)

src/gg.regression.unity.bots/Runtime/Scripts/RGServiceManager.cs
```diff
@@ -498,6 +498,119 @@ public async Task UploadLogs(long botInstanceId, string filePath, Action onSucce
...
+                    }
+                );
+            }
+            else
+            {
+                onFailure();
+            }
+        }
+        
+        public async Task UploadGameplaySessionData(long gameplaySessionId, string zipPath, Action onSuccess, Action onFailure)
```
rcornwall-reg, line 531: [Non-Blocking] UploadGameplaySessionData and UploadGameplaySessionScreenshots are almost identitical. We could create a single method with one extra `contentType` param:
```
private async Task UploadData(string uri, string filePath, string contentType, Action onSuccess, Action onFailure)
{
    if (LoadAuth())
    {
        await SendWebFileUploadRequest(
            uri: uri,
            method: "POST",
            filePath: filePath,
            contentType: contentType,
            onSuccess: (s) =>
            {
                RGDebug.LogDebug($"Response received for {uri}");
                onSuccess.Invoke();
            },
            onFailure: (f) =>
            {
                RGDebug.LogWarning($"Failed to upload for {uri}: {f}");
                onFailure.Invoke();
            }
        );
    }
    else
    {
        onFailure();
    }
}

public async Task UploadGameplaySessionData(long gameplaySessionId, string zipPath, Action onSuccess, Action onFailure)
{
    await UploadData(
        $"{GetRgServiceBaseUri()}/gameplay-session/{gameplaySessionId}/data",
        zipPath,
        "application/zip",
        onSuccess,
        onFailure);
}

// same thing for screenshots
```

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -86,7 +93,12 @@ public void Awake()
             DontDestroyOnLoad(gameObject);
             _this = this;
         }
-
+        
+        public void Start()
+        {
+            _rgServiceManager = RGServiceManager.GetInstance();
+            _rgBotManager = RGBotManager.GetInstance();
+        }
```
rcornwall-reg, line 101: I think it's bad practice to get these instances and store them in static variables.. We should either access them directly each usage with `.GetInstance()` or store them in private variables.
```
private RGServiceManager _rgServiceManager;
private RGBotManager _rgBotManager;

public void Start()
{
   _rgServiceManager = RGServiceManager.GetInstance();
   _rgBotManager = RGBotManager.GetInstance();
} 
```

rcornwall-reg (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
vontell: I see what you are saying now... I realize why overcomplicate it, and just call create directory no matter what since it will create if it doesn't exist, and be a no-op otherwise... Fixed!

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
vontell: If we access directly in the call sites, that will add an extra method access operator to the calls, which is why I prefer to store them in the variables.

But also, not sure if there is any functional difference for us here if we use a private static variable or a private class variable - since we are only ever creating one ScreenRecorder, I believe they are equivalent (modulo some weird C# stuff)

src/gg.regression.unity.bots/Runtime/Scripts/RGServiceManager.cs
vontell: I think this could be fine, but also since we are only doing this in two spots, I feel that an extra function may just add more redirection for readers of the code.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
rcornwall-reg: That's true - probably should have marked this as a non-blocking suggestion as well

rcornwall-reg (approved):
LGTM

PR #149, opened 2024-03-26
[REG-1669] Use API Keys rather than email/password

vontell:
Loom: https://www.loom.com/share/b7664270b94e4d20b6d9fc91dca4fbc3?sid=1243f474-ac89-461e-96ce-bfe1b8a76512

This PR removes any storage of email and password from project assets (in favor of an API key) and also stores this API key in a location that can be used by a build. More specifically, it has the following changes:

- Removed email and password storage from the existing RGUserSettings
- Consolidated all settings into a single RGSettings ScriptableObject
- Saves and loads the RGSettings asset into the resources folder, so that it is available in builds
- Updates the RGSettingsUIRegistrar functionality to have the API key field and links to get these
- Adds a section to the welcome screen that allows the user to login (before there was no way to get back to this screen)
- Simplifies a lot of the API code, because we no longer have to await for a login - we just assume the key is good to go.
- Removes some of the deprecated fields

Some design decisions:
- Because we don't really have users yet, I did not add code for migrating from the old system to the new system. Users will need to reconfigure.


vontell:
This relies on this PR: https://github.com/Regression-Games/RGBossRoom/pull/53

src/gg.regression.unity.bots/Runtime/Scripts/RGSettings.cs
```diff
@@ -16,27 +16,21 @@ public enum DebugLogLevel
...
-        [SerializeField] private uint nextBotId;
-#pragma warning disable CS0414 // suppress unused field warning
-        [SerializeField] private uint nextBotInstanceId;
-#pragma warning restore CS0414
         // ReSharper disable once InconsistentNaming
         [SerializeField] private bool feature_StateRecordingAndReplay;
+        
+        // Authentication settings
+        [SerializeField] private string rgHostAddress;
+        [SerializeField] private string apiKey;
```
RG-nAmKcAz, line 33: This is 'ok' for now, but long term this is not a secure way to preserve their API key in their project. We will want to find a better answer and/or encode/encrypt this field so that it isn't easily readable at rest.

src/gg.regression.unity.bots/Runtime/Scripts/RGServiceManager.cs
```diff
@@ -54,49 +54,29 @@ public bool IsAuthed()
...
-                    if (!string.IsNullOrEmpty(email) && !string.IsNullOrEmpty(password))
+                    if (!string.IsNullOrEmpty(settings.GetApiKey()))
                     {
-                        await Auth(
-                            email: email,
-                            password: password,
-                            onSuccess: s => { rgAuthToken = s; },
-                            onFailure: f => { }
-                        );
+                        rgAuthToken = settings.GetApiKey();
```
RG-nAmKcAz, line 75: this calls GetApiKey 2x, should only call it once

RG-nAmKcAz (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/RGSettings.cs
vontell: Yeah I was considering this originally when looking at email/password as well - if I encoded/decoded this as b64 for now, what are your thoughts on that?

src/gg.regression.unity.bots/Runtime/Scripts/RGServiceManager.cs
vontell: Good catch, will fix.

src/gg.regression.unity.bots/Runtime/Scripts/RGSettings.cs
```diff
@@ -70,35 +62,13 @@ public static RGSettings GetOrCreateSettings()
...
-#pragma warning restore CS0618 // Type or member is obsolete
-
-            // backwards compat for migrating RG devs before we had a single host address field
-            if (string.IsNullOrEmpty(_settings.rgHostAddress))
-            {
-                _settings.rgHostAddress = "https://play.regression.gg";
-#if UNITY_EDITOR
-                AssetDatabase.SaveAssets();
+#if !UNITY_EDITOR
+                Debug.LogWarning("RG settings could not be loaded. Make sure to log into Regression Games within the Unity Editor before building your project. For now, an empty user settings object will be used.");
```
rcornwall-reg, line 72: Let's use our internal `RGDebug.LogWarning`

src/gg.regression.unity.bots/Runtime/Scripts/RGSettings.cs
vontell: Ok so this was fun, but not trivial... the UI has some nice features for editor fields + serializable properties, but because we want the scriptable object to store the encoded version, I had to find a way to properly manage the UI and still encode and decode stuff properly. Check out the latest PR for this, I had one comment on a part I thought was fishy.

rcornwall-reg (changes requested):
Looking good! Definitely need a way to secure that API key, but like Zack said that can be another task

src/gg.regression.unity.bots/Runtime/Scripts/RGSettings.cs
vontell: Funnily enough I actually can't use that here I think, because it would cause an infinite recursion. RGDebug makes a call to get the RGSettings, and if it doesn't exist, it calls this debug function... which in turn calls RGSettings creation. So I think I'll have to keep it like this, since this is the case where the RGSettings doesn't load.

rcornwall-reg: Ah - makes sense!

rcornwall-reg (approved):

src/gg.regression.unity.bots/Runtime/Scripts/RGSettings.cs
```diff
@@ -120,9 +121,20 @@ public DebugLogLevel GetLogLevel()
...
+            get => encodedApiKey != null ? Encoding.UTF8.GetString(Convert.FromBase64String(encodedApiKey)) : null;
+            set => encodedApiKey = Convert.ToBase64String(Encoding.UTF8.GetBytes(value));
+        }
+
+        /**
+         * In some cases, like in the UI registrar, we need to store the API key in a serialized property, rather
+         * than through this class. This method allows us to encode the API key for storage in serialized properties
+         * without the calling location needing to know how we encode and decode.
+         */
+        public static string EncodeApiKeyForSerializedProperties(string decodedApiKey)
```
RG-nAmKcAz, line 135: I would leave these changes out of the current PR as was suggested in the reviews.

Encoding/Decoding Base64 does nothing to address the security of the field.  We should only be encoding to Base64 if we have non-serializable char values in the key.

As far as the structure of this code, the get/set for the ApiKey property should use common  Encode/Decode methods that the RGSettingsUIRegistrar would use.  Right now it has code duplication for reading an encoded key

RG-nAmKcAz (changes requested):

src/gg.regression.unity.bots/Editor/Scripts/RGSettingsUIRegistrar.cs
```diff
@@ -62,29 +67,14 @@ public static SettingsProvider CreateRGSettingsProvider()
...
-                        if (priorHost != hostField.stringValue || priorUser != emailField.stringValue || priorPassword != passwordField.stringValue)
-                        {
-                            priorHost = null;
-                            token = null;
-                            priorPassword = null;
-                            priorUser = null;
-                        }
+                        // This next line is purely for the UI - since we aren't using the apiKeyField.stringValue for
+                        // the UI, we need to update it within the settings. It then still gets saved in the line after.
+                        RGSettings.GetOrCreateSettings().ApiKey = tempApiKey;
```
vontell, line 75: @RG-nAmKcAz this is the fishy part - in order to make sure the UI actually shows the right value, since I am not using the direct property, I need to have a field that is not connected to the serialized property, and then set it later. From my testing, this seems to work entirely correctly, but we no longer will get undo capabilities on the editor field.

src/gg.regression.unity.bots/Runtime/Scripts/RGSettings.cs
vontell: Oh ok, I'll revert those commits, I thought the suggestion was that either a secure or non-secure approach was fine, just so people can't directly copy paste. And yes that was a mistake I have fixed in an upcoming commit, but I'll just remove that for now haha

vontell:
Reverted my changes for basic encoding to avoid simple copying, we can implement some proper encryption later.

src/gg.regression.unity.bots/Runtime/Scripts/RGServiceManager.cs
RG-nAmKcAz: this is unresolved

RG-nAmKcAz (changes requested):

vontell:
Ahhh sorry @RG-nAmKcAz, I reverted one commit too many. Fixed now.

RG-nAmKcAz (approved):

PR #150, opened 2024-04-02
[REG-1673] Performance Tweaks and Handling for World Space Game objects not being recorded correctly

RG-nAmKcAz:
- Cleans up some concurrency issues and other code bits from the Auth and Session uploading PRs
- Fixes the broken overlay from the session uploading PR
- Fixes an issue where in game objects weren't being properly included in the state
- Tweaks some performance aspects of state recording
- Optimizes how we identify and capture key frames especially around mouse inputs (This changes the recording JSON schema quite a bit)
  - Captures world space coordinates for in game object clicks to use for more accurate replay selection with player positional variance (see known bugs 1 below)
  - IF we hit UI elements, ignores the game elements behind them for the purposes of that click
  - fixes an issue where our start recording click was being captured in the state
  - determines the clicked on objects at recording time instead of replay time
- Restructures the code around how we record mouse inputs (Keyboard stuff will be dealt with later)
- Fixes some quirks with the virtual mouse cursor (hides it on start recording and load replay file)

Known bugs / deficiencies 
1. Handling variance in world object spawn positions better while still using world space coordinates <- whole multi-day task
2. We have an issue with scene transitions where we sometimes capture a bad recording.  This is most evident on the transition from CharacterSelect to BossRoom (DungeonEntrance) where scene elements are loaded in before unloading the prior scene... this is often done on purpose and is nearly indistinguishable from loading multiple scenes at the same time on purpose... I'll need some more experimentation time here later
3. Performance.. we have some performance issues with the large number of in game objects (specifically around trying to write out that much state info).  Will make addressing this a separate task.  I'm suspecting we'll need to minimize our state capture , maybe only to objects related specifically to triggering the keyframe... This may also harden our playback/replay as there is less to go wrong.. BUT.. would limit what we can validate / might not catch cases where the 'wrong' stuff is there if we're only checking for the 'right' stuf.

See also Bossroom PR (https://github.com/Regression-Games/RGBossRoom/pull/55)

In the demo (https://regressiongames.slack.com/archives/C044PCK6214/p1712164764781629?thread_ts=1712159587.525199&cid=C044PCK6214) ... note how the player gets a different spawn point the second time through, yet the replay still works

---


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseInputActionObserver.cs
```diff
@@ -78,126 +80,141 @@ public bool ButtonStatesEqual(object obj)
...
-        }
 
-        public void StopRecording()
-        {
-            _recording = false;
-        }
+                    var ray = Camera.main.ScreenPointToRay(mousePosition);
+                    mouseRayHits = Physics.RaycastNonAlloc(ray,
+                        _cachedRaycastHits,
+                        maxZDepth+0.1f);
```
addisonbgross, line 126:
```suggestion
```

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseInputActionObserver.cs
```diff
@@ -207,18 +224,63 @@ public MouseInputActionData GetCurrentMouseState()
...
+                result.Add(_priorMouseState);
+            }
+
+            return result;
+        }
+
+        private IEnumerable<RecordedGameObjectState> FindObjectsAtPosition(Vector2 position, List<RecordedGameObjectState> state, out float maxZDepth)
+        {
+            // make sure screen space position Z is around 0
+            var vec3Position = new Vector3(position.x, position.y, 0);
```
vontell, line 253: If for a given game this is not the case, is that ok?

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -93,16 +95,20 @@ public void Awake()
...
 
         private void OnDestroy()
         {
             _frameQueue?.CompleteAdding();
+            if (_fileWriteTasks.Count > 0)
+            {
+                RGDebug.LogInfo($"Waiting for unfinished file write tasks before stopping ["+ string.Join(",", _fileWriteTasks.Select(a=>a.Item1).ToArray())+ "]");
+            }
+            Task.WaitAll(_fileWriteTasks.Select(a=>a.Item2).ToArray());
+            _fileWriteTasks.Clear();
```
vontell, line 111: With the new file writing approach, any worry about memory? I suppose we aren't in the realm of worrying about that yet?

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -112,84 +118,103 @@ private void OnDestroy()
 #endif
             if (_isRecording)
             {
-                HandleRecordingSaving();
+                KeyboardInputActionObserver.GetInstance()?.StopRecording();
+                _mouseObserver.ClearBuffer();
+                _isRecording = false;
+                _ = HandleEndRecording(_tickNumber, _startTime, DateTime.Now, _currentGameplaySessionDataDirectoryPrefix, _currentGameplaySessionScreenshotsDirectoryPrefix, _currentGameplaySessionThumbnailPath);
```
vontell, line 124: Nit, non-blocking: we are passing along a lot of class variables that can just be referenced in the helper function itself.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -112,84 +118,103 @@ private void OnDestroy()
...
+                    () => { RGDebug.LogInfo($"Uploaded gameplay session screenshots from {screenshotsDirectoryPrefix}.zip"); },
+                    () => { });
+
+                // Finally, upload the thumbnail
+                await RGServiceManager.GetInstance().UploadGameplaySessionThumbnail(gameplaySessionId,
+                    thumbnailPath,
+                    () => { RGDebug.LogInfo($"Uploaded gameplay session thumbnail from {thumbnailPath}"); },
+                    () => { });
+            }
+            catch (Exception e)
```
vontell, line 204: Non-blocking nit - I walked through the call stack for these API calls and couldn't find a scenario where an exception would be thrown - everything would be caught and referred to the onFailure invocations (which is a spot where an exception could happen, but all of ours are empty functions). Guessing I missed something? I ended up leaving it out since I found it misleading in a lot of my failed API calls.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
```diff
@@ -234,37 +267,56 @@ private RecordedGameObjectState CreateStateForTransform(int screenWidth, int scr
...
+                if (includeOnlyOnCameraObjects)
+                {
+                    var yBuffer = 0;
+                    var xBuffer = 0;
+                    if (replay)
+                    {
+                        // look a bit outside the view on replay to account for edge cases and small intentional variances in the game code
+                        yBuffer = screenHeight / 2;
+                        xBuffer = screenWidth / 2;
+                    }
```
addisonbgross, line 291: I'm wondering if we should be concerned about false positives by using these buffers. They seem a bit large.
Are we able to get with with smaller buffers?

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
```diff
@@ -234,37 +267,56 @@ private RecordedGameObjectState CreateStateForTransform(int screenWidth, int scr
...
                     }
                 }
 
                 if (onCamera)
                 {
-                    var size = new Vector3(maxX - minX, maxY - minY);
-                    var center = new Vector3(minX + (maxX - minX) / 2, minY + (maxY - minY) / 2);
+                    // make sure the screen space bounds has a non-zero Z size around 0
+                    // we track the true z offset separately for ease of mouse selection on replay
+                    var size = new Vector3(maxX - minX, maxY - minY, 0.1f);
```
vontell, line 308: (non-blocking) Dangerous bookkeeping idea for later speed improvements - we are doing lots of object allocations in this code - new vectors, lists... might be able to recycle the objects later on. But def not this PR, and plus that can get messy.. just brainstorming a bit

vontell (approved):
All of my comments ended up being very nit-picky nit picks, so LGTM! No changes required, appreciate tidying up some of my code.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseInputActionObserver.cs
RG-nAmKcAz: this has nothing to do with any specific game..

when projecting world space to screen space, the numbers always come with a Z offset from the camera

for simplicity of comparison to mouse clicks, I exclude that information from what is recorded and handle the max zOffset as a separate field on the object.  This allows us to do Bounds.contains for mouse positions without it erroneously returning false

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
RG-nAmKcAz: This approach isn't new.  I simply added the tracking to ensure all files are written before we move on to completing the recording as I found cases where that wasn't happening properly with big json/images

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
RG-nAmKcAz: no they can't... the very reason I changed this is that we are calling async functions and were relying on class level variables which very well may change while that async is running

abeizer (changes requested):
The delay in recording (to prevent capturing the initial "start recording" click) may be too aggressive. Here's a demonstration:

In this loom, I create a recording where I want to cycle through some of the characters during character selection in BossRoom. I give the tool a few moments to start recording and then select "archer" before going on to select other characters and then ending the recording. The playback expects the archer to already be selected rather than performing that click as part of the playback.

https://www.loom.com/share/010ace77d02845d3b5508c69ce1f4a98?sid=c29faf7a-ceb8-4291-8f02-fafac1a4dd8b

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
RG-nAmKcAz: reasonable conversation point, but the very reason I put this in here for evaluation only during replay was that I was having the player spawn position break replays because the object it needed was offscreen

addisonbgross: That makes sense. I guess its something we can tweak as we apply this to more use cases 👍🏻

addisonbgross (approved):
Pending @abeizer 's remark, this looks good

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
RG-nAmKcAz: always some good things to keep in mind looking for performance

These allocations are currently all required as we're building the full state observed for this frame and each game object needs uniquely allocated information until the end of the frame.  Every PR I optimize a bit further, but re-using state objects between frames isn't one I've gotten to you (it's on my list).  As you'll see in the screenshot though, the Unity API calls we have to make actually do most of the allocation overhead.

The biggest areas we die on performance today
1. Serializing to JSON .. I've optimized this a ton along the way, almost to the point of no-where left to optimize short of just writing less data. <- called ONLY ON KEY FRAMES
2. Order of magnitude less than writing JSON, but next biggest thing is indeed around this function <- called EVERY FRAME

image

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseInputActionObserver.cs
vontell: Ah ok, I misread the comment, that makes more sense to me now

RG-nAmKcAz:
> The delay in recording (to prevent capturing the initial "start recording" click) may be too aggressive. Here's a demonstration:
> 
> In this loom, I create a recording where I want to cycle through some of the characters during character selection in BossRoom. I give the tool a few moments to start recording and then select "archer" before going on to select other characters and then ending the recording. The playback expects the archer to already be selected rather than performing that click as part of the playback. If I click the archer myself, the playback can pick up from there.
> 
> https://www.loom.com/share/010ace77d02845d3b5508c69ce1f4a98?sid=c29faf7a-ceb8-4291-8f02-fafac1a4dd8b

As far as I can tell, what you're seeing has nothing to do with the 1 frame delay before recording mouse input.  Starting a huddle / need your data

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
vontell: 🤦 my sincerest apologies, of course... this train ride debacle this morning has me more scrambled than I like

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
vontell: Got it, good to know! I do wonder when it would make sense to switch to some binary format like protobuf, in my past experience it helps a ton with size for sure, and I believe it's gotta be more CPU performant...

vontell (approved):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
vontell: For our future reference, here is some benchmarking someone did: https://dotnetcoretutorials.com/protobuf-in-c-net-part-4-performance-comparisons/?expand_article=1

abeizer (approved):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
RG-nAmKcAz: Find me a cross-platform unity safe protobuf package and I'll give it a try.. I went down that rabbit hole for a few minutes one day recently before going.. nope, not spending that level of dev time yet

PR #152, opened 2024-04-06
[REG-1686] Fix resource directory creation issue, fix sample

vontell:
Fixes an issue found during UAT: When the resources folder doesn't exist, we need to create it before making the settings asset.

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/RGSettings.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/RGSettings.cs
@@ -1,4 +1,5 @@
 using System;
+using System.IO;
 using UnityEditor;
 using UnityEngine;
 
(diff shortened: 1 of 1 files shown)


RG-nAmKcAz (approved):

src/gg.regression.unity.bots/Runtime/Scripts/RGSettings.cs
```diff
@@ -18,8 +19,9 @@ public class RGSettings : ScriptableObject
     {
         // We store this asset in the Resources folder so that it can be accessed from any build
         // Note that it can only be created in the editor, however.
+        private static readonly string SETTINGS_DIRECTORY = "Assets/Resources";
         private static readonly string SETTINGS_RESOURCE_NAME = "RGSettings";
-        private static readonly string SETTINGS_PATH = $"Assets/Resources/{SETTINGS_RESOURCE_NAME}.asset";
+        private static readonly string SETTINGS_PATH = $"{SETTINGS_DIRECTORY}/{SETTINGS_RESOURCE_NAME}.asset";
```
addisonbgross, line 24: do we need to construct this path string in a platform-generic way? `/` vs `\`

addisonbgross (approved):
LGTM
1 small, non-blocking comment

abeizer (approved):

src/gg.regression.unity.bots/Runtime/Scripts/RGSettings.cs
RG-nAmKcAz: luckily no, the api accepts / prefers `/`

PR #157, opened 2024-04-11
[REG-1692] Detect changes in GameFace rendering from frame to frame

RG-nAmKcAz:
- Basic code to detect delta's in the game face html rendering layer texture so we can mark Key frames when the UI changes.
- Converts the render texture into a 256x256 grayscale+alpha pixel array and compares to the prior frame
- Handles GameFace detection without requiring compile dependencies on gameface
  - Tested in BossRoom
 
**NOTE: This DOES NOT handle replay yet, only recording**

---


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/GameFaceDeltaObserver.cs
```diff
@@ -0,0 +1,152 @@
...
+        private const int _grayHeight = 256;
+        private byte[] _grayArray = new byte[_grayWidth * _grayHeight * 2];
+
+        private bool _firstRun = true;
+        private bool _isRecording = false;
+
+        // uses 0=false, 1=true to do interlocked threadsafe updates
+        public int _gameFaceDelta;
+
+        private RenderTexture _cothmlViewTexture;
```
abeizer, line 23: Not blocking but you know I have to point out typos...

```suggestion
        private RenderTexture _cohtmlViewTexture;
```

abeizer (approved):
Very interested to see how this performs. Since we're scaling down the texture, I assume this means that slight variation in greyscale values won't flag a key frame so we're more likely to pick up "significant" changes to the texture

vontell (approved):

RG-nAmKcAz:
Part 2 PR included these changes and was merged to avoid hellscape of rebasing/merging/etc

PR #162, opened 2024-04-19
[REG-1680] Optimize performance of state for current frame

RG-nAmKcAz:
- DOES NOT address JSON serialization performance
- Heavily micro-optimizes object constructions, pooling, reuse, loop iterations, and other code to limit GC allocs and improve performance.. kinda turns the code into a thread safety nightmare.. but luckily.. unity requires all this to run on the main thread.. so we can cheat a lot to get better performance and re-use structures
- Also handles replay performance optimization both for getting current state and for analyzing key frames and processing mouse inputs
- Fixes issues with dynamic resolution scaling on replay related to mouse click and release points <- this was actually annoying to fix :/ 
- Reduces flicker of keyframe wait text by delaying presentation of said text for a few seconds worth of frames if not matched
- Updates file loading code to load async and unblock the game ui.  I have a task for a future progress indicator here, but for now it disables the overlay buttons and lets the game resume until it finishes loading

BEFORE (Boss room main menu recording)
OLD

AFTER (Boss room main menu recording)
FINAL_AFTER

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
@@ -501,8 +501,8 @@ namespace RegressionGames.StateRecorder
                 {
                     BehaviourState cObject = new BehaviourState
                     {
-                        path = (ts ??= GetUniqueTransformPath(childTransform)).Path,
-                        name = (ts ??= GetUniqueTransformPath(childTransform)).TypeFullName,
+                        path = (ts ??= GetUniqueTransformPath(childTransform, childBehaviour)).Path,
+                        name = (ts ??= GetUniqueTransformPath(childTransform, childBehaviour)).TypeFullName,
                         state = childBehaviour
                     };
 
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using JetBrains.Annotations;
 using StateRecorder.Types;
@@ -418,19 +419,23 @@ namespace RegressionGames.StateRecorder
                     _currentParentTransforms.Clear();
                     while (_nextParentTransforms.Count > 0)
                     {
-                        // swap the list references
+                        var nptCount = _nextParentTransforms.Count;
+                        for (var i = 0; i < nptCount; i++)
+                        {
+                            var currentTransform = _nextParentTransforms[i];
+                            ProcessChildTransformComponents(currentTransform, behaviours, collidersState, rigidbodiesState);
+                        }
+                        // swap the list references and load the next layer
                         (_currentParentTransforms, _nextParentTransforms) = (_nextParentTransforms, _currentParentTransforms);
                         _nextParentTransforms.Clear();
                         var currentCount = _currentParentTransforms.Count;
                         for (var i = 0; i < currentCount; i++)
                         {
                             var currentParentTransform = _currentParentTransforms[i];
-                            var childCount = _currentParentTransforms[i].childCount;
+                            var childCount = currentParentTransform.childCount;
                             for (var j = 0; j < childCount; j++)
                             {
-                                var currentChildTransform = currentParentTransform.GetChild(j);
-                                ProcessChildTransformComponents(currentChildTransform, behaviours, collidersState, rigidbodiesState);
-                                _nextParentTransforms.Add(currentChildTransform);
+                                _nextParentTransforms.Add(currentParentTransform.GetChild(j));
                             }
                         }
                     }
@@ -646,18 +651,24 @@ namespace RegressionGames.StateRecorder
                             _currentParentTransforms.Clear();
                             while (_nextParentTransforms.Count > 0)
                             {
-                                // swap the list references
+                                var nptCount = _nextParentTransforms.Count;
+                                for (var i = 0; i < nptCount; i++)
+                                {
+                                    var currentTransform = _nextParentTransforms[i];
+                                    ProcessChildTransformComponents(currentTransform, behaviours, null, null);
+                                }
+
+                                // swap the list references and load the next layer
                                 (_currentParentTransforms, _nextParentTransforms) = (_nextParentTransforms, _currentParentTransforms);
                                 _nextParentTransforms.Clear();
                                 var currentCount = _currentParentTransforms.Count;
                                 for (var i = 0; i < currentCount; i++)
                                 {
                                     var currentParentTransform = _currentParentTransforms[i];
-                                    var childCount = _currentParentTransforms[i].childCount;
+                                    var childCount = currentParentTransform.childCount;
                                     for (var k = 0; k < childCount; k++)
                                     {
                                         var currentChildTransform = currentParentTransform.GetChild(k);
-                                        ProcessChildTransformComponents(currentChildTransform, behaviours, null, null);
                                         _nextParentTransforms.Add(currentChildTransform);
                                     }
                                 }
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
@@ -5,6 +5,7 @@ using StateRecorder.Types;
 using UnityEngine;
 using Component = UnityEngine.Component;
 // ReSharper disable ForCanBeConvertedToForeach - indexed for is faster and has less allocs than enumerator
+// ReSharper disable LoopCanBeConvertedToQuery
 
 namespace RegressionGames.StateRecorder
 {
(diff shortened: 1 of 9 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
```diff
@@ -306,109 +348,286 @@ private RecordedGameObjectState CreateStateForTransform(bool replay, int screenW
...
                     // make sure the screen space bounds has a non-zero Z size around 0
                     // we track the true z offset separately for ease of mouse selection on replay
-                    var size = new Vector3(maxX - minX, maxY - minY, 0.1f);
-                    var center = new Vector3(minX + size.x / 2, minY + size.y / 2, 0);
-                    var screenSpaceBounds = new Bounds(center, size);
+                    var extents = new Vector3((maxX - minX)/2, (maxY - minY)/2, 0.05f);
+                    var center = new Vector3(minX + extents.x, minY + extents.y, 0);
+
+                    var worldExtents = new Vector3((maxWorldX - minWorldX)/2, (maxWorldY - minWorldY)/2, (maxWorldZ - minWorldZ)/2);
+                    var worldCenter = new Vector3(minWorldX + worldExtents.x, minWorldY + worldExtents.y, minWorldZ + worldExtents.z);
```
addisonbgross, line 391: Are these `Vector3`s worth pre-allocating + reusing?

RG-nAmKcAz: could I, I guess technically yes, but not in a reasonable way

The way the unity code is setup for Bounds/Extents/etc you get a copy of the Vector3 instead of being able to reference the value.  

Take for example this sample code and note how changing the value of x on the extents does not affect in internal bounds extents object.  So while yes we 'can' do it, we unfortunately don't actually save any new object creations.

image

addisonbgross: makes sense. Thanks for clarifying

addisonbgross (approved):
The content here is a bit over my head, but the logic looks sound

abeizer (approved):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ReplayDataPlaybackController.cs
```diff
@@ -133,6 +134,12 @@ public void Stop()
 
         private double CurrentTimePoint => Time.unscaledTime - _lastStartTime + _priorKeyFrameTime ?? 0.0;
 
+        // used to avoid spam building/showing the keyframe text until at least a few frames have gone past
+        private int _checkOfKeyFrameCount = 0;
+
+        // allow up to 360 frames before saying it failed, at 120 fps this is 3 seconds, at 100fps this is ~3.6 seconds , at 60fps this is 6 seconds , at 50 fps this is ~7.2 seconds, at 30 fps this is 12 seconds
+        private const int KeyFrameChecksBeforePrompting = 360;
```
vontell, line 141: (non-blocking) Should this be a public variable that uses can edit? Might be nice to control the timeout (for instance, maybe I know that my network latency results in a highly-varied wait time)

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -298,74 +298,174 @@ public void StartRecording()
 
         }
 
-        private KeyFrameType[] GetKeyFrameType(List<RecordedGameObjectState> priorState, List<RecordedGameObjectState> currentState, string pixelHash)
+        /**
+         * <summary>Modifies the lists with removal leaving only the unique entries in each list.  Assumes lists have unique entries as though created from a hashset</summary>
+         */
+        private bool IntListsMatch(List<int> priorScenes, List<int> currentScenes)
```
vontell, line 304: I didn't want to add this comment since it really doesn't matter and I don't want to clog up the PR, but I had fun pondering if this approach or a sort + compare would be faster... with some constants included, sorting would be 2(nlogn + n) while this approach would potentially be n^2, but I bet with the `break;` and size of these lists, the practical analysis isn't as straightforward... 

Actually, an even faster approach, since the lists have unique entries, it might be even faster to just convert each list to a set, and then hash the sets (you'd have to implement that yourself though and that math is probably kinda weird), and compare the hashes... which would be O(n) I think... 

Once again, completely ignore this probably, I just miss doing this sort of algorithmic thinking

vontell (approved):
Like the optimizations here! Especially the re-use of objects, nice work, this didn't seem easy...

I'm guessing for most practical cases, the size of these lists is sufficient for our use cases? Just something we should be aware of as a potential issue in the future.

No blocking comments, I had fun reviewing this though!

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
vontell: There is also https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.setequals?view=net-8.0

RG-nAmKcAz: Tried and evaluated all these suggestions and more in many places in the code, in practice, given the relatively small sizes of our lists (normally in the 100s or 1000s, and < 10000) entries, the looping iteration is both faster and avoids lots of GC allocations that hashing does in all cases.

In this specific case the list is always less than 10 elements, so anything other than just checking is 1 to 2 orders of magnitude longer time in practice and incurs GC allocations.

Also.. on sorting .. I had the same thought that some of these n^2 cases.. sorting it first would be faster as I could quick search, but in practice the math comparisons and allocations required to do the sorting still took more time than just iterating for our data set sizes.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ReplayDataPlaybackController.cs
RG-nAmKcAz: in short no, this is a nicety to avoid some screen flicker but does not affect whether this system functions or not

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/StateRecorderUtils.cs
```diff
@@ -0,0 +1,89 @@
...
+            for (var i = 0; i < listCount; i++)
+            {
+                if (list[i] == theScene)
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
```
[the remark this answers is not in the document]
RG-nAmKcAz: Linq.. has some very nasty things said about it online.  It is definitely NOT Java streams :/

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/StateRecorderUtils.cs
```diff
@@ -0,0 +1,89 @@
...
+            {
+                if (list[i] == theInt)
+                {
+                    list.RemoveAt(i);
+                    return true;
+                }
+            }
+
+            return false;
+        }
```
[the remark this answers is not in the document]
RG-nAmKcAz: Removing at end by index would ALWAYS be faster than any remove action requiring object value comparisons

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
```diff
@@ -177,8 +193,12 @@ private RecordedGameObjectState CreateStateForTransform(bool replay, int screenW
 
             var hasVisibleRenderer = false;
 
-            foreach (var nextRenderer in renderers)
+            var psCount = priorState?.Count ?? -1;
+
+            var rendererListLength = _rendererQueryList.Count;
+            for (var i = 0; i < rendererListLength; i++)
```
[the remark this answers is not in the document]
RG-nAmKcAz: .Count is the number of elements in the list, not the list's allocated size

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
vontell: got it, makes perfect sense! Thanks for entertaining my comment XD

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ReplayDataPlaybackController.cs
vontell: Ahhh I see, thanks for the clarification.

RG-nAmKcAz:
@vontell about your question in the top on the size of the lists.  The lists are pre-sized to avoid extra allocations in most cases, but if we do exceed the size of the list, c# will automatically allocate more space for it and that 1 tick will just be a bit slower

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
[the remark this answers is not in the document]
vontell: See your comment now, nice! I think that answers all my questions.

PR #166, opened 2024-05-01
[REG-1708] Use real paths for UI object clicks

RG-nAmKcAz:
- Now writes the 'true' path to `path` and the 'normalized' path (what we had before without uniqueness numbers) to `normalizedPath`...
  **- This is NOT strictly backwards compatible, but in a non-breaking way... The data in `path` is similar, but not the same from the client UI perspective**
  - This negatively impacts performance as we are now computing and writing out 2 similar strings for every object... if we determine that we don't need this on the Client UI, I'd LOVE to remove `normalizedPath` as I can compute it on replay load easily for handling mouse clicks ... if we remove it though.. similar game objects cloned from prefabs will no longer have the same `normalizedPath` to work from in the client UI validations
- Uses the `normalizedPath` for game object and `path` for UI objects to properly handle weird UI resolution scaling changes like BossRoom has
- Further cleans up String to Json patterns
- Fixes some bugs on mouse replay clicks when doing multiple recordings or replays in the same game engine run 
- Fixes a copy/pasta issue with the JSON for keyboard inputs

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Text;
 using JetBrains.Annotations;
 using StateRecorder.Types;
 using UnityEngine;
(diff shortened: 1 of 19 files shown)


addisonbgross:
I think we going to need both the `normalizedPath` and `path`.

When people are creating scenarios, they're going to want to refer to generalized gameobjects like: "When 2 Imps' Health < 40" (`normalizedPath` for the Imps). When users are using the state + screenshot viewer they may want to view specific gameobjects in a scene, like specific Imps (`path` for a specific Imp).

vontell (approved):

addisonbgross (approved):

addisonbgross:
I have a note to switch the use of `path` for `normalizedPath` in the client. I'll make that PR whenever this gets merged

PR #171, opened 2024-05-07
[REG-1739] Handle concurrency issue with tick #

RG-nAmKcAz:
- Fixes a bug where concurrent code was referring to the incorrect tick number
- Cleans up some naming between frame and tick
- Hardens (belt/suspenders) code paths around async gpu capture... even though this wasn't the issue in and of itself

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
@@ -59,7 +59,7 @@ namespace RegressionGames.StateRecorder
         private DateTime _startTime;
 
         private BlockingCollection<((string, long), (byte[], int, int, GraphicsFormat, Color32[], Action))>
-            _frameQueue;
+            _tickQueue;
 
         private readonly List<(string, Task)> _fileWriteTasks = new();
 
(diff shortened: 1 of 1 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -594,11 +594,16 @@ private IEnumerator RecordFrame()
...
-                        RGDebug.LogException(e, "Exception capturing state for frame");
+                        RGDebug.LogException(e, $"Exception capturing state for tick # {_tickNumber}");
                     }
 
+                    var didQueue = 0;
+
                     if (jsonData != null)
                     {
+                        // save this off because we're about to operate on a different thread :)
+                        var currentTickNumber = _tickNumber;
```
RG-nAmKcAz, line 605: This guy right here... this is the actual fix for the bug.  Need a copy of the `_tickNumber` for the async code to use.  It was currently getting the `_tickNumber` field which had already been updated by the next frame processing.

vontell: Good catch!

vontell (approved):

PR #173, opened 2024-05-08
Update docs links and getting started description

vontell:
Updates the relevant docs links in the package to use the URLs. Also, adds a slightly more relevant first box in the welcome screen.


RG-nAmKcAz:
**Looks Good**

PR #175, opened 2024-05-15
[REG-1696] [REG-1697] Cleanup SDK Toolbar and Support Looping

RG-nAmKcAz:
- Compacts the layout of our SDK toolbar
- Supports looping a replay until failure or playing the same replay again w/out having to reselect the files
  - Includes a looping count UI element
- Changes the folder names to be more meaningful including datetime and sessionIds
- Changes the default UI element detection method to Delta, where it handles delta amounts from the prior tick that are non exact.  Thus if you went up in count of some path, it passes if you go up, even if the count isn't exact.
- Puts back in the code that forces the 'new' input system temporarily.  We can/should remove this with @svolokh 's legacy input system work (`ReplayDataPlaybackController.SetupEventSystem`)
- Fixes some bugs/deficiencies around screen space bounds computation when using canvas modes other than screen space overlay.. this should get us very close on world space canvases also
- Fixes a bug where particle renderers attached to UI elements would make the UI element appear as a world space object in the recording
- Fixes the icon pulsing equation and adds a text pulsing behaviour
- Adds pulsing status text for both data upload and data load from file
- Updates the 'success' graphic to align with the play button as a green highlight
- Fixes some threading bugs waiting on the async GPU data at end of recording
- Fixes an issue where our virtual mouse cursor event handler wasn't being managed correctly and sometimes the cursor wouldn't move
- Fixes an issue where sometimes we wouldn't move the world space key frame forward when blocked on the UI key frame
- Fixes an issue where sometimes we would drop input from key frames when ui and world space key frames were far apart 

image

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
@@ -143,8 +143,7 @@ namespace RegressionGames.StateRecorder
                 tPath = tName;
                 tPathNormalized = tNameNormalized;
                 var parent = theTransform.parent;
-                // don't need overloaded unity != null-ness check as if we were found alive, our parent is alive
-                while (parent is not null)
+                while (parent != null)
                 {
                     _tPathBuilder.Clear();
                     tPath = _tPathBuilder.Append(parent.gameObject.name).Append("/").Append(tPath).ToString();
@@ -649,6 +648,7 @@ namespace RegressionGames.StateRecorder
         {
             _priorStates.Clear();
             _newStates.Clear();
+            _frameNumber = -1;
         }
 
         /**
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Text;
 using JetBrains.Annotations;
 using StateRecorder.Types;
@@ -33,8 +34,8 @@ namespace RegressionGames.StateRecorder
         private static readonly List<RigidbodyRecordState> _rigidbodyStateObjectPool = new (100);
         private static readonly List<Rigidbody2DRecordState> _rigidbody2DStateObjectPool = new (100);
 
-        private static readonly List<ColliderRecordState> _emptyColliderStateList = new(0);
-        private static readonly List<RigidbodyRecordState> _emptyRigidbodyStateList = new(0);
+        private static readonly IList<ColliderRecordState> _emptyColliderStateList = ImmutableList.Create<ColliderRecordState>();
+        private static readonly IList<RigidbodyRecordState> _emptyRigidbodyStateList = ImmutableList.Create<RigidbodyRecordState>();
 
         private static InGameObjectFinder _this;
 
@@ -422,8 +423,6 @@ namespace RegressionGames.StateRecorder
                     }
                     resultObject.screenSpaceZOffset = maxZ;
 
-
-
                     var behaviours = resultObject.behaviours;
                     _behaviourStateObjectPool.AddRange(behaviours);
                     behaviours.Clear();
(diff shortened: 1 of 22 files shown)


RG-nAmKcAz:
This leaves us in a state where we can replay Parhelion up until the mouse over effect on their card.  Our input events do NOT trigger OnMouseOver even though we force the correct input system into the game on replay :/ 

Note: If you manually move the mouse to trigger the mouseover, the replay continues but almost always hangs a few steps later looking for the wrong debuffs on npcs since the spells are random.

RG-nAmKcAz:
Apologies.. this PR got HUGE .  I tried to itemize all the key bullet points.  I know the code for the replay delta stuff is NOT clean, but I suspect we will be back in here soon to performance optimize and/or add more dynamic function anyway.

vontell:
Will try to find time to review this today!

svolokh (approved):
Everything looks good to me from my perspective.

Non-blocking: Why is ReplayToolbarManager in a namespace called "Unity.Multiplayer.Samples.BossRoom"? I was surprised to find a new import for BossRoom in the ScreenRecorder class, but then noticed this, so I suppose it wouldn't break the code.

RG-nAmKcAz:
@svolokh ... I don't know why it was in that namespace but it has been fixed :) ... it was created there incorrectly from the beginning (*shameful look*)...

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/InGameObjectFinder.cs
```diff
@@ -675,142 +682,212 @@ public void Cleanup()
...
-                                // load the next layer as we go
-                                var currentCount = _currentParentTransforms.Count;
-                                for (var i = 0; i < currentCount; i++)
-                                {
-                                    var currentTransform = _currentParentTransforms[i];
-                                    // only process down so far.. if we hit another ui component.. stop traversing that part of the tree
-                                    var hasRenderer = currentTransform.GetComponent(typeof(CanvasRenderer)) != null;
-                                    if (!hasRenderer)
+                                    var worldExtents = new Vector3((worldMax.x - worldMin.x) / 2, (worldMax.y - worldMin.y) / 2, 0.05f);
+                                    var worldCenter = new Vector3(worldMin.x + worldExtents.x, worldMin.y + worldExtents.y, 0f);
```
vontell, line 804: I forget, why is it that the z component is set this way here?

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -291,11 +262,16 @@ private IEnumerator StartRecordingCoroutine(string referenceSessionId)
 
                 Directory.CreateDirectory(stateRecordingsDirectory);
 
+                var prefix = referenceSessionId != null ? "replay" : "recording";
+                var pf = _currentSessionId.Substring(Math.Max(0, _currentSessionId.Length - 6));
+                var postfix = referenceSessionId != null ? pf + "_" + referenceSessionId.Substring(Math.Max(0, referenceSessionId.Length - 6)) : pf;
+                var dateTimeString =  System.DateTime.Now.ToString("MM-dd-yyyy_HH.mm");
```
vontell, line 268: Nice! I suppose a next step for us (i'll track this somewhere) is to introduce some name indexing on the backend.

vontell (approved):
Awesome work! Especially excited to see the introduction of some new modes for comparisons. Not really any change requests, all non-blocking, mostly questions for my own understanding.

Not sure if you want to do it as part of this ticket or make a new ticket, but we should also update the documentation to reflect the looping, strict mode changes, and new file naming for recordings.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ReplayDataPlaybackController.cs
```diff
@@ -252,35 +355,47 @@ private string CheckKeyFrameState(int keyFrameCheckCount, Dictionary<int,Recorde
...
 
-                            if (uiElements == null || uiElements.Count == 0)
+                            if (uiElements != null)
                             {
-                                // only bail out here if we are where the ui is the oldest awaited key frame
-                                if (eldestKeyFrame.tickNumber == uiObjectKeyFrame.tickNumber)
+                                if (uiElements.TryGetValue(hashCode, out var count))
                                 {
-                                    if (keyFrameCheckCount < KeyFrameChecksBeforePrompting)
+                                    if (myReplayEnforcementMode == UIReplayEnforcement.Strict && count.Item2 <= 0)
```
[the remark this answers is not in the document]
RG-nAmKcAz: There are a few evaluations for UI elements.. 

first is to go through the current state and subtract from the keyframe count any matches found in the current state

then evaluate based on the enforcement mode and those counts whether we had the correct new state amount for each path

and then a final pass is taken to check for anything that should have been in the current state that isn't based on the keyframe counts


The meaning of the counts was attempted to be documented on the enum values of `StateElementDeltaType`

PR #176, opened 2024-05-15
[REG-1729] Capture memory and CPU utilization

svolokh:
This is the bare minimum implementation of capturing CPU utilization (in terms of time spent on the main thread), and memory utilization (amount of bytes the operating system reports in use by the application). 

This uses the [ProfilerRecorder](https://docs.unity3d.com/ScriptReference/Unity.Profiling.ProfilerRecorder.html) API available in recent versions of Unity (2021.3+). It begins profiling when the recording starts, and stops profiling when the recording ends. 

The memory usage is obtained via reading the profiler counter for *System Used Memory* (which is defined by the Unity docs as the memory usage reported by the OS: https://docs.unity3d.com/Manual/ProfilerMemory.html). 

CPU time is obtained in terms of nanoseconds spent on the main thread. The "Main Thread" stat appears to be the best one I can find for measuring this based on my read of https://docs.unity3d.com/ScriptReference/Unity.Profiling.ProfilerRecorder.html and https://docs.unity3d.com/Manual/ProfilerCPU.html.

I have tested this both in-editor and in standalone build and have been able to get the CPU time and memory to appear in the state dump JSON files. Links to example JSON files:
[In-Editor](https://github.com/Regression-Games/RGUnityBots/files/15326121/5.json)
[Standalone](https://github.com/Regression-Games/RGUnityBots/files/15326123/5.json)

Some considerations that may need further discussion:
1. Is it reasonable to require Unity 2021.3+ to support this approach to profiling, given that this is a newer API? Should macros be added to check the Unity version and conditionally disable the profiler for older versions?
2. Is "time spent on the main thread per frame" a reasonable metric for CPU usage, or were the clients looking for something more like core percentage reported by the OS? (I do not see a way to directly obtain the latter via this API, but perhaps it could be approximated).
3. The profiler gives very different results for in-editor vs standalone builds, due to its stats including both the game and editor.
4. Because the profiler is started only when the recording starts, the CPU/memory usage on the first frame is listed as zero. 
5. For CPU time, this implementation currently only reads the last frame, however this can fluctuate a lot across frames. Would it be better to instead report an average of the last N frame's times (or since the last tick)?
6. There are a *lot* of profile markers/counters available via this API. It could be worth exploring making it a configurable setting to specify which are of interest. Here is a list I extracted via the API: [profiler.txt](https://github.com/Regression-Games/RGUnityBots/files/15326168/profiler.txt)

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ProfilerObserver.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ProfilerObserver.cs
@@ -12,7 +12,7 @@ namespace StateRecorder
         // Used heap size (in bytes) that is garbage collected
         public long? gcUsedMemory;
 
-        // Time spent (in nanoseconds) by the CPU on the main thread since the last tick
+        // Time spent (in milliseconds) by the CPU on the main thread since the last tick
         public double? cpuTimeSincePreviousTick;
     }
 
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ProfilerObserver.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ProfilerObserver.cs
@@ -12,8 +12,8 @@ namespace StateRecorder
         // Used heap size (in bytes) that is garbage collected
         public long? gcUsedMemory;
 
-        // Time spent (in milliseconds) by the CPU on the main thread since the last tick
-        public double? cpuTimeSincePreviousTick;
+        // Time spent (in nanoseconds) by the CPU on the main thread since the last tick
+        public long? cpuTimeSincePreviousTick;
     }
 
     public class ProfilerObserver : MonoBehaviour
@@ -52,9 +52,9 @@ namespace StateRecorder
         /**
          * Computes the sum of the last N values of the round-robin sample buffer.
          */
-        private static double SumOfLastFrames(ProfilerRecorder recorder, List<ProfilerRecorderSample> samples, int numFrames, out int framesRead)
+        private static long SumOfLastFrames(ProfilerRecorder recorder, List<ProfilerRecorderSample> samples, int numFrames, out int framesRead)
         {
-            double sum = 0.0;
+            long sum = 0;
             framesRead = 0;
             for (int frameIndex = recorder.Count - 1; frameIndex >= 0 && framesRead < numFrames; --frameIndex)
             {
@@ -90,11 +90,11 @@ namespace StateRecorder
                 _cpuTimeSampleBuf.Clear();
                 _cpuTimeRecorder.CopyTo(_cpuTimeSampleBuf);
                 int framesRead;
-                double cpuTime = SumOfLastFrames(_cpuTimeRecorder, _cpuTimeSampleBuf, frameCountSinceLastTick,
+                long cpuTime = SumOfLastFrames(_cpuTimeRecorder, _cpuTimeSampleBuf, frameCountSinceLastTick,
                     out framesRead);
                 if (framesRead == frameCountSinceLastTick) // only report the total cpuTime if there were sufficient frames recorded for the request
                 {
-                    result.cpuTimeSincePreviousTick = cpuTime * 1e-6; // convert to milliseconds
+                    result.cpuTimeSincePreviousTick = cpuTime;
                 }
             }
             return result;
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ProfilerObserver.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ProfilerObserver.cs
@@ -1,4 +1,5 @@
-﻿using Unity.Profiling;
+﻿using System.Collections.Generic;
+using Unity.Profiling;
 using UnityEngine;
 
 namespace StateRecorder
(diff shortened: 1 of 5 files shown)


vontell:
Thank you for adding this! In response to your questions:

- Unity 2021.3+ is fine, this seems to match our projects so far, so I think that restriction is fine.
- Time spent on CPU so far is fine with me, but I will include that in our questions to Playable Worlds this week :+1:
- I like the idea of getting the average across frames, but would this be too much of a performance hit?

svolokh:
@vontell I don't think it would have much impact on performance, as far as I can tell each sample is just storing a long integer value. The profiler API has a "capacity" option to remember more samples (keep the last N values), there's an example of that usage here https://docs.unity3d.com/ScriptReference/Unity.Profiling.ProfilerRecorder.html. So, it seems feasible to add in an "avgCpuTime" stat.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ProfilerObserver.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace StateRecorder
+{
+    public struct ProfilerObserverResult
+    {
+        // Amount of memory (in bytes) the operating system reports in use by the application
+        public long? systemUsedMemory;
+
+        // Time spent (in nanoseconds) by the CPU on the main thread on the current frame
+        public long? cpuTime;
```
RG-nAmKcAz, line 12: To answer part of the discussion, this for sure needs to be an accumulation since the last tick, not just for the current frame when we captured.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/RecordingAndReplayContainerTypes.cs
```diff
@@ -133,6 +135,16 @@ public void WriteToStringBuilder(StringBuilder stringBuilder)
             IntJsonConverter.WriteToStringBuilder(stringBuilder, framesSincePreviousTick);
             stringBuilder.Append(",\"fps\":");
             IntJsonConverter.WriteToStringBuilder(stringBuilder, fps);
+            if (cpuTime.HasValue)
```
RG-nAmKcAz, line 138: We should probably just use `WriteToStringBuilderNullable` on `LongJsonConverter` which handles writing `null` into the json payload.  We generally haven't excluded fields from the json when null, instead we write the null.

RG-nAmKcAz (changes requested):
A few change requests to address that are blocking


... 
Non Blocking Food For Thought
...

Touching on the bigger topics,  the memory usage looks fine / as expected to me.  We may want to consider also showing the GC memory (https://docs.unity3d.com/ScriptReference/Unity.Profiling.ProfilerRecorder.html).

The CPU time I don't think will be that helpful to users.  If we get it to be for the whole tick instead of 1 frame, then we have something that might be useful as you could at least compare time between ticks to time by cpu, but it would still be quite hard to make use of.  We don't have a way to extrapolate this value to % cpu used by the system and thus can't really tell if the system is cpu bound, or if our cpu time is low because of long render times / stalling /etc or if jobs or other threads are starving the cpu.

I think its fine as a first pass on cpu, but we'll need a more system accurate solution long term.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ProfilerObserver.cs
svolokh: I have updated this to to be an accumulation since the last tick, and have renamed the field to `cpuTimeSincePreviousTick` for clarity. The average CPU time per frame can now be calculated by dividing this value by `framesSincePreviousTick`.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/RecordingAndReplayContainerTypes.cs
svolokh: Updated to use this function instead.

svolokh:
> Touching on the bigger topics, the memory usage looks fine / as expected to me. We may want to consider also showing the GC memory (https://docs.unity3d.com/ScriptReference/Unity.Profiling.ProfilerRecorder.html).

This is easy to do, so I went ahead and added the used GC memory as "gcMemory".

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ProfilerObserver.cs
svolokh: Side note: The current implementation requires specifying an upper bound on the number of profile frames to keep in memory (MAX_FRAMES_ACCUM). I've set this to 16k based on my perception of what is a reasonable upper bound. As far as I can tell storing this profiling data is not very expensive.

I did initially experiment with periodically flushing the profiling buffer to avoid this requirement, but I was running into issues of double-counting or losing profiling data. So, currently if `framesSinceLastTick` exceeds MAX_FRAMES_ACCUM it will just report `null`.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ProfilerObserver.cs
```diff
@@ -0,0 +1,103 @@
...
+{
+    public struct ProfilerObserverResult
+    {
+        // Amount of memory (in bytes) the operating system reports in use by the application
+        public long? systemUsedMemory;
+
+        // Used heap size (in bytes) that is garbage collected
+        public long? gcUsedMemory;
+
+        // Time spent (in milliseconds) by the CPU on the main thread since the last tick
```
RG-nAmKcAz, line 15: One remaining question/concern, sorry I didn't notice this the first time

This says `in milliseconds` but is a `double`.  Is this supposed to be an integer type (ie: `long`) ? or is this time actually measured in some other unit ?

RG-nAmKcAz (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ProfilerObserver.cs
svolokh: This was something I changed as part of this PR. The profiler outputs in `long` nanoseconds, but I found that for large frame counts summing everything up would cause an overflow in the JSON output (even after converting it to double for some reason). So, I now output in milliseconds double to avoid this problem.

RG-nAmKcAz: Hmm, this doesn't seem right.  Lets figure out why you were seeing overflow because we should be able to add those longs without any real issue since they aren't that large.

Max c# signed `long` is
9,223,372,036,854,775,807

A whole day in nanoseconds is
86,400,000,000,000

svolokh: Ok so, the following code gives me "-2147483648" as output:
`DoubleJsonConverter.WriteToStringBuilderNullable(sb, 86_400_000_000_000.0);`

So, maybe there's a bug there?

Using `LongJsonConverter` works fine though, so I suppose given your example it could make sense to continue using longs for summing up the time (I was basically just following the Unity docs approach to summing up the profiler outputs as shown here: https://docs.unity3d.com/ScriptReference/Unity.Profiling.ProfilerRecorder.html)

RG-nAmKcAz: Yeah, lets go with longs for tracking/summing it up so we don't overflow or have weird FP behavior or format

svolokh: Updated to use long nanoseconds again.

RG-nAmKcAz (approved):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ProfilerObserver.cs
```diff
@@ -0,0 +1,103 @@
+﻿using System.Collections.Generic;
+using Unity.Profiling;
+using UnityEngine;
+
+namespace StateRecorder
```
RG-nAmKcAz, line 5: @svolokh I didn't catch this, but this namespace is wrong.

Should be `RegressionGames.StateRecorder`... When I pulled in these changes before fixing the namespace, my project won't compile.  Please open another PR (same linear ticket is fine by me) asap to correct this

svolokh: @RG-nAmKcAz The main branch seems to compile fine on my end currently. I saw that you had performed the namespace fix in your newest PR, do you think it would be cleaner to just have this be part of the that PR's changes?

PR #178, opened 2024-05-22
[REG-1762] Fix error message showing paths not hashes

RG-nAmKcAz:
- Fixes a bug where we showed the hash id instead of the pathname in keyframe errors

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ReplayDataContainer.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ReplayDataContainer.cs
@@ -276,7 +276,7 @@ namespace RegressionGames.StateRecorder
                             }
                             else
                             {
-                                uiElementsCounts[hashCode] = (val.Item1, 1);
+                                uiElementsCounts[hashCode] = (replayGameObjectState.path, 1);
                             }
                         }
                         else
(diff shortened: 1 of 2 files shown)


svolokh (approved):
Non-blocking: I think it would be good to have a more descriptive name than `theVal` for the changes in question. I had to dig deep into the code to figure out what `theVal.Item1` is supposed to represent, which made it somewhat challenging to validate this PR.

Also, I had a hard time making sense of this code from ReplayDataContainer. Is there a bug here? Why is the else branch still reading from `val` if `TryGetValue` returned false? https://github.com/Regression-Games/RGUnityBots/blob/7b7d380dc27133527727e04726493e2965ca9cec/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ReplayDataContainer.cs#L273-L280

Why is it reading

RG-nAmKcAz:
@svolokh please don't approve the PR if you find valid bugs like that, use request changes on those lines

what you reported is a real bug, i'll look into a fix, but please update this to request changes

svolokh:
@RG-nAmKcAz noted! The lines I noted actually were not technically part of this PR's changes, so I don't have a way to directly reference them from within the "Files Changed" tab. I will change my review to request changes.

RG-nAmKcAz:
@svolokh Good catch, should be resolved now

svolokh (approved):

vontell (approved):

abeizer (approved):

PR #179, opened 2024-05-22
[REG-1716] Animation Rigging Extension Package

RG-nAmKcAz:
- Adds an extension package to support Animation Rigging

- Minor Inconvenience, but seems fine --- Unfortunately, Unity insists on not allowing packages to refer to other packages by git url... thus, users must manually add the core package for this to work.  You can't just add the extension pack and have it magically add the core package.

- If we are ok with this package style, I will add a documentation task to start tracking how to install our extension packages

---


vontell (approved):
Nice! Can you provide an example of the steps though for how this would be added? I see the bossroom example, but from your description in this PR, I was now confused on what the step is for adding this.

RG-nAmKcAz:
@vontell You add it similarly to how you add our base package

from URL https://github.com/Regression-Games/RGUnityBots.git?path=src/gg.regression.unity.bots.animation.rigging#v0.0.19

The package will add just fine on its own, but won't compile correctly until you also add our base package

vontell:
> @vontell You add it similarly to how you add our base package
> 
> from URL https://github.com/Regression-Games/RGUnityBots.git?path=src/gg.regression.unity.bots.animation.rigging#v0.0.19
> 
> The package will add just fine on its own, but won't compile correctly until you also add our base package

Ahhhh got it, awesome! That sounds great to me

svolokh (approved):
Looks good!

One thing I'm wondering about: for my legacy input manager work, I've needed to enable the `overrideReferences` flag to be able to pull in the `Mono.Cecil.Pdb.dll` library, which is an extension library for reading/writing PDB debug symbols so that we can preserve them after instrumentation. Do you think this will cause an issue for the extension system have here? (example shown below from my branch)

https://github.com/Regression-Games/RGUnityBots/blob/4da1412860112c94a901ce7d8cd8a858617ce720/src/gg.regression.unity.bots/Editor/Scripts/RegressionGames.Editor.asmdef#L11-L15

RG-nAmKcAz:
> Looks good!
> 
> One thing I'm wondering about: for my legacy input manager work, I've needed to enable the `overrideReferences` flag to be able to pull in the `Mono.Cecil.Pdb.dll` library, which is an extension library for reading/writing PDB debug symbols so that we can preserve them after instrumentation. Do you think this will cause an issue for the extension system have here? (example shown below from my branch)
> 
> https://github.com/Regression-Games/RGUnityBots/blob/4da1412860112c94a901ce7d8cd8a858617ce720/src/gg.regression.unity.bots/Editor/Scripts/RegressionGames.Editor.asmdef#L11-L15

Since the extension system includes no pre-compiled items, I would highly doubt it.  Everything in the extension will be pulled in as source and built as part of their build.

PR #181, opened 2024-05-23
[REG-1758] Support capture and replay of inputs handled via Input.GetKey()

svolokh:
This is the first step towards supporting the legacy input manager in the gameplay recordings, by supporting Input.GetKey().

Key parts of this PR:
1. **RGLegacyInputInstrumentation**: This hooks into the Unity build process and performs the instrumentation both when the editor first launches, and when assemblies are being compiled (either in-editor or standalone). If the game is not using legacy inputs, the assemblies remain untouched.
2. **RGLegacyInputWrapper**: The instrumentation replaces all calls to UnityEngine.Input APIs with calls into this class. 
3.  ReplayDataPlaybackController now simulates inputs through the legacy input manager as well via RGLegacyInputWrapper.SimulateKey().

I have tested this against a determinized version of Tetris (the game can be found here: https://github.com/Regression-Games/LegacyInputGames/). Here is a video of the inputs being simulated: https://www.loom.com/share/faaaecc6a9d747bba18c04c4319c43c0. I have also verified that there were no regressions with RGBossRoom.

For now, I have kept in the logic that puts in an InputSystemUIInputModule, since this does not affect this functionality and is not yet covered by what is currently supported by this initial implementation.

When any instrumentation occurs, an info debug message will be printed to the console indicating the affected assemblies:
image

---


src/RGUnityBots/Assets/Tests/Runtime/RGLegacyInputUtilityTests.cs
```diff
@@ -0,0 +1,40 @@
+﻿#if ENABLE_LEGACY_INPUT_MANAGER
+using System.Collections.Generic;
+using RegressionGames;
+using NUnit.Framework;
+using RegressionGames.RGLegacyInputUtility;
+using UnityEngine;
+
+namespace Tests.Runtime
+{
+    public class RGLegacyInputUtilityTests
```
RG-nAmKcAz, line 10: 👍

src/gg.regression.unity.bots/Editor/Scripts/RGLegacyInputUtility/RGLegacyInputInstrumentation.cs
```diff
@@ -0,0 +1,232 @@
...
+        {
+            // Safer to do any initial instrumentation within the editor update loop.
+            // Initial experiments with running directly from static initializers
+            // caused errors sometimes.
+            EditorApplication.update += SetUpHooks;
+        }
+
+        static void SetUpHooks()
+        {
+            if (EditorApplication.timeSinceStartup > 2.0)
```
RG-nAmKcAz, line 34: Timing things scare me, what is this trying to avoid ?

src/gg.regression.unity.bots/Editor/Scripts/RGLegacyInputUtility/RGLegacyInputInstrumentation.cs
```diff
@@ -0,0 +1,232 @@
...
+                EditorApplication.update -= SetUpHooks;
+                CompilationPipeline.assemblyCompilationFinished += OnAssemblyCompiled;
+                InstrumentExistingAssemblies();
+            }
+        }
+
+        private static bool IsAssemblyIgnored(string assemblyPath)
+        {
+            string fileName = Path.GetFileName(assemblyPath);
+            return fileName.Contains("RegressionGames")  // ignore RG assemblies
```
RG-nAmKcAz, line 45: We may want to put a developer focused README.md into the `Plugins` folder of gg.regression.unity.bots to remind people to update this file whenever they add a new package.

src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        private static void ScheduleRemoveNewKey(KeyCode keyCode)
+        {
+            _removeNewCoroutines.Add(keyCode, _simulationContext.StartCoroutine(RemoveNewKey(keyCode)));
+        }
+        
+        private static void ClearKeyState(KeyCode keyCode)
+        {
+            if (_removeNewCoroutines.TryGetValue(keyCode, out Coroutine coro))
```
RG-nAmKcAz, line 87: we can replace `TryGetValue` with `Remove` to avoid accessing the dictionary twice here

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ReplayDataPlaybackController.cs
```diff
@@ -2,8 +2,9 @@
...
 using System.Linq;
 using StateRecorder;
-using TMPro;
-using UnityEditor.AnimatedValues;
+#if ENABLE_LEGACY_INPUT_MANAGER
+using RegressionGames.RGLegacyInputUtility;
+#endif
 using UnityEngine;
 using UnityEngine.EventSystems;
 using UnityEngine.InputSystem;
```
RG-nAmKcAz, line 10: Check line ~95 .. I think we need to take back out the code that errors / auto injects the new UI Input System

RG-nAmKcAz (changes requested):
Some small changes in various places.

Non-Blocking
I did not comment on anything frame timing of inputs related, will cover that in Slack and/or Linear ticket REG-1768

src/gg.regression.unity.bots/Editor/Scripts/RGLegacyInputUtility/RGLegacyInputInstrumentation.cs
svolokh: The point of this is to do any initial instrumentation within the editor loop at some arbitrary point after the editor has started. When the package is first loaded, it iterates through any existing assemblies to check if they need to be instrumented (`InstrumentExistingAssemblies()`).

On my first attempt, I tried doing this immediately upon package load from within a static initializer, which caused the Unity runtime to refuse to execute some methods related to fetching the list of assemblies.

It is possible that this code would be perfectly fine without the timeSinceStartup check, but it seems safer with it and would still catch all cases: 1) Any assemblies that were compiled prior to this code running would be handled by `InstrumentExistingAssemblies()`; 2) Any assemblies compiled after this code runs would be checked by the `OnAssemblyCompiled` callback.

svolokh: @RG-nAmKcAz As I suspected, the timeSinceStartup check is not necessary here, so I've removed it and everything still works fine (verified on several different games).

RG-nAmKcAz (approved):

src/RGUnityBots/Packages/manifest.json
```diff
@@ -12,6 +12,7 @@
     "com.unity.timeline": "1.6.4",
     "com.unity.ugui": "1.0.0",
     "com.unity.visualscripting": "1.8.0",
+    "com.unity.nuget.mono-cecil": "1.11.4",
```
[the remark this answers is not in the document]
svolokh: This is just a Unity package that adds the Mono.Cecil assemblies. So, the only implication I can think of is a corner case where (1) the game project is already using Mono.Cecil, and (2) they have manually added the Mono.Cecil assemblies to their project directory rather than using the package provided by Unity. This would cause an assembly conflict, and I would suggest using the UPM package in this case.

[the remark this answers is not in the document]
svolokh: Also, to answer the question more broadly about the instrumentation: it happens behind the scenes as an added post-processing step to the build pipeline, and only makes any changes if legacy input API usage is detected. Any assemblies that do not use it will not be affected, and even for those that we do modify we preserve the debug information. So, from the developer's perspective no action is needed / everything should look the same.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ReplayDataPlaybackController.cs
```diff
@@ -613,8 +617,36 @@ private void PlayInputs(Dictionary<int, RecordedGameObjectState> priorStates, Di
             _mouseQueue.RemoveAll(a => a.IsDone);
         }
 
+#if ENABLE_LEGACY_INPUT_MANAGER
```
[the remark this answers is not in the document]
svolokh: This is a flag that is built into the Unity engine (since around version 2019.2, as far as I can tell). https://docs.unity3d.com/2021.3/Documentation/Manual/PlatformDependentCompilation.html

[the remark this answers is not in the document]
svolokh: So, for a project that was already using the legacy input manager, it will be automatically enabled. Note that it is possible for both ENABLE_LEGACY_INPUT_MANAGER and ENABLE_INPUT_SYSTEM to be enabled together if the project happens to be using both systems.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ReplayDataPlaybackController.cs
```diff
@@ -613,8 +617,36 @@ private void PlayInputs(Dictionary<int, RecordedGameObjectState> priorStates, Di
...
+                    RGDebug.LogError($"Unexpected key state {upOrDown}");
+                    break;
+            }
+        }
+#endif
+
         private void SendKeyEvent(long tickNumber, Key key, KeyState upOrDown)
         {
+            #if ENABLE_LEGACY_INPUT_MANAGER
+            SendKeyEventLegacy(tickNumber, key, upOrDown);
```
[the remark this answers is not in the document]
svolokh: It is possible for both the old and new input systems to be used together (or for the ENABLE_LEGACY_INPUT_MANAGER flag to be enabled even if only the new input system is being used), so for now it is best to continue generating events through the new input system as well. 

Eventually we will want to separate out the RG package's dependency on the new input system (REG-1764), which we still use for recording inputs, and once we do so then we can conditionally enable/disable which input system to generate events for based on the values of the ENABLE_INPUT_SYSTEM and ENABLE_LEGACY_INPUT_MANAGER flags.

PR #183, opened 2024-05-29
[REG-1706] Allow recordings to be used programmatically in Unity Test Runner

vontell:
Adds an `RGTestUtils` class that assists in playing back recordings in a scene. Here is a loom video and an example test file using this new API, but you can check out the link below to the RGBossRoom project for a full example. After recording, I verified that the new recording is properly saved.

LOOM: https://www.loom.com/share/dae3882584f042f5a2db866a5253f862?sid=df1c3c01-af58-4b1b-ba9f-9c50ac7481d0

Example:

```csharp
public class RecordingTest
{

    [UnityTest]
    public IEnumerator TestRecording()
    {

        // Allow error logs to be ignored
        LogAssert.ignoreFailingMessages = true;

        // Define which recording to use
        string recordingPath = System.IO.Path.Combine(System.IO.Directory.GetCurrentDirectory(), "Assets/Tests/Runtime/Recordings/MainMenuRecording.zip");

        // Wait for the scene
        SceneManager.LoadSceneAsync("Startup", LoadSceneMode.Single);
        yield return RGTestUtils.WaitForScene("MainMenu");

        // Start the playback, which will also create a new recording
        string saveLocation = null;
        yield return RGTestUtils.StartPlaybackFromFile(recordingPath, result => saveLocation = result);

        // Print out the recording path for viewing later
        RGDebug.LogInfo("Played back and recorded the test run - saved to " + saveLocation);
        Assert.IsNotNull(saveLocation);

    }
}
```

BossRoom PR: https://github.com/Regression-Games/RGBossRoom/pull/57
Docs PR: https://github.com/Regression-Games/RegressionDocs/pull/73

Questions for reviewers:
- I always struggle with yield return null - here I was trying to get it to yield a final string that is the saved recording path, but for some reason I could never get it to work. Any suggestions here?


src/gg.regression.unity.bots/Runtime/Scripts/RGTestUtils.cs
```diff
@@ -0,0 +1,68 @@
...
+                    var timeElapsed = Time.time - startTime;
+                    if (timeElapsed > timeout)
+                    {
+                        break;
+                    }
+                }
+                yield return null;
+            }
+            yield return null;
+            Assert.IsTrue(loaded);
```
svolokh, line 41: I would recommend adding a `message` parameter to this assertion to more clearly indicate to the developer the reason for the assertion failure (e.g. $"Scene {sceneName} failed to load within {timeout} seconds").

src/gg.regression.unity.bots/Runtime/Scripts/RGTestUtils.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+        
+        
+        /// <summary>
+        /// Plays back an existing recording, and then returns the save location of the recording.
+        /// </summary>
+        /// <param name="recordingPath">The path to the recording to play back (the full data.zip file path)</param>
+        /// <param name="setSaveLocation">A callback that will be called with the save location of the recording</param>
+        public static IEnumerator StartPlaybackFromFile(string recordingPath, Action<string> setSaveLocation)
```
svolokh, line 50: I think using a callback here is a reasonable approach to returning values. I do not believe there is a clean way to directly return values from coroutines in the way you were intending 

A suggestion: You may want to replace the `Action<string>` callback with a more general `Action<PlaybackEvent>` callback, since in the future you may want to notify the testing script with more types of events besides just "playback finished". For example, you may want to allow the test script to be notified whenever a new keyframe is encountered.

svolokh (approved):
Looks good to me. I have provided non-blocking feedback below.

src/gg.regression.unity.bots/Runtime/Scripts/RGTestUtils.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+        
+        /// <summary>
+        /// Waits until a specific scene has been loaded, and asserts that it has been loaded
+        /// </summary>
+        /// <param name="sceneName">The name of the scene to wait for</param>
+        /// <param name="timeout">The maximum time to wait for the scene to load (in seconds). Defaults to 5.</param>
+        public static IEnumerator WaitForScene(string sceneName, int timeout = 5)
+        {
+            var startTime = Time.time;
```
RG-nAmKcAz, line 24: `Time.time` is scaled time and affected by the game settings.

You probably want `Time.unscaledTime` (https://docs.unity3d.com/ScriptReference/Time-unscaledTime.html)

RG-nAmKcAz (changes requested):
One small change in addition to @svolokh 's comments

RG-nAmKcAz (approved):

PR #187, opened 2024-06-03
[REG-1763] Support mouse inputs handled via MonoBehaviour methods + instrumentation bugfix

svolokh:
This PR introduces support for simulating inputs for the legacy input mouse event handlers found in MonoBehaviour methods:
`OnMouseDown()`
`OnMouseDrag()`
`OnMouseEnter()`
`OnMouseExit()`
`OnMouseOver()`
`OnMouseUp()`
`OnMouseUpAsButton()`

This emulates the same logic used by the Unity runtime for invoking these methods. The Unity engine handles these in the same module as the legacy input handling APIs, so I consider them to be part of the legacy input as well and implement their handling in the `RGLegacyInputWrapper` event loop.

Also, this PR includes a small bugfix/optimization for the legacy input instrumentation to compare assemblies by name instead of path (this was causing unnecessary assemblies to be processed when creating standalone builds, since the assembly output path changes for these).

Test cases (these games can be found in LegacyInputGames):
AngryBirds:
https://www.loom.com/share/2dccda235ab946378c771228f3abd973

Match3:
https://www.loom.com/share/d7a0c9842b354219969fbe913f84f5ca
_Caveat for Match3_: The click correction was causing the playback to miss for this game, so for this one I was just using the simple record/playback [script](https://gist.github.com/svolokh/60f3afefef35cd2d9e6cdf25273fbb11) I made a while ago. Perhaps an issue worth investigating at some point. Regardless, this illustrates that the legacy input simulation is working.

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs
@@ -61,6 +61,11 @@ namespace RegressionGames.RGLegacyInputUtility
         private static Vector3 _mousePosDelta;
         private static Vector2 _mouseScrollDelta;
         private static Dictionary<InputManagerEntry, RGLegacyAxisInputState> _axisStates;
+        private static Camera[] _camerasBuf;
+        private static (GameObject, Camera)? _lastHitObject;
+        private static (GameObject, Camera)? _mouseDownObject;
+        private static (GameObject, Camera)? _lastHitObject2D;
+        private static (GameObject, Camera)? _mouseDownObject2D;
         private static RGLegacyInputManagerSettings _inputManagerSettings;
         private static Coroutine _inputSimLoopCoro;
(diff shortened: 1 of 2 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs
```diff
@@ -259,6 +272,53 @@ private static IEnumerator InputSimLoop()
...
+                            Ray mouseRay = camera.ScreenPointToRay(_mousePosition);
+                            if (Physics.Raycast(mouseRay, out RaycastHit hit, maxDistance: Mathf.Infinity,
+                                    layerMask: cameraRaycastMask))
+                            {
+                                hitObject = (hit.collider.gameObject, camera);
+                            }
+                        }
+                        
+                        // 2D raycast
+                        {
```
RG-nAmKcAz, line 305: if we hit a 3d object, should we still be doing the 2d object check ?

RG-nAmKcAz (approved):
Approved, but with one skepticism/question

vontell (approved):

PR #193, opened 2024-06-11
[REG-1780] Add all parent objects to state for tree view on UI

RG-nAmKcAz:
- Adds a parentId field to all objects in the state data
  - Some objects may have a NULL parentId, if they do, they are at the top level of their respective scene hierarchy
- Includes all 'filler' parent objects to fill out the tree between renderable objects
  - Non-renderable or Off-Camera objects will have NULL screenSpaceBounds
  - DOES NOT include information about any children below the lowest child level objects with renderers
- **WARNING** (@addisonbgross) Removes the path/normalizedPath fields from the nested information like behaviours/colliders/rigidbodies/etc
  - If this will BREAK the UI, I can split it off into a separate PR to merged after the UI changes, just let me know

See also https://github.com/Regression-Games/RGBossRoom/pull/58 , which makes BossRoom spawn point deterministic

---


addisonbgross:
@RG-nAmKcAz the removal of `path` and `normalizedPath` from behaviours will indeed cause an issue. If you don't mind separating that out into anoher PR, that would be awesome 👍🏻

abeizer:
With the removal of `path` and `normalizedPath`, will this imply that any behaviours have the same path values as the entity?

vontell (approved):
Seems reasonable to me, but I'll leave @addisonbgross to review the particulars of the approach for his use case.

addisonbgross (approved):
Pending my comment about keeping the `path` and `normalizedPath` in behaviours/colliders/etc. this looks good

svolokh (approved):

RG-nAmKcAz:
@abeizer 
`With the removal of path and normalizedPath, will this imply that any behaviours have the same path values as the entity?`

Correct.  These fields are removed because those things will only/always appear on their exact gameObject in the state.

RG-nAmKcAz:
@addisonbgross path removal changes undone for this PR

Still be aware, you will be able to get NULL for screenSpaceBounds where the old code did not allow that... IOW... The UI might still get something it doesn't expect because of this PR, but not as catastrophic as the paths being gone :D

PR #208, opened 2024-06-25
[REG-1797] Action manager skeleton and monkey bot

svolokh:
This is the first PR for the action analysis work, which includes the monkey bot, data structures, action types, and in-editor UI for configuring the actions made available to the bots. With this PR it is possible to start up monkey testing bots from the Regression Games Bot Manager overlay, as well as configure the actions taken via an in-editor UI.

For this skeleton version, currently the actions are specified via a class implementing `IRGActionProvider`. The next milestone will be automating the action provider via analysis of the game code and resources, which will enable using the monkey bot with any keyboard/mouse based game.

Here is a narrated video showcasing the current capabilities included in this PR: https://www.loom.com/share/f930ea7f9eb84d9192be99285fca90a1

Here is a run-through of the various games I have tested the monkey bot with:
https://www.loom.com/share/159dfb92158f4097b305e23ff514ed04

Key components of this PR:
1. **ActionManager/RGActionManager.cs**: This is the API that the bots call into to obtain a model of the valid actions available in the current game frame.
2. **GenericBots/RGMonkeyBot.cs**: This is a monkey bot that implements the IRGBot interface and interacts with the action manager API to randomly test the game. It is always available through the Regression Games Bot Manager overlay and can be started/stopped at any arbitrary game state.
3. **ActionManager/Actions/*.cs**: All the files in this folder implement the variety of different common action types suitable for the typical keyboard and mouse actions found in Unity games.
4. **ActionManager/RGActionManagerUI.cs**: This adds a menu item “Regression Games -> Configure Bot Actions” and provides a window where the user can view and enable/disable actions used by the bot.

Here is a gist with all the action providers if you want to experiment with the monkey bot: https://gist.github.com/svolokh/0212f2262ee1b992b13c59b72f2a1f70. The legacy input games are available in the LegacyInputGames repository. Because this is a temporary setup until the automated analysis is implemented, I did not add these to the game repos. You can use these by just adding the provider MonoBehaviour to the first scene of the game.


src/gg.regression.unity.bots/Runtime/Scripts/GenericBots/RGMonkeyBot.cs
```diff
@@ -0,0 +1,114 @@
+﻿using System.Collections.Generic;
+using RegressionGames.ActionManager;
+using UnityEngine;
+
+namespace RegressionGames.GenericBots
+{
+    public class RGMonkeyBot : MonoBehaviour, IRGBot
+    {
+        public float actionInterval = 0.0f; // every frame by default
```
RG-nAmKcAz, line 9: I could see arguments either way, but I'd make this something like 0.1f by default.  That way the tester can at least see what is happening.

src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs
```diff
@@ -185,19 +191,19 @@ private static IEnumerator InputSimLoop()
                             Vector3 newMousePosition = evt.newMousePosition;
                             if (evt.newMouseDelta.HasValue)
                             {
-                                _mousePosDelta = evt.newMouseDelta.Value;
+                                _mousePosDelta += evt.newMouseDelta.Value;
```
RG-nAmKcAz, line 194: this being += doesn't seem right at a glance, but I'm probably not understanding the code here

RG-nAmKcAz (changes requested):
Some small changes / clarifications in the code, and some general comments / questions overall.
- We can do this in a follow up PR, but we need to handle overlapping buttons.
  - I did a POC of this in a recent branch if you want some sample code, but when we disable a button, we should not click in that 'area' if another button overlapped that button, otherwise we may click that disabled button anyway by accident.
- Have you tested parhelion project steelline to make sure the clicking works in their game ?  (I ask because I have had intermittent issues click in their game when not using replay)
- Do we handle keyboard input into fields when we click on them and/or do we detect when input fields are active when choosing what key inputs are valid?  I didn't see anything get typed in the replay demo videos

svolokh:
@RG-nAmKcAz 
> * We can do this in a follow up PR, but we need to handle overlapping buttons.
>   
>   * I did a POC of this in a recent branch if you want some sample code, but when we disable a button, we should not click in that 'area' if another button overlapped that button, otherwise we may click that disabled button anyway by accident.

That's a good point, I'm just using the center coordinate currently (for both game objects and mouse objects). I created a Linear task and will incorporate this in a future PR. 

> * Have you tested parhelion project steelline to make sure the clicking works in their game ?  (I ask because I have had intermittent issues click in their game when not using replay)

I don't currently have access to that game, I will ask about it.

> * Do we handle keyboard input into fields when we click on them and/or do we detect when input fields are active when choosing what key inputs are valid?  I didn't see anything get typed in the replay demo videos

Not yet. The main emphasis on this PR has been gameplay keyboard/mouse actions (such as Input.GetKey, Input.mousePosition, Keyboard.current, Mouse.current). The Unity UI support currently is really basic (only does button presses) just so it can get through the menus. I plan to do a more thorough coverage of the Unity UI components in the second milestone, which will involve adding more action types besides the `UIButtonPressAction`.

src/gg.regression.unity.bots/Runtime/Scripts/GenericBots/RGMonkeyBot.cs
svolokh: I was going back and forth on this, since the faster interval generally covers more functionality faster. But the longer interval has some advantages for some of the games too. I've set it to 0.1f.

src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs
svolokh: The problem I was trying to address here is that the monkey bot might generate multiple mouse pos/scroll delta events within a single frame, and I want to have those stack up rather than only using the value of the last one. The deltas would get cleared anyways at the start of the following frame at the bottom of the event processing loop.

Though, as I write this I realize that this new behavior is probably inconsistent with how the new Input System would handle these events :/. I'll revert it back to how it was. Thanks for pointing it out!

RG-nAmKcAz (approved):

vontell (approved):
Approving this, and I know this is still in the prototyping phase, but @svolokh @RG-nAmKcAz it might be good for you two to chat about how we might want to get this into the bot segment system as well.

PR #212, opened 2024-06-28
[REG-1793] Initial ECS code

RG-nAmKcAz:
- Completely restructures the existing state recording and playback code to handle transforms and entities almost interchangeably
- Adds a new ECS SDK module to encapsulate the ECS code
  - **Add something like this to your packages manifest to enable this code ( "gg.regression.unity.bots.ecs": "file:C:/Users/zack/RG/RGUnityBots/src/gg.regression.unity.bots.ecs")**
- Moves all colliders, rigidbodies, behaviours to a single 'components' list (left the old arrays empty for now but in the json to avoid blowing up the client
- Adds Type (Entity or Transform) to all state entries
_See private Slack for more details_
---


RG-nAmKcAz:
@abeizer  FYI

svolokh (approved):
Looks good on my end. I will probably need to update my usage of the TransformStatus on the automated analysis action manager branch, but that should be straight-forward based on the changes I see in this PR.

vontell (commented):
Still going through this review, but wanted to make sure there is a conversation on this between @abeizer, @addisonbgross, and @RG-nAmKcAz to ensure that the new data format is going to be supported in both the backend and the frontend. Has this format been recorded, and is there a plan for that yet?

RG-nAmKcAz:
> Still going through this review, but wanted to make sure there is a conversation on this between @abeizer, @addisonbgross, and @RG-nAmKcAz to ensure that the new data format is going to be supported in both the backend and the frontend. Has this format been recorded, and is there a plan for that yet?

So today, bot segment files are versioned... State data is NOT.

We can start versioning state data from this PR forward and I think we should.  That would have helped on both this PR and the perframestatistics pr that we just merged.  

Thoughts / Agreement on adding versioning to state data ?

vontell (approved):

vontell:
@RG-nAmKcAz yeah adding a version number would be great! With that in mind, however, @addisonbgross @abeizer I don't know if we have any switches yet or approaches on the backend and frontend for seeing and using the versions correctly, that's something we should think about.

RG-nAmKcAz:
@vontell The bot segments code in the SDK very much uses the versions to support old vs new files and provide user feedback.

Since the backend no longer loads the state recording files, that will be a UI only processing for now.

vontell:
@abeizer can you confirm this? Which files is the backend using for doing validations?

RG-nAmKcAz:
@vontell the state files are used on the backend.. the bot segments are solely for use by the SDK

In the future when/if we have a bot segment builder that answer will change.

RG-nAmKcAz:
Sample state file with entities and apiVersions on state

[1.json](https://github.com/user-attachments/files/16168912/1.json)

addisonbgross:
I love the Entity / Transform delineation 👍🏻

abeizer (approved):

PR #210, opened 2024-06-27
[REG-1818] Record CPU and memory utilization per frame

svolokh:
This replaces the `cpuTimeSincePreviousTick`, `memory`, and `gcMemory` performance stats with 
`cpuTimesPerFrame`, `memoryPerFrame`, `gcMemoryPerFrame`, which give the CPU and memory utilization for every frame in a list since the last tick. Note that these fields will typically be an empty list on the first tick.

Examples:
[1.json](https://github.com/user-attachments/files/16021604/1.json)
[8.json](https://github.com/user-attachments/files/16021605/8.json)

I've verified that the JSON output is correct for BossRoom for both in-editor and standalone build.

---

Find the pull request instructions [here](https://www.notion.so/regressiongg/Pull-Request-Process-0d57a7eb582a446983edfacafa406f1e?pvs=4)

Every reviewer and the owner of the PR should consider these points in their request (feel free to copy this checklist so you can fill it out yourself in the overall PR comment)

- [ ] The code is extensible and backward compatible
- [ ] New public interfaces are extensible and open to backward compatibility in the future
- [ ] If preparing to remove a field in the future (i.e. this PR removes an argument), the argument stays but is no longer functional, and attaches a deprecation warning. A linear task is also created to track this deletion task.
- [ ] Non-critical or potentially modifiable arguments are optional
- [ ] Breaking changes and the approach to handling them have been verified with the team (in the Linear task, design doc, or PR itself)
- [ ] The code is easy to read
- [ ] Unit tests are added for expected and edge cases
- [ ] Integration tests are added for expected and edge cases
- [ ] Functions and classes are documented
- [ ] Migrations for both up and down operations are completed
- [ ] A documentation PR is created and being reviewed for anything in this PR that requires knowledge to use
- [ ] Implications on other dependent code (i.e. sample games and sample bots) is considered, mentioned, and properly handled
- [ ] Style changes and other non-blocking changes are marked as non-blocking from reviewers

PR #215, opened 2024-07-02
[REG-1818] Record profiler statistics per frame

svolokh:
This PR introduces a `perFrameStatistics` field to the performance field, which collects profiler and engine stats for every frame of the tick. This is a revised version of an earlier PR (see [here](https://github.com/Regression-Games/RGUnityBots/pull/210) for background).

This implements the format that was suggested in the prior PR. The data size is about 350 to 380 bytes per frame. Note that the first tick will have `null` values for the profiler data.

I've tested this code both in the in-editor and standalone builds of Boss Room.

Examples:
[1.json](https://github.com/user-attachments/files/16073618/1.json)
[7.json](https://github.com/user-attachments/files/16073626/7.json)

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/Models/PerformanceMetricData.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/Models/PerformanceMetricData.cs
@@ -13,49 +13,26 @@ namespace RegressionGames.StateRecorder.Models
         public double previousTickTime;
         public int framesSincePreviousTick;
         public int fps;
-        public List<long> cpuTimesPerFrame;
-        public List<long> memoryPerFrame;
-        public List<long> gcMemoryPerFrame;
-        public EngineStatsData engineStats;
+        public List<PerFrameStatisticsData> perFrameStatistics;
 
         public void WriteToStringBuilder(StringBuilder stringBuilder)
         {
-            stringBuilder.Append("{\"previousTickTime\":");
+            stringBuilder.Append("{\n\"previousTickTime\":");
             DoubleJsonConverter.WriteToStringBuilder(stringBuilder, previousTickTime);
-            stringBuilder.Append(",\"framesSincePreviousTick\":");
+            stringBuilder.Append(",\n\"framesSincePreviousTick\":");
             IntJsonConverter.WriteToStringBuilder(stringBuilder, framesSincePreviousTick);
-            stringBuilder.Append(",\"fps\":");
+            stringBuilder.Append(",\n\"fps\":");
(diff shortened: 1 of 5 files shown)


vontell (approved):

addisonbgross (approved):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/Models/PerformanceMetricData.cs
```diff
@@ -12,28 +13,26 @@ public class PerformanceMetricData
...
-            LongJsonConverter.WriteToStringBuilderNullable(stringBuilder, cpuTimeSincePreviousTick);
-            stringBuilder.Append(",\"memory\":");
-            LongJsonConverter.WriteToStringBuilderNullable(stringBuilder, memory);
-            stringBuilder.Append(",\"gcMemory\":");
-            LongJsonConverter.WriteToStringBuilderNullable(stringBuilder, gcMemory);
-            stringBuilder.Append(",\"engineStats\":");
-            engineStats.WriteToStringBuilder(stringBuilder);
-            stringBuilder.Append("}");
+            stringBuilder.Append(",\n\"perFrameStatistics\":[\n");
+            for (int i = 0, n = perFrameStatistics.Count; i < n; ++i)
```
RG-nAmKcAz, line 27: Functionally this is fine, but it is inconsistent stylistically with all the other code

RG-nAmKcAz (approved):
1 pseudo blocking code style change request

What I'm most interested in though is why our 'frameTime' is soo different from the editor engine's reported 'frameTime'.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/Models/PerformanceMetricData.cs
svolokh: I've updated the code to match the style of the others,  let me know if there's anything I missed!

svolokh:
@RG-nAmKcAz Regarding the frame time, currently I just define this as the unscaled time since the last frame. The actual UnityStats.frameTime likely would exclude any idle time since the last frame, however I am unable to find any metric in the profiler API that corresponds to this (not even the CPU time metric that we are currently recording). I figure since the  `UnityStats.frameTime` is only available in the editor build it would be better to at least have something we can use in both standalone and in-editor builds.

RG-nAmKcAz:
@svolokh I agree with us keeping our frameTime calculation as it is 'true' time between frames.  I was more musing as to how unity can be so wildly optimistic with the time it reports.

RG-nAmKcAz (approved):

PR #216, opened 2024-07-03
[REG-1798] Initial automated action analysis

svolokh:
This PR adds the automated action analysis, which enables using the monkey bot introduced in the [action manager skeleton PR](https://github.com/Regression-Games/RGUnityBots/pull/208) with any game. This adds support for all commonly used keyboard and mouse input handling patterns (both Legacy Input Manager and Input System) that were planned for the first milestone.

Loom demo: https://www.loom.com/share/018bb5c25f064f6fa9756a481aff220d

Documentation for the usage of the monkey bot based on this work is also pending review here: https://github.com/Regression-Games/RegressionDocs/pull/78

Key components of this PR:
1. **RGActionAnalysis.cs**: This is the core implementation of the automated analysis. It combs through the game code and scene files to identify the inputs that are being listened for by objects, then saves the analysis results to disk. The [Roslyn](https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/) SDK, which is built into Unity, is used for code analysis. Beyond just matching the input APIs it also conducts data flow analysis to handle cases of local variables, fields, etc. See the design doc for the full list of patterns.
2. **RGActionInput.cs**: This is an improvement to the RGActionManager API: instead of immediately performing an action, each action now returns the device inputs needed to perform it. The bot can then reason about the inputs in order to finally decide on what set of inputs to perform.
3. **RGActionParamFunc.cs**: This is a serializable function type that is used for serializing/deserializing the action analysis results. It can either represent a constant value, or a sequence of field/property accesses (e.g. `gameSettings.Bindings.fireKey`). 
4. **RGUnityBots/Assets/Scripts/ActionManagerTests/*.cs**: I've created a set of MonoBehaviours that cover the different input handling patterns supported by the analysis. This is not an automated test yet, but eventually it could be.

Limitations (besides those listed in the design doc):
1. Before the user can use the bots, they need to first go into "Configure Bot Actions" and click the "Analyze Actions" button. For Boss Room, the analysis can take up to 10 seconds to run, so I have held off on having the analysis run by itself.
2. Unity UI support is very limited currently (it only supports clicking buttons). The main focus of this PR has been on gameplay keyboard/mouse actions. More work will be done on this in the next milestone.
3. The new Input System support is currently restricted to Keyboard.current and Mouse.current. The other input-handling mechanisms will be covered in the next milestone.
4. If an input is handled outside of MonoBehaviour (e.g. in a helper method), the analysis will emit a warning. This isn't supported yet.
5. For mouse click actions, overlapping buttons/objects are not yet given any special handling (currently it is just center coordinate for everything)


src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/RGGameAction.cs
```diff
@@ -27,15 +21,24 @@ public abstract class RGGameAction
...
         {
             Paths = new List<string[]> { path };
             ObjectType = objectType;
-            ActionGroup = actionGroup;
+        }
+
+        public RGGameAction(RGSerializedAction serializedAction)
+        {
+            Paths = new List<string[]>(serializedAction.paths.Select(path => path.Split("/")));
+            ObjectType = Type.GetType(serializedAction.objectTypeName);
```
RG-nAmKcAz, line 37: This needs the second boolean argument to avoid throwing an exception on not found.

```suggestion
            ObjectType = Type.GetType(serializedAction.objectTypeName, false);
```

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/RGGameAction.cs
```diff
@@ -69,6 +72,53 @@ public virtual bool IsEquivalentTo(RGGameAction other)
...
+        public string objectTypeName;
+
+        // Used by actions that have a serializable function parameter (e.g. key function, axis name function, etc.)
+        public ActionParamFuncType actionFuncType;
+        public string actionFuncData;
+        
+        // Field where arbitrary string data can be stored
+        public string actionStringData;
+
+        public RGGameAction Deserialize()
```
RG-nAmKcAz, line 112: This works, but is inconsistent with how we are doing things in the code in other places.  Everywhere else, we are using newtonsoft json converters to serialize and deserialize objects.  Using newtonsoft everywhere has provided consistency and also allowed us to be able to to versioned deserialization more easily when needed.

src/gg.regression.unity.bots/Runtime/Scripts/GenericBots/RGMonkeyBot.cs
```diff
@@ -53,55 +61,133 @@ void Update()
...
+                if (entry.Inputs.Count > 0)
+                {
+                    _actionsBuf.Add(entry);
+                }
+            }
+
+            // Randomly perform actions from the list 
+            // This repeatedly selects actions whose inputs do not overlap
+            // with the inputs that have already been performed.
+            for (;;)
```
RG-nAmKcAz, line 167: while(true) loops scare me

let's make this something like

`do {....} while(_remainingActionsBuf.Count >0)`

src/RGUnityBots/Assets/Scripts/ActionManagerTests/InputSysKeyListeningObject.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    public class InputSysKeyListeningObject : MonoBehaviour
+    {
+        private Key fireKey = Key.Backspace;
+    
+        void Update()
+        {
+            if (Keyboard.current[Key.F2].isPressed)
+            {
+                Debug.Log("Keyboard.current[Key.A].isPressed");
```
RG-nAmKcAz, line 16: We should use `RGDebug` instead of `Debug` in all our SDK code

RG-nAmKcAz (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/RGActionProvider.cs
```diff
@@ -1,23 +1,88 @@
...
+            }
+            #else
+            {
+                TextAsset jsonFile = Resources.Load<TextAsset>(ANALYSIS_RESULT_NAME);
+                jsonText = jsonFile?.text;
+            }
+            #endif
+            if (jsonText != null)
+            {
+                return JsonUtility.FromJson<RGActionAnalysisResult>(jsonText);
```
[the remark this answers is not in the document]
svolokh: I consider the action analysis output to be ephemeral data, since it needs to be re-generated whenever the game code changes. So, I don't really see the need to have backwards compatibility for the saved results format. If it substantially changes then we just re-run the analysis.

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/RGGameAction.cs
svolokh: My thought on this is that I would actually prefer it to throw an exception, since an invalid type name signifies that the analysis results are no longer applicable to the current code base and the deserialization process should be aborted.

Here I have handling for any exceptions that come out of the deserialization, notifying that the analysis needs to be re-run if an exception is thrown.
https://github.com/Regression-Games/RGUnityBots/blob/dc42e2d139b369b4fffc86f50ff2875c00278aca/src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/RGActionProvider.cs#L48-L52

RG-nAmKcAz: if we're expecting it to throw the exception, then we can remove the check for null on the line below

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/RGActionProvider.cs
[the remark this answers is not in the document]
RG-nAmKcAz: It would be a nice user experience when they update the SDK to an incompatible version, to be able to tell them clearly that they need to regenerate their action data.

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/RGGameAction.cs
svolokh: I've updated this to use Newtonsoft JSON converters, as well as do serialization via string builders in a similar manner as the other objects.

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/RGGameAction.cs
svolokh: I've removed the null check and this will just throw an exception if the type isn't found now.

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/RGActionProvider.cs
```diff
@@ -1,23 +1,88 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
+using UnityEngine;
 
 namespace RegressionGames.ActionManager
 {
+    [Serializable]
+    public class RGActionAnalysisResult
```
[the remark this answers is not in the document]
svolokh: I've split out the class into its own file.

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/RGActionProvider.cs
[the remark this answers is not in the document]
svolokh: I've added an API version field to RGActionAnalysisResult. If the current API version does not match the one that was parsed, then an exception is raised and the user will be requested to regenerate the action data.

src/RGUnityBots/Assets/Scripts/ActionManagerTests/InputSysKeyListeningObject.cs
svolokh: Normally I use RGDebug, but this particular line of code is for a test scene in the RGUnityBots project. Eventually I want to use LogAssert.Expect to write automated tests that the action manager is covering all these input types, so I want to avoid the extra data that is introduced by RGDebug.

src/gg.regression.unity.bots/Runtime/Scripts/GenericBots/RGMonkeyBot.cs
RG-nAmKcAz: @svolokh other changes look good as far as I can tell

just this last comment remaining from my end

svolokh: Updated this one to be do-while as suggested.

RG-nAmKcAz (approved):

vontell (approved):

PR #221, opened 2024-07-11
[REG-1848] Gracefully handle IO errors during legacy input instrumentation

svolokh:
For various reasons, the initial attempt at instrumenting the legacy input API usage in the assemblies could fail (e.g. file in use by another job). 

This PR introduces graceful handling of these errors. If the initial instrumentation fails, a couple more attempts are scheduled with exponential backoff. Ultimately if all these attempts fail, the user is warned that the SDK's legacy input support may not work and the user should consider rebuilding the game code.

---

### DIFF
--- a/src/gg.regression.unity.bots/Editor/Scripts/RGLegacyInputUtility/RGLegacyInputInstrumentation.cs
+++ b/src/gg.regression.unity.bots/Editor/Scripts/RGLegacyInputUtility/RGLegacyInputInstrumentation.cs
@@ -302,6 +302,7 @@ namespace RegressionGames.Editor.RGLegacyInputUtility
             }
             if (done)
             {
+                _numInstrumentationAttempts = 0;
                 EditorApplication.update -= ScheduledInstrumentationLoop;
             }
         }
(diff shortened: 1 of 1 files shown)


src/gg.regression.unity.bots/Editor/Scripts/RGLegacyInputUtility/RGLegacyInputInstrumentation.cs
```diff
@@ -260,17 +259,77 @@ public static Assembly FindRGAssembly()
...
+        {
+            try
+            {
+                Assembly[] assemblies = CompilationPipeline.GetAssemblies(AssembliesType.Player);
+                Assembly rgAssembly = FindRGAssembly();
+                foreach (Assembly assembly in assemblies)
+                {
+                    InstrumentAssemblyIfNeeded(assembly.outputPath, rgAssembly);
+                }
+                return true;
```
RG-nAmKcAz, line 319: do we need a 

`_numInstrumentationAttempts = 0;` 

code line here to reset this ?

RG-nAmKcAz, line 319: do we need a `_numInstrumentationAttempts = 0;` code line here before the return to reset this ?

RG-nAmKcAz (approved):
One question about resetting the counters on a working instrumentation, but overall the concept seems good.

src/gg.regression.unity.bots/Editor/Scripts/RGLegacyInputUtility/RGLegacyInputInstrumentation.cs
svolokh: Technically it isn't needed since the loop can only ever happen once per static initialization, but for good measure I've added in a reset after the loop terminates:

https://github.com/Regression-Games/RGUnityBots/blob/3ab36906770a6cfa866c7bc823d03dee7fba3f81/src/gg.regression.unity.bots/Editor/Scripts/RGLegacyInputUtility/RGLegacyInputInstrumentation.cs#L303-L307

vontell (approved):
Great approach! Glad you implemented a backoff approach with limited retries.

PR #223, opened 2024-07-13
[WIP] Code coverage measurement prototype

svolokh:
This draft PR is a prototype that introduces code coverage measurement into the RG SDK. It would be good to get feedback on the data format and whether this is something we'd like to have in the SDK. Some possible applications of this work: 1) Displaying code coverage statistics in the client, 2) Finding out during what ticks a particular line of code was executed, 3) Validations that verify a certain region of code was covered, 4) Visualizing code execution along with replays

This introduces a `codeCoverage` object to each tick state recording that includes all the code points that have been covered since the last tick, and also includes some metadata about the total number of possible code points (for calculating coverage percentage).

Here is an example Boss Room data.zip with the new format:
[data.zip](https://github.com/user-attachments/files/16200197/data.zip)

Here's an example plot I made with the code coverage data (to illustrate how you would calculate % coverage over time using this):
https://gist.github.com/svolokh/20602bbdf6579f60c855a3562d0ba60f

This works by automatically inserting `RGCodeCoverage.Visit` calls at every point in the game assemblies that have an associated line number in the source code. When a key frame is encountered, all the code points that were covered up to that point are saved (and then reset, so that we are measuring the code coverage between ticks).

image

This is actually a pretty stable implementation, the main things that are missing currently are proper handling of multi-threaded games and the ability to disable the code coverage measurement (although, for many games it may not actually have a noticeable impact).


abeizer:
Code coverage is a neat idea, especially being able to get this kind of feedback outside the context of the Unity Test Runner (which we've identified as sometimes useful but generally not). Going to spitball some thoughts here to kick off discussion...

> Some possible applications of this work: 1) Displaying code coverage statistics in the client, 2) Finding out during what ticks a particular line of code was executed, 3) Validations that verify a certain region of code was covered, 4) Visualizing code execution along with replays

I really like some of these ideas, like number 2. It has potential to help diagnose bugs, especially in games that behave differently when they pause at breakpoints.

However, I think all of these use cases would be much more useful to devs than QA. From personal experience, code coverage reports are only really useful if they show the code that was hit/missed (not just the line numbers). I wouldn't expect a QA tester to dig into the code, so that immediately makes these unity-side features to me rather than something for the client to display. 

Now it's possible a QA team could use code coverage as one of several metrics to determine whether their QA test coverage is acceptable. In that case, we'd probably be looking at cumulative coverage over multiple scenarios/suites, but I see a couple problems here:

1.  it's not enough to tell QA "hey you're missing some stuff" without a summary of what that "stuff" is. We'd need to consider how we'd collect relevant source code to show which lines were missed from the client, and whether that would even be acceptable (what would studios think of us saving their source during recordings?)
2. Right now, having a recording doesn't necessarily mean that everything in that recording is being validated. As QA, I could make a recording that covers a range of functionality - for example, movement, combat, and player trading. But I could choose to only write Validation Scenarios for movement for that specific recording. What does this mean in terms of the line coverage? The recording may hit logic for combat and player trading, but we aren't validating those behaviours so does that really count as "coverage"? Maybe I'm overthinking and conflating line coverage with test coverage, but I figure it's worth asking what this metric would mean to QA testers.

TL;DR: For any client-side code coverage features, we'd have to throw around more ideas of how to bridge the gap of making dev-centric metrics useful to non-devs.

PR #226, opened 2024-07-16
[REG-1848] Address instrumentation conflicts and improve error handling

svolokh:
This is a second pass at addressing issues that can occur with the input instrumentation:
1. Check the `EditorApplication.isCompiling` and `EditorApplication.isUpdating` flags. If either is true, defer the instrumentation until later.
2. Remove the inner exception handler in InstrumentAssemblyIfNeeded (we already have exception handlers in the callers), and broaden the exceptions handled to all Exceptions (this will handle any errors that could come out of Mono.Cecil's own APIs).

I've tested this by re-opening Boss Room and some other Unity projects from scratch (deleting the Library folder). The whole instrumentation process works in my testing.

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/Actions/MouseHoverObjectAction.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/Actions/MouseHoverObjectAction.cs
@@ -1,8 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Text;
-using System.Collections.Generic;
-using System.Text;
 using Newtonsoft.Json.Linq;
 using UnityEngine;
 using UnityEngine.UI;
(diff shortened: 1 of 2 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/Actions/MouseHoverObjectAction.cs
```diff
@@ -1,8 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Text;
-using System.Collections.Generic;
```
svolokh, line 4: Fixing a warning that was introduced from an earlier PR

RG-nAmKcAz:
@svolokh This looks like it will prevent us from starting while Unity is actively compiling.

Do we have a way to either interrupt us, or block Unity from compiling until we finish if a user makes & saves a script change or attaches the debugger while we're mid compilation ?

svolokh:
@RG-nAmKcAz The logic for deferring while Unity is compiling is only applied to the initial instrumentation pass (when the RG SDK is first loaded and it iterates through the assemblies to check if any need instrumentation). This is generally where the source of IO errors comes from.

If a **new** compilation is triggered while our SDK is active, then this will instead trigger the `OnAssemblyCompiled` callback and our instrumentation will still be applied during Unity's compilation pipeline. So this will still work correctly if the user makes changes.

As far as locking the Unity editor from compiling, the best thing I can find is [EditorApplication.LockReloadAssemblies()](https://docs.unity3d.com/ScriptReference/EditorApplication.LockReloadAssemblies.html), but it's not clear to me whether this will be an improvement.

RG-nAmKcAz (approved):
I think we may still need 1 more pass to handle the discussion in this PR about blocking unity from recompiling changes while we are in process, but this is another good step.

PR #230, opened 2024-07-17
[REG-1817] [REG-1842] [REG-1847] Action manager automated test suite + precision improvements

svolokh:
This PR adds an automated Unity Test Runner test suite for the different action types supported by the action manager. You can see it passing in GameCI:
image

It also introduces the following improvements to improve the precision of the identified actions:
1. **Handling of overlapping game objects and buttons:** when checking the validity of a button press or hover over game object, a raycast will now be performed to test that the object is actually reachable (i.e. not being overlapped by another object). This is implemented for both Unity UI buttons and game objects (OnMouseOver, OnMouseDown, etc.).

2. **Handling of mouse position involved in a raycast** (e.g. `Camera.ScreenPointToRay(Input.mousePosition)`). If the analysis detects that a mouse position is being used as an argument to a call to `Physics.Raycast` or `Physics2D.Raycast`, it will generate a more precise action that only targets the specified type of collider (3D or 2D), and also further filters by layer mask if provided.  Note that raycasts that use ContactFilter is not supported (this will still generate a more imprecise mouse position action).

---


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/KeyboardInputActionObserver.cs
```diff
@@ -79,7 +79,7 @@ public class KeyboardInputActionObserver : MonoBehaviour
             { "p", Key.P },
             { "leftBracket", Key.LeftBracket },
             { "rightBracket", Key.RightBracket },
-            { "backSlash", Key.Backslash },
+            { "backslash", Key.Backslash },
```
svolokh, line 82: This is a fix for an error when simulating backslash key events (it is actually named "backslash" in the control path, rather than "backSlash").

RG-nAmKcAz:
@svolokh I need some help reviewing this, specifically pointers to some key code areas.

I'm looking for the code that says.. try to click on button X, but raycast hit button Y.  However.. part of button X is not covered by button Y bounds, so if I slide over the click location a bit, i can still validly click button X.

svolokh:
@RG-nAmKcAz 
> I'm looking for the code that says.. try to click on button X, but raycast hit button Y. However.. part of button X is not covered by button Y bounds, so if I slide over the click location a bit, i can still validly click button X.

To handle the case of partial overlap, there are three raycasts performed: center, bottom left, and top right. 
The points from which raycast is done can be found in GetRaycastPoints() inner method.

Here is the one for UI:
https://github.com/Regression-Games/RGUnityBots/blob/f44c2206d198ce55bb5cd515c0de1c84c0c6c947/src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/RGActionManagerUtils.cs#L403-L408

This is the one for colliders (same heuristic):
https://github.com/Regression-Games/RGUnityBots/blob/f44c2206d198ce55bb5cd515c0de1c84c0c6c947/src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/RGActionManagerUtils.cs#L454-L459

RG-nAmKcAz:
@svolokh Let's definitely add some line comments in the code then explaining that and making it more obvious to future devs what is happening there.

RG-nAmKcAz (approved):

src/RGUnityBots/Assets/Tests/Runtime/RGActionManagerTests.cs
```diff
@@ -82,5 +91,401 @@ public void TestValueRanges()
...
+            // Test Input System Keyboard
+            RGActionManager.StartSession(eventSys);
+            GameObject inputSysKeyListener = FindGameObject("InputSysKeyListener");
+            inputSysKeyListener.SetActive(true);
+            try
+            {
+                FindAndPerformAction("Keyboard.current.anyKey", true);
+                yield return null;
+                yield return null;
+                LogAssert.Expect(LogType.Log, "Keyboard.current.anyKey.wasPressedThisFrame");
```
vontell, line 141: Nice! Glad we can use these as examples as well for monobehaviours that would need to interact with this API

src/RGUnityBots/Assets/Tests/Runtime/RGActionManagerTests.cs
```diff
@@ -82,5 +91,401 @@ public void TestValueRanges()
...
+
+                FindAndPerformAction("[key]", true);
+                yield return null;
+                yield return null;
+                LogAssert.Expect(LogType.Log, "keyboard[key].isPressed");
```
vontell, line 146: Any specific reason why we need to wait two frames? Just to play it safe?

src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs
```diff
@@ -562,20 +623,71 @@ public override void VisitInvocationExpression(InvocationExpressionSyntax node)
...
+                        ? MousePositionType.COLLIDER_3D
+                        : MousePositionType.COLLIDER_2D;
+                    
+                    string methodName = methodSymbol.Name;
+                    // Raycast
+                    switch (methodName)
+                    {
+                        case "Raycast":
+                        case "RaycastAll":
+                        case "RaycastNonAlloc":
```
vontell, line 644: Additional high level comments on what is happening in these code blocks will be useful for future reviewers.

src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs
```diff
@@ -894,6 +1018,19 @@ private void RunCodeAnalysis()
                 }
             }
         }
+
+        private ISet<string> _buttonClickListeners;
+
+        private void AnalyzeGameObject(GameObject gameObject)
```
vontell, line 1024: Comments also would be helpful here - this method says "AnalyzeGameObjects" but this method code makes me think "SetButtonClickListeners"?

vontell (changes requested):
Mostly questions asking for some clarifications

src/RGUnityBots/Assets/Tests/Runtime/RGActionManagerTests.cs
svolokh: For the most part I expect that the APIs should trigger the event within one frame, but I've noticed some test flakiness if I only wait for one frame. Waiting for two frames guarantees that the event is triggered and the expected log is generated.

src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs
svolokh: Added more comments to the RGActionAnalysis class overall (as well as these code blocks as well).

src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs
svolokh: In the future there will be more game object analysis done here besides just button click listeners (for example, my current work on InputAction support examines all the component fields to see if any have embedded InputActions). I've added more comments here on what this analysis is doing.

vontell (approved):

PR #233, opened 2024-07-24
[REG-1829] [REG-1857] Action manager support for InputAction and Action Assets + Interprocedural analysis

svolokh:
This PR updates the action manager to support games that use InputAction and [Action Assets](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.8/manual/ActionsEditor.html) from the new Input System. This enables the Monkey Bot to identify the relevant inputs for more types of games that rely on the new Input System. 

Secondly, this PR also expands the range of games supported by updating `RGActionAnalysis` to support identifying actions even if the input-handling code resides outside of a MonoBehaviour (for example, in a helper method). This is done by iteratively propagating identified actions through the method calls until the analysis results no longer change (this also means the code analysis may now require multiple passes).

Automated tests have been added for both of the above capabilities to the RGUnityBots test project.

Key changes in this PR:
1. **RGActionAnalysis.cs**: Updated to scan for InputAction and InputActionAsset in the project, as well as support interprocedural analysis of input-handling code
2. **InputActionAction.cs**: This is where the bulk of the logic is for determining whether an InputAction is applicable to a given game state, as well as what keyboard/mouse inputs need to be simulated to trigger its bindings. This supports all keyboard/mouse bindings as well as "composite" bindings (see Input System [docs](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.8/manual/ActionBindings.html)).

---


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/KeyboardEventSender.cs
```diff
@@ -24,83 +25,101 @@ public static void Reset()
          */
         public static void SendKeysInOneEvent(int replaySegment, List<(Key, KeyState)> keyStates)
```
svolokh, line 26: @RG-nAmKcAz I've made two changes to the SendKeysInOneEvent method (since I'm now using it to simulate key events from the action manager):
1. Also have it send key events through the legacy input manager (added at the top)
2. I noticed that it was not releasing the NativeArray allocated from `DeltaStateEvent.From`, so I added some code to Dispose it after the event is sent.

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/RGActionManager.cs
```diff
@@ -306,7 +312,16 @@ public static void SimulateKeyState(Key key, bool isPressed)
         {
             if (key != Key.None)
             {
-                KeyboardEventSender.SendKeyEvent(0, key, isPressed ? KeyState.Down : KeyState.Up);
+                // The new Input System does not support generating multiple keyboard events
+                // within one frame for individual keys. So, we generate events for multiple keys
+                // at once whenever sending keyboard events.
+                _keyStates[key] = isPressed ? KeyState.Down : KeyState.Up;
```
RG-nAmKcAz, line 318: I'm concerned that we shouldn't this across frames due to real keyboard input potentially corrupting our state.

can we add some code to Clear _keyStates on the start of each new frame  ?

we could probably then just changer the SendKeysInOneEvent method to only accept a Dictionary not a list.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/KeyboardEventSender.cs
```diff
@@ -24,83 +25,101 @@ public static void Reset()
...
-
-            if (deltaStateEvent.HasValue)
+            finally
             {
-                RGDebug.LogInfo($"({replaySegment}) Sending Multiple Key Event: [" + string.Join(", ", logList.Select(a => a.Item1 + ":" + a.Item2).ToArray()) + "]");
-                InputSystem.QueueEvent(deltaStateEvent.Value);
-                logList.Clear();
+                if (deltaStateEventArray.HasValue)
+                {
+                    deltaStateEventArray.Value.Dispose();
```
RG-nAmKcAz, line 121: could this possibly dispose the event before the input system processes the queued event ?

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/KeyboardEventSender.cs
```diff
@@ -24,83 +25,101 @@ public static void Reset()
...
                             logList.Clear();
                             InputSystem.QueueEvent(deltaStateEvent.Value);
                             deltaStateEvent = null;
-                            keys.Clear();
+                            deltaStateEventArray.Value.Dispose();
+                            deltaStateEventArray = null;
+                        }
+                        else
+                        {
+                            if (keys.Contains(key))
```
RG-nAmKcAz, line 62: This code I wrote still allows 2 events in the same update for the same keys.  Based on what you reported with the input system and how it handles events, would it be better to make this whole method only accept a dictionary of key states ?

If we wanted to still support a method with the current signature (Which would be nice so I don't have to move all this buffer logic into application code for starsreach), then if feels like we would need to manage our own key buffer across frames and hold off on building/sending the follow up event until the next frame.

RG-nAmKcAz (changes requested):
I requested changes with some feedback on the method signatures and design of the multi key sender, but mostly for discussion so we can land on a design pattern that makes the application code side easy, but also doesn't leave us open to sending 2 events for the same key in one frame.

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/RGActionManager.cs
svolokh: I like the idea of changing SendKeysInOneEvent to accept a dictionary. 

I thought about clearing keyStates at the start of each frame, but the problem is that some games could configure the Input System to process events at a different rate (i.e. not every frame) (see the [Update Mode](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.8/manual/Settings.html#update-mode) setting). If the keyStates buffer is cleared too early, then any new key events sent would cause the previous key state changes to get ignored.

Every StateEvent carries the entire state of the keyboard with it, so I don't actually think it's a problem to have multiple events queued for the same key (it will just land on whatever the last value was).

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/KeyboardEventSender.cs
svolokh: I like the idea of converting it to accept a dictionary. 

If we want to have an API that allows for sending multiple individual key events within one frame, then I think we do need to do some buffering ourselves (similar to what I'm doing with keyStates in the action manager), since we can't send delta events for individual keys directly to the Input System due to the bitfield issue.

Though I don't think this necessarily means we need to hold off until the next frame - it just means that if another call to our API is made, then we should queue another keyboard event to the Input System that includes any buffered key changes + the new key change.

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/RGActionManager.cs
RG-nAmKcAz: So the problem with changing it to a dictionary, and why I think maybe we keep a running buffer of events still to send in KeyboardEventSender....

Is that in my behaviour bot for stars reach, i send multiple key event requests per update call, and in fact send both key up and key down for the same key in that call.

Example, I determine that I want to crouch starting in this update, but i want to keep moving forward.

To avoid 'rolling'. I have to un-press W, hit CTRL, then press W again.  Since this logic is all in a single update from that application space code, having to manage the event logic on updates myself there would be less than ideal.


Thus, my goal is we end up with something in the keyboard event sender that can handle all the sequential key inputs for the current frame, but be smart enough to do it in the minimum number of update calls while still ordered... so example



Events sent by application

Update 1
A - UP
W - UP
CTRL - DOWN
W - DOWN
D - DOWN
R - DOWN

LateUpdate 1
T - DOWN

Update 2
G - DOWN
R - UP


KeyboardEventSender buffering logic

LateUpdate 1
Send Event with (A-UP, W-UP, CTRL-DOWN, D-DOWN, R-DOWN) - May or may not also include T-DOWN depending on script execution ordering...

LateUpdate 2
Send Event with (W-DOWN, G-DOWN, R-UP) - May or may not also include T-DOWN depending on script execution ordering and whether it was already sent in LateUpdate 1

RG-nAmKcAz: but.. typing that.. maybe that also means that update loop specific code does truly belong in the application layer as it would be game specific... thus looping this conversation back to me agreeing that the dictionary argument seems like the best option

svolokh: I can see how the list would be useful in this scenario, but I think the current implementation of SendKeysInOneEvent wouldn't do what is desired? If it encounters a duplicate key entry (or shift) and then starts a new event, that new event would just result in any key state changes that happened up to that point getting discarded, since the call to `DeltaStateEvent.From` will just make a fresh copy of the keyboard state.

RG-nAmKcAz: see 
https://github.com/Regression-Games/RGUnityBots/blob/main/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Samples/SimpleRandomMovementBehaviour.cs

for context on how it was being used...

it would be awesome if all the logic in there about trying to group events `List<(Key, KeyState)> keyStates = new();` was un-necessary and you could just naively send key up/down requests one by one and they were handled as a single event appropriately by the sender as efficiently as possible

... maybe that's an unreachable pancea .. but i can dream

svolokh: I thought a bit about this and realized I can use the `InputSystem.onAfterUpdate` hook to clear the key state buffer after it's been processed by the input system (commit pushed). Perhaps similar logic could be moved into KeyboardEventSender as well so we can indeed have the desired single key up/down API (buffer key events until the input system completes an update).

svolokh:
@RG-nAmKcAz I've committed an update to KeyboardEventSender based on our discussion above (effectively moving my keyState buffering logic out of RGActionManager and into KeyboardEventSender), this should enable the desired functionality of sending key up/down events individually via KeyboardEventSender.SendKeyEvent. I've kept in the multi-key variant anyways in case something wants to send multiple key changes at once more efficiently.

RG-nAmKcAz:
@svolokh Acknowledged .. I'll need to test / update the code in the `SimpleRandomMovementBehaviour` in our SDK before confirming this PR

src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs
```diff
@@ -872,47 +863,30 @@ public override void VisitMemberAccessExpression(MemberAccessExpressionSyntax no
...
-
-                    string[] path = GetActionPathFromSyntaxNode(objectType, node);
+                    string[] path = GetActionPathFromSyntaxNode(node);
                     switch (propSym.Name)
                     {
                         case "position":
                         case "delta":
                         {
-                            AddAction(new MousePositionAction(path, objectType), node);
+                            AddAction(new MousePositionAction(path, null), node);
```
vontell, line 889: I notice the second parameter is almost always `null`, are there cases where it isn't?

src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs
```diff
@@ -959,43 +933,116 @@ public override void VisitMethodDeclaration(MethodDeclarationSyntax node)
...
+                    _changed = true;
+                }
+            }
+            else
+            {
+                // If an action was identified outside a MonoBehaviour (e.g. in a helper method), then
+                // it is added as an "unbound" action (i.e. it is unknown what component is listening for it).
+                // The action is associated with its containing method. Eventually, if a method invocation is found
+                // from a MonoBehaviour to the method that contained the actions, then copies of all the unbound actions
+                // will be associated with the containing MonoBehaviour.
```
vontell, line 1024: Appreciate the comments! It helps a lot for reviewing and later changes.

vontell: (non-blocking) What happens if multiple MonoBehaviours use those actions? Will it allow mapping to multiple?

src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs
```diff
@@ -1095,28 +1207,36 @@ private void RunResourceAnalysis()
...
-                        var scene = SceneManager.GetSceneAt(i);
-                        foreach (GameObject gameObject in IterateGameObjects(scene))
+                        string scenePath = AssetDatabase.GUIDToAssetPath(sceneGuid);
+                        if (scenePath.StartsWith("Packages/"))
                         {
-                            AnalyzeGameObject(gameObject);
+                            continue;
                         }
+                        NotifyProgress($"Analyzing {Path.GetFileNameWithoutExtension(scenePath)}", progress);
+                        UnityEditor.SceneManagement.EditorSceneManager.OpenScene(scenePath);
```
vontell, line 1227: (non-blocking) Will this cause the scene to change in the editor?

src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs
```diff
@@ -1168,9 +1325,19 @@ public bool RunAnalysis()
...
-                RunCodeAnalysis();
+
+                {
+                    int passNum = 1;
+                    do
+                    {
+                        _changed = false;
+                        RunCodeAnalysis(passNum);
+                        ++passNum;
+                    } while (_changed);
```
vontell, line 1338: Should we also set a max number of passes? Or is it guaranteed to terminate?

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/KeyboardEventSender.cs
```diff
@@ -12,95 +13,102 @@ namespace RegressionGames.StateRecorder
...
+        /**
+         * Allows updating multiple keys in the same event. 
+         */
+        public static void SendKeysInOneEvent(int replaySegment, List<(Key, KeyState)> keyStates)
+        {
+            #if ENABLE_LEGACY_INPUT_MANAGER
+            SendKeysInOneEventLegacy(keyStates);
+            #endif
+            
+            RGDebug.LogInfo($"({replaySegment}) Sending Multiple Key Event: [" + string.Join(", ", keyStates.Select(a => a.Item1 + ":" + a.Item2).ToArray()) + "]");
```
vontell, line 97: (non-blocking) Does that mean legacy input systems sent inputs here and then again down below? Curious why this doesn't return here.

vontell (changes requested):
Just a few blocking comments - looks good otherwise, besides any remaining things from Zack's comments. Appreciate you adding some tests to this!

vontell:
@RG-nAmKcAz yeah that's something I will need to get on top of, will begin looking into that

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/KeyboardEventSender.cs
```diff
@@ -12,95 +13,102 @@ namespace RegressionGames.StateRecorder
...
+                var inputEvent = TextEvent.Create(Keyboard.current.deviceId, value, InputState.currentTime);
+                RGDebug.LogInfo($"({replaySegment}) Sending Text Event for char: '{value}'");
+                InputSystem.QueueEvent(ref inputEvent);
+            }
+        }
+
+        /**
+         * Allows updating multiple keys in the same event. 
+         */
+        public static void SendKeysInOneEvent(int replaySegment, List<(Key, KeyState)> keyStates)
```
RG-nAmKcAz, line 91: I thought this was going to accept a `Dictionary` instead of a list.

Right now, when I pass it a list of (A-UP, W-UP, W-DOWN, D-DOWN)... the A-UP and W-UP get thrown away and not sent.

It looks like we either need to make this take in a dictionary... or add code to detect when we already have the specific key in the dictionary during the for loop and send the existing event and start a new one for the rest of the keys

RG-nAmKcAz (changes requested):
Some questions still about list vs dictionary... or auto splitting the events from the list

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/KeyboardEventSender.cs
svolokh: @RG-nAmKcAz I think I'd prefer to convert this to a dictionary. If the application wants to do something over multiple frames, I think it would be better to have it handle that in its own logic. Otherwise we'll have to do something complicated with queuing up key state changes until after the next input system update and it could get messy accounting for the different ways the APIs could be used.

src/RGUnityBots/Assets/Tests/Runtime/RegressionGames.Tests.asmdef
```diff
@@ -7,5 +7,8 @@
         "RegressionGames",
         "Unity.InputSystem",
         "Unity.InputSystem.TestFramework"
+    ],
+    "precompiledReferences": [
+        "Newtonsoft.Json.dll"
```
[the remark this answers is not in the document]
svolokh: This is just making it possible to use the Newtonsoft APIs within the test code. The actual decision about which package/version to use is still happening in the package manager.

src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs
svolokh: The `AddAction` method resolves the object type later - I made this change because I was previously duplicating the same code over and over for resolving the object type, and also because we need to handle the possibility of the target object not being known yet if this code is from a helper method.

See the following code for where the resolution of ObjectType happens:
https://github.com/Regression-Games/RGUnityBots/blob/5c2dda94d8b2e73542f78ad09fa0849cea500b7e/src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs#L991-L1000

src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs
svolokh: Yes, if the same helper method is re-used in multiple places then copies of the action will be made for each of those locations.

src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs
svolokh: All of the propagation logic is guaranteed to terminate. Even if we somehow end up with a bug where the analysis never terminates, the user can just click the "Cancel" button on the progress bar and the analysis will abort (I've made sure that all parts of the analysis procedure are interruptible by the user whenever a call to `NotifyProgress` is made).

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/KeyboardEventSender.cs
svolokh: The code that follows afterwards is the code for sending the key events down the new input system - we still want to send events down both systems in case the game is using both the new and old input systems.

The actual implementation of `SendKeysInOneEventLegacy` just sends multiple individual events through `RGLegacyInputWrapper`. There isn't really any need to group the events for the legacy input support in the same way as we do for the input system

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/KeyboardEventSender.cs
svolokh: I've made the following changes:
1. I've converted SendKeysInOneEvent to take a dictionary: this method now serves to just create a single event for the given set of keys.
2. In order to handle the case you mentioned above, I've introduced a coroutine called `KeyboardEventSender.SendKeyEvents` and updated the sample exploration MonoBehaviour to use it instead. It will do the auto-splitting of events as done previously, except that it will actually wait until the next Input System update before queuing up a new event, since otherwise the Input System will just flush all the events right away.

Note that beyond some local testing I am not actually able to test the entire MonoBehaviour currently, so you may want to test it and feel free to push any other changes as needed to this branch.

vontell (approved):

vontell:
Approving on my end as my comments has been addressed, but please note Zack's remaining requests

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/KeyboardEventSender.cs
```diff
@@ -135,18 +148,46 @@ public static void SendKeyEvent(int replaySegment, Key key, KeyState upOrDown)
...
+                        SendKeysInOneEvent(replaySegment, keyStates);
+                        yield return WaitForInputSystemUpdate();
+                        keyStates.Clear();
+                    }
+                }
+                keyStates[key] = upOrDown;
+            }
+            if (keyStates.Count > 0)
+            {
+                SendKeysInOneEvent(replaySegment, keyStates);
```
RG-nAmKcAz, line 178: I think we might need a `yield return WaitForInputSystemUpdate()` after this line.

If i were to call this method 2x from my calling code, and neither call had duplicated keys in it, but I did have the same keys in call 2 that I had in call 1, we would send 2 events in a single update pass with conflicting keys.

RG-nAmKcAz (changes requested):

RG-nAmKcAz:
@svolokh one blocking comment on the new code...

that said, it seems to works better than the prior code in testing on PW

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/KeyboardEventSender.cs
svolokh: Fair point, I will make that change. I now wonder if I should just convert the entire SimpleRandomMovementBehaviour's Update() method to a coroutine, which waits until the SendKeyEvents coroutine finishes?

RG-nAmKcAz: that would cover our current case, but not all cases / future user code patterns

svolokh: I've added the suggestion, and I've also updated SimpleRandomMovementBehaviour to wait until the key event sending coroutine finishes before doing another update step, since this could cause a similar issue as the one you noted if multiple coroutines get scheduled in parallel.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Samples/SimpleRandomMovementBehaviour.cs
```diff
@@ -40,8 +40,20 @@ public class SimpleRandomMovementBehaviour : MonoBehaviour
 
         private float _lastPositionTime = float.MinValue;
 
+        private bool _sendingKeysInProgress = false;
+
+        private IEnumerator SendKeys(List<(Key, KeyState)> keyStates)
+        {
+            _sendingKeysInProgress = true;
```
RG-nAmKcAz, line 47: This concept seems good, but I don't like it being in user space code.  I feel like we should be able to contain this logic within the KeyboardEventSender.  I'll push a commit with my suggestions

RG-nAmKcAz:
Summary of the last commit I just pushed

1) User space DOES still need to handle buffering their own inputs.
2) Use of co-routines removed for 2 reasons.  1 - To simplify user integration patterns , 2 - to remove latency... where we were waiting for the last input update to finish, we were delaying by a whole extra update loop cycle before moving to the next inputs

svolokh:
@RG-nAmKcAz I've pushed an updated version of SimpleRandomMovementBehaviour that directly operates within a coroutine, rather than having the Update cycle wait for a flag to be set by another coroutine. I believe this should reduce the latency. 

My only other idea for further reducing the latency would be to embed the logic in a callback for `InputSystem.onAfterUpdate`, but I could see how that may not be the best example for users.

RG-nAmKcAz:
@svolokh I still don't prefer the coroutines... I was hoping that changes you were making would be in the event sender less so than in this sample user code.  I think where we are right now, we should finish out this PR and open a new task.  I would like to push some changes to the sample class I tried earlier though (no other files), as even the current  version you have pushed potentially can drop keys in some cases based on my original code

RG-nAmKcAz (approved):

PR #236, opened 2024-07-31
[REG-1828] Expanded support for Unity UI (uGUI) in the action manager

svolokh:
This PR adds the capability to the action manager to identify and generate relevant inputs for more types of Unity UI components (besides buttons): toggles, dropdowns, input fields, sliders, and scrollbars. This includes support for both the vanilla and TextMeshPro variants of the components (where applicable). Automated tests are included for all of these capabilities.

With this PR, the monkey bot can now exercise more in-game UI (in particular, in Boss Room it can now create a profile from scratch). This work also enables record/replay of input fields, which were previously unsupported. Demo video showcasing the new capabilities.
https://www.loom.com/share/995667c395124774a9f579c276b18906

Key changes:
1. **RGActionAnalysis.cs**: Scans game objects and prefabs for other UI component types besides buttons
2. **ActionManager/Actions/*.cs**: Lots of new action types added for the newly supported UI components
3. **RGLegacyInputWrapper.cs**: This has been updated to now simulate an appropriate keyboard event for all input fields in the scene. Note that this specifically only targets input fields, the existing input simulation code is sufficient for all the other UI types. 

---


svolokh:
Looks like the test is failing on the CI pipeline (but it passes locally), will debug this tomorrow.

svolokh:
Tests fixed - it looks like the scrollbar interactions don't work when running in headless/batch mode, so I've disabled these when running in the CI pipeline (but they are still enabled when running tests locally).

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/Actions/UIInputFieldSubmitAction.cs
```diff
@@ -0,0 +1,113 @@
...
+    /// Action for pressing/releasing the submit key on the focused input field.
+    /// </summary>
+    public class UIInputFieldSubmitAction : RGGameAction
+    {
+        public string NormalizedGameObjectName { get; set; }
+        
+        public UIInputFieldSubmitAction(string[] path, Type objectType, string normalizedGameObjectName) : 
+            base(path, objectType, new RGBoolRange())
+        {
+            Debug.Assert(typeof(InputField).IsAssignableFrom(objectType) || typeof(TMP_InputField).IsAssignableFrom(objectType));
```
RG-nAmKcAz, line 25: I haven't seen us use this before. Can you give a short summary of how it works.

Normally for things like this we've written code to manually check and used `RGDebug.LogError` explicitly or thrown an approriate exception depending on the code location

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/Actions/UIInputFieldTextEntryAction.cs
```diff
@@ -0,0 +1,195 @@
...
+
+        protected override IEnumerable<RGActionInput> GetActionInputs(int param)
+        {
+            if (param == UIInputFieldTextEntryAction.PARAM_NULL)
+            {
+                foreach (Key key in TextEntryKeys())
+                {
+                    yield return new InputSystemKeyInput(key, false);
+                }
+                yield return new InputSystemKeyInput(Key.Backspace, false);
```
RG-nAmKcAz, line 144: Can you point to an example where this method is called.  It looks unusual with all the chained yield return statements and I'd like to better understand the full usage pattern

src/gg.regression.unity.bots/Runtime/Scripts/RGTestUtils.cs
```diff
@@ -41,7 +41,21 @@ public static IEnumerator WaitForScene(string sceneName, int timeout = 5)
...
         }
-
+        
+        /// <summary>
+        /// Wait for the specified condition to become true with a timeout.
+        /// This is similar to WaitUntil(), but also includes a failing assertion if the timeout expires.
+        /// </summary>
+        /// <param name="condition">The condition that should become true</param>
+        /// <param name="timeout">Maximum time to wait for the condition to become true (in seconds)</param>
+        public static IEnumerator WaitUntil(Func<bool> condition, int timeout = 5)
```
RG-nAmKcAz, line 51: Not related to this method, but having TestUtils in our Runtime folder of our SDK feels off

RG-nAmKcAz (changes requested):
A few questions/concerns.  Most of which are my lack of understanding and needing to understand before approving.

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/Actions/UIInputFieldSubmitAction.cs
svolokh: Every action that is identified during the static analysis is associated with a particular object/component type. In this case, the assertion is just to serve as a restriction on the types of objects this action can be used with. The actual construction of these actions is all handled within the `RGActionAnalysis` class (i.e. it is not a user-facing API) so I just impose this restriction with a simple assertion to serve as a sanity check that the analysis is not doing something unexpected.

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/Actions/UIInputFieldTextEntryAction.cs
svolokh: This ultimately gets called by the bot to determine the set of keyboard inputs to perform on a particular frame. I've decided on a coroutine API for this because for some bots they don't necessarily need to store all the inputs in a buffer, so they could just iterate over all the inputs for that action, e.g.
```
foreach (RGActionInput input in chosenAction.GetInputs(actionParam))
  input.Perform();
```

In the default `RGMonkeyBot` implementation it stores all the inputs into a buffer for that frame:
https://github.com/Regression-Games/RGUnityBots/blob/be91bd43937b64e38da6bed2f9f58027ba0a1ee9/src/gg.regression.unity.bots/Runtime/Scripts/GenericBots/RGMonkeyBotLogic.cs#L174

src/gg.regression.unity.bots/Runtime/Scripts/RGTestUtils.cs
vontell: Maybe we can name this differently - this is more of an artifact of our SDK being a testing framework, so this is not a utility for our tests, but rather for users of our SDK to use in their tests. Happy to rename this!

vontell: Could just be RGUtils

src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs
```diff
@@ -1126,9 +1126,80 @@ private void AnalyzeGameObject(GameObject gameObject)
             {
                 foreach (string listener in RGActionManagerUtils.GetEventListenerMethodNames(btn.onClick))
                 {
-                    _buttonClickListeners.Add(listener);
+                    string[] path = {"Unity UI", "Button", listener};
+                    AddAction(new UIButtonPressAction(path, typeof(Button), listener), null);
                 }
             }
+            if (gameObject.TryGetComponent(out Toggle _) && !IsRGOverlayObject(gameObject))
```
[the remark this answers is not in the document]
svolokh: With the current structure the RGOverlay test only happens if the game object actually contains a UI component. I figure this would be more efficient than doing the overlay test on every game object, which traverses the parents and checks the object names. Though I am open to feedback on better ways to detect if the object is part of the RG overlay canvas.

src/gg.regression.unity.bots/Runtime/Scripts/RGTestUtils.cs
RG-nAmKcAz: We already have at least 1 `RGUtils` class elsewhere in the code, so be careful with naming here

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/Actions/UIInputFieldTextEntryAction.cs
RG-nAmKcAz: Ok, then lets add some big summary code docs around that interface explaining the usage pattern for extenders / implementers, probably including at least one sample code snippet

svolokh: I'm working on a technical reference here on another branch that will reside within the action manager directory for all this work, I think this should cover it? 
https://github.com/Regression-Games/RGUnityBots/blob/sasha/reg-1873/src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/README.md

src/gg.regression.unity.bots/Runtime/Scripts/RGTestUtils.cs
svolokh: Do we want to move the file out into its own assembly, e.g. `RegressionGames.TestFramework`? This is similar to what the Input System does for its own [test framework](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.0/manual/Testing.html).

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/Actions/UIInputFieldSubmitAction.cs
vontell: I'm unfamiliar with the intricacies of this within C#, but in other languages I like to stay away from Assert-based functionality because it can be turned off. Why not do an explicit check and throw an error? That way in release builds, this would still be caught (in releases, [Debug.Assert is removed](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.debug.assert?view=net-8.0))

vontell (approved):
Just a few comments, but nothing blocking as I think they may have been considered from Zack's comments.

src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs
```diff
@@ -324,8 +327,41 @@ private static IEnumerator InputSimLoop()
                     _lastHitObject = hitObject;
                     _lastHitObject2D = hitObject2D;
                 }
+                
+                // If there are active UI input fields, simulate a KeyDown UI event for newly pressed keys
```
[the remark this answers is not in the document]
svolokh: Yes, the event is only sent if the input fields have focus as indicated by the `isFocused` property:

https://github.com/Regression-Games/RGUnityBots/blob/fd09234c971376926c518cafbb654de3b97f8b2a/src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs#L349-L352

src/RGUnityBots/Assets/Scripts/ActionManagerTests/UIHandler.cs
```diff
@@ -0,0 +1,80 @@
+﻿using TMPro;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace ActionManagerTests
```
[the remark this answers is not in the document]
svolokh: All the files in this folder are scripts/components for a "test game" that is part of the ActionManagerTestScene. There are two problems with moving this into the test folder: 1) The scene would be referencing MonoBehaviours from inside the test assembly, which is an odd setup. More importantly, 2) this would cause these files to get ignored by the action analysis, which only targets the Player assemblies.

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/Actions/UIInputFieldSubmitAction.cs
svolokh: Normally that's what I would do if these were user-facing APIs, but these checks are just light sanity checks to catch the errors from RGActionAnalysis sooner during development. Anyways, I have no problem with converting these to exceptions - I'll make the change.

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/Actions/UIInputFieldTextEntryAction.cs
svolokh: I've created a PR with the technical reference here https://github.com/Regression-Games/RGUnityBots/pull/238. Do let me know if you think we should still add an inline comment with this type of example elsewhere.

svolokh:
@RG-nAmKcAz @vontell Since https://github.com/Regression-Games/RGUnityBots/pull/237 has been merged, I've needed to move the input field simulation out of RGLegacyInputWrapper and into KeyboardEventSender, since the legacy input wrapper will no longer be present if the game is exclusively using the new Input System. Re-requesting a review since it's a somewhat non-trivial change.

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/Actions/UIInputFieldTextEntryAction.cs
RG-nAmKcAz: Even if we have the docs, we should have the correct <summary> docs on our APIs and interfaces to enable users within their IDEs and when reading our source.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/KeyboardEventSender.cs
```diff
@@ -83,6 +94,34 @@ private static void QueueTextEvent(int replaySegment, Key key)
                 RGDebug.LogInfo($"({replaySegment}) Sending Text Event for char: '{value}'");
                 InputSystem.QueueEvent(ref inputEvent);
             }
+            
+            // If there are active UI input fields, simulate a KeyDown UI event for newly pressed keys
```
RG-nAmKcAz, line 98: I feel like this code should be wrapped in LEGACY # blocks as I don't think it is necessary for the newer input system.  If a field is selected in the UI, then text events will go to that active field.

RG-nAmKcAz (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/KeyboardEventSender.cs
svolokh: I wasn't able to record/replay input field entry (e.g. the profile name in Boss Room) until I added this code. I'm not sure that these TextEvents actually make it to the InputFields, which are reading keyboard events via `Event.PopEvent` from the UnityGUI event system?

RG-nAmKcAz: interesting.. it indeed does not work

tested by recording bossroom and replaying;  last time i tried this on PW it WAS working, but that was forwarding the text event to gameface and re-rendering.  That should be impacted by this new code as it is not a unity UI component

RG-nAmKcAz: I guess the only part that feels weird is `RGLegacyInputUtils` methods for the new input system as the event system for UI components

src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs
```diff
@@ -3,7 +3,10 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Text.RegularExpressions;
+using TMPro;
```
vontell, line 6: I see imports added but no code changes, is that correct?

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/Actions/UISliderReleaseAction.cs
```diff
@@ -0,0 +1,166 @@
...
+                return null;
+            }
+        }
+
+        protected override bool IsValidActionParameter(float param)
+        {
+            return true;
+        }
+
+        protected override IEnumerable<RGActionInput> GetActionInputs(float param)
```
vontell, line 156: I guess one part I am confused about is exactly how the monkey bot decides where to click on a slider, or what to type into an input. Does the action encapsulate the whole "select slider AND click on a part of it", or is it just selecting, and a separate click action needs to happen? Same for the input fields - is the focus and key press part of that entire action, or are they all separate?

vontell (changes requested):
Just a few more additional comments from me

src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs
svolokh: Good catch - these aren't needed anymore (updated).

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/Actions/UISliderReleaseAction.cs
svolokh: Good question: you'll note that we have two actions for the slider (UISliderPress, UISliderRelease), and for some of the other UI types as well (e.g. InputField has separate actions for pressing/focusing it and text entry). This goes back to the fundamental design of the action manager, where every action is defined as a set of inputs applicable to a _single_ frame. 

For the slider, you'll note that both the press and release actions have a `float` parameter ranging from 0 to 1 that indicate where the mouse should be hovering over the slider. For the press action, it will move the mouse cursor to that location and press the mouse button down, and the release action will move the mouse cursor and release the mouse button. (I've added some comments to the code to elaborate on this)

For the input fields, there are separate actions for focusing the text field (this is generated as a UIObjectPressAction) and entering text (UIInputFieldTextEntryAction), since the latter has a precondition that the text field is already focused. You'll note that the text entry action checks that the input field is focused in its `IsValidForObject` function.

In general, it is up to the agent logic whether to prioritize doing certain sequences of actions (this could be a fixed heuristic, or something learned). As mentioned in the action manager [reference](https://github.com/Regression-Games/RGUnityBots/blob/main/src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/README.md), the RGMonkeyBot has some heuristics to prioritize some common sequences of actions, e.g. releasing the mouse over the same location it was pressed with some greater chance.

src/gg.regression.unity.bots/Runtime/Scripts/RGTestUtils.cs
svolokh: @RG-nAmKcAz @vontell What would be your preferred approach for resolving this?
Options I can see from the above discussion:
1) rename to RGUtils 
2) move into its own test assembly
3) leave as is

RG-nAmKcAz: 1 . Would need to be something other than RGUtils as mentioned above
2. Is fine

For simplicity I would probably go with #1 but just pick some other naming

src/gg.regression.unity.bots/Runtime/Scripts/ActionManager/Actions/UIInputFieldTextEntryAction.cs
svolokh: I've added more usage examples and comments to the summary docs in the code for RGGameAction and related APIs.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/KeyboardEventSender.cs
svolokh: I've moved the UI key event creation/sending methods out of RGLegacyInputUtils and into KeyboardEventSender. This seems like the most appropriate place for them since they are specific to sending keyboard events and aren't related to either the legacy Input Manager or new Input System (they are part of the UnityGUI's own event system).

src/gg.regression.unity.bots/Runtime/Scripts/RGTestUtils.cs
svolokh: I will go with option 2 (moving into its own test assembly), since if we want to have utilities for others to use in Unity Test Runner then having it in its own test framework assembly will also allow us to access/extend the NUnit APIs in the future if desired (these are normally unavailable in the Runtime assembly).

However, we have various places that depend on this class ([SDK docs](https://docs.regression.gg/gameplay-sessions/gameplay-sessions-reference) and  [BossRoom's RecordingTest](https://github.com/Regression-Games/RGBossRoom/blob/main/Assets/Tests/Runtime/RecordingTest.cs)), so I've created a ticket REG-1888 for this and I think it would be better to have this be a separate set of PRs that addresses all these at once.

RG-nAmKcAz (approved):

src/RGUnityBots/Assets/Scripts/ActionManagerTests/UIHandler.cs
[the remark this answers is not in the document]
vontell: Ah I see, thanks for the clarification

vontell:
Looks like my comments are addressed, thanks!

vontell (approved):

PR #237, opened 2024-07-31
[REG-1874] Support games whose active input handling is set to exclusively use the new Input System

svolokh:
This PR addresses an issue where if Build Settings -> Player Settings -> Active Input Handling is set to “Input System Package (New)”, rather than “Both”, the SDK currently errors out due to a variety of dependency on both the old/new input systems throughout the SDK. This adds ENABLE_LEGACY_INPUT_MANAGER checks in all the remaining places where they should be to disable the legacy input usage when the game is exclusively using the new Input System.

As before, it is out of scope to fully remove dependency on the new Input System in the SDK for games that only use the legacy input system. This PR is just to ensure that the configurations of "Both" and "Input System Package (New)" are both working.

This was tested in two ways:
1. The unit tests in the RGUnityBots project were run with active input handling set to New Input System only.
2. Boss Room was re-configured to be set to use the new Input System only, and the record/replay and monkey bot were checked to work as expected.

---

### DIFF
--- a/src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs
+++ b/src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs
@@ -11,15 +11,19 @@ using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using RegressionGames.ActionManager.Actions;
-using RegressionGames.Editor.RGLegacyInputUtility;
 using UnityEditor;
 using UnityEngine.InputSystem;
 using UnityEngine.InputSystem.Controls;
 using Assembly = UnityEditor.Compilation.Assembly;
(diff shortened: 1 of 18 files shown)


src/gg.regression.unity.bots/Editor/Scripts/ActionManager/RGActionAnalysis.cs
```diff
@@ -253,7 +257,7 @@ private IEnumerable<ExpressionSyntax> FindCandidateValuesForLocalVariable(ILocal
         /// </summary>
         private IEnumerable<SyntaxNode> FindDerivedUserInput(ExpressionSyntax expr)
         {
-            const int maxDepth = 5;
+            const int maxDepth = 10;
```
svolokh, line 260: This change (and the other one in this method) is to fix a bug I found with the analysis of raycasts when checking raycast from the new Input System's mouse position.

RG-nAmKcAz (approved):

vontell (approved):

PR #240, opened 2024-08-02
[REG-1882] Make virtual mouse cursor not messup CV analysis

RG-nAmKcAz:
- Changes our playback mouse cursor to be only a tiny dot and not the screen blocking williward head
- Hides the cursor on any run finish so that the cursor isn't stuck on the screen
- downsizes and re-centers the 'click highlight' so that we can more accurately see where the click occurred

https://github.com/user-attachments/assets/7b162867-88ca-467c-ad54-044a266a219d

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Prefabs/RGOverlayCanvasV2.prefab
+++ b/src/gg.regression.unity.bots/Runtime/Prefabs/RGOverlayCanvasV2.prefab
@@ -1,5 +1,81 @@
 %YAML 1.1
 %TAG !u! tag:unity3d.com,2011:
+--- !u!1 &124407930152948339
+GameObject:
+  m_ObjectHideFlags: 0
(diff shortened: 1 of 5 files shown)


vontell (approved):
Looks good! I always wondered if we could just take a screenshot of certain layers of the game by ignoring certain game objects from the texture... but that may be difficult. This looks pretty good for now!

abeizer (approved):
This looks much less intrusive. Also, nice loom background music.

RG-nAmKcAz:
@vontell probably not impossible, but not something that I think would be worth our time anytime soon

it would probably involve writing a shader and putting in the right location in the render pipeline for each game to capture the appropriate layers you want 

we would then also probably need to make the virtual mouse cursor a custom shader instead of an actual  image on the game object

... that and I'd have to look into how to capture and asynchronously read back stored data from a shader (assuming that is something possible)

PR #241, opened 2024-08-02
[REG-1844] Add ActionComplete keyFrameCriteria

abeizer:
Adds new bot segment criteria `ActionComplete` which will wait until a BotAction is complete. For a monobehaviour, this means calling `Destroy` on itself so that the action knows when to flag itself as "done". 

This allows us to move ahead with the blue flower test in the short term, but we may need to consider allowing the behaviour to report a pass/failure state to interrupt the sequence if something goes wrong.

---


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BehaviourCompleteKeyFrameCriteriaData.cs
```diff
@@ -0,0 +1,31 @@
+using System;
+using System.Text;
+using RegressionGames.StateRecorder.JsonConverters;
+
+namespace RegressionGames.StateRecorder.BotSegments.Models
+{
+    [Serializable]
+    public class BehaviourCompleteKeyFrameCriteriaData : IKeyFrameCriteriaData
+    {
+        public int apiVersion = SdkApiVersion.VERSION_1;
```
RG-nAmKcAz, line 10: This needs a new SdkApiVersion defined with a line comment in `SdkApiVersion` describing when this support was added

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/OrKeyFrameCriteriaEvaluator.cs
```diff
@@ -5,13 +5,14 @@ namespace RegressionGames.StateRecorder.BotSegments
 {
     public static class OrKeyFrameCriteriaEvaluator
     {
-        public static bool Matched(bool firstSegment, int segmentNumber, KeyFrameCriteria criteria)
+        public static bool Matched(bool firstSegment, BotSegment botSegment, KeyFrameCriteria criteria)
```
RG-nAmKcAz, line 8: I had intentionally not passed the full BotSegment reference in this methods to limit data access and avoid the potential ambiguity in introduces.  By passing this, implements can now access the entire raw key frame criteria from the segment instead of being limited to the one criteria being passed.  Let's undo this part of the changes please.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/KeyFrameEvaluator.cs
```diff
@@ -80,13 +80,15 @@ public void PersistPriorFrameStatus()
          * <summary>Publicly callable.. caches the statuses of the last passed key frame for computing delta counts from</summary>
          */
 
-        public bool Matched(bool firstSegment, int segmentNumber, List<KeyFrameCriteria> criteriaList)
+        public bool Matched(bool firstSegment, BotSegment botSegment)
```
RG-nAmKcAz, line 83: This method accepting BotSegment seems fine/good, but MatchedHelper and other methods and evaluator class methods should only get the segmentNumber like before these changes as noted in my other comment.

RG-nAmKcAz (changes requested):

abeizer:
@RG-nAmKcAz Updated - back to you

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/KeyFrameEvaluator.cs
```diff
@@ -184,6 +182,13 @@ internal bool MatchedHelper(bool firstSegment, int segmentNumber, BooleanCriteri
                      case KeyFrameCriteriaType.CVImage:
                         cvImagesToMatch.Add(entry);
                         break;
+                    case KeyFrameCriteriaType.BehaviourComplete:
```
RG-nAmKcAz, line 185: Now that I see this all put together.. what if we made this `ActionComplete` , would that work / be more powerful so that we can just wait for any segment action to complete ?

RG-nAmKcAz (commented):
One question before approving... which may set us up better for future use cases.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/KeyFrameEvaluator.cs
vontell: I'll +1 this idea of making it more generic so we can use it elsewhere!

vontell (approved):
Nice! Nothing in addition to what Zack already has commented here. Note that we should make sure to document this, but that might need to wait until we add a new section to the docs for all of this work anyways.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/KeyFrameEvaluator.cs
```diff
@@ -182,7 +182,7 @@ internal bool MatchedHelper(bool firstSegment, int segmentNumber, BotActionType
                      case KeyFrameCriteriaType.CVImage:
                         cvImagesToMatch.Add(entry);
                         break;
-                    case KeyFrameCriteriaType.BehaviourComplete:
+                    case KeyFrameCriteriaType.ActionComplete:
                         if (botActionType == BotActionType.Behaviour && !botActionCompleted)
```
RG-nAmKcAz, line 186: We don't need the action type check here anymore

RG-nAmKcAz (changes requested):
1 small fix remaining

And.. I gave you a merge conflict in SdkApiVersion .... sorry

abeizer:
@RG-nAmKcAz should all be fixed now. Back to you for hopefully the final time :-P

RG-nAmKcAz (approved):

PR #242, opened 2024-08-05
[REG-1876] CV Image Click Actions

RG-nAmKcAz:
- Introduce CV Image click actions
- Fix a few small bugs in CV Image detection around withinRect

- Python Side PR https://github.com/Regression-Games/CV_Python/pull/3

- Sample video, though this doesn't look that impressive....  Basically it makes sure change profile exists in segment 1, then in segment 2, it clicks on that button and confirms that the create new profile button exists
  - You will note that the processing takes a while ... this is better in the current code, but when I took the video there was a bug where the SDK was flooding python with requests

https://github.com/user-attachments/assets/c48a9326-4e2e-46fa-a578-c6f5dba2322a

---


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVImageMouseActionData.cs
```diff
@@ -0,0 +1,376 @@
...
+    public class CVMouseActionDetails
+    {
+        public static readonly int apiVersion = SdkApiVersion.VERSION_10;
+
+        //main 5 buttons
+        public bool leftButton;
+        public bool middleButton;
+        public bool rightButton;
+        public bool forwardButton;
+        public bool backButton;
```
svolokh, line 27: I think it could be a useful feature to have a mouse position offset from the recognized CV image be an optional parameter of the actions, e.g. to aid in specifying click->drag actions.

svolokh (approved):
This looks good! I like the ability to specify sequences of actions for the recognized image. I noted a non-blocking suggestion below.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVImageMouseActionData.cs
RG-nAmKcAz: I started with something like that, but for now decided that a click drag (at least for PW) would be handled as 2 segment files.  One with the mouse down on the first detected image, and a second with a mouse up on the destination detected image.

abeizer (approved):
Non-blocking, wondering if it could get confusing for users seeing CV presented as both a criteria and action. Mainly, how are they different and why are they configured differently. Maybe that's a problem for the UI to solve

PR #243, opened 2024-08-05
[REG-1888] Move RGTestUtils into its own test assembly

svolokh:
Move RGTestUtils into its own test assembly `RegressionGames.TestFramework`. Unity Test Runner assemblies should now reference this assembly to use the SDK's test utilities.

This change is a follow-up to a previous discussion (https://github.com/Regression-Games/RGUnityBots/pull/236#discussion_r1702307430) about separating the test code from the runtime SDK code. Separating into a different assembly gives us access to the test-only APIs (e.g. NUnit) so we can build/extend on these in the future if desired.

Associated PRs:
https://github.com/Regression-Games/RGBossRoom/pull/62
https://github.com/Regression-Games/RegressionDocs/pull/80


src/gg.regression.unity.bots/Tests/TestFramework/Scripts/RegressionGames.TestFramework.asmdef
```diff
@@ -0,0 +1,16 @@
+﻿{
+  "name": "RegressionGames.TestFramework",
+  "references": [
+    "RegressionGames",
+    "UnityEngine.TestRunner",
+    "UnityEditor.TestRunner"
```
RG-nAmKcAz, line 6: Does this `UnityEditor` reference work when we run our build tests in docker ?

RG-nAmKcAz (approved):
Approved with one question

abeizer (approved):

src/gg.regression.unity.bots/Tests/TestFramework/Scripts/RegressionGames.TestFramework.asmdef
svolokh: Yes, you can see it passing in this PR as well above. I also verified that the tests work when run in standalone mode. It is standard for both play mode and edit mode tests to refer to both the `UnityEngine.TestRunner` and `UnityEditor.TestRunner` assemblies (this is the same practice employed by other test framwork libraries too, such as InputSystem.TestFramework).

PR #245, opened 2024-08-07
[REG-1886] Add example Q-learning bot based on action manager

svolokh:
This PR adds an experimental example Q-learning bot that can serve as a reference for how to implement learning agents using the action manager. The bot generates a discrete action space from the action manager contents. It uses a generic/compact state representation based on the set of components present in the scene, and by default uses a curiosity reward based on camera position (tries to maximize the number of camera positions visited). This enables it to get better exploration performance than the monkey bot on some games such as platformers. 

Note that by default this bot is disabled (it does not implement the IRGBot interface), since this is an experimental feature. To get it to appear in the overlay, you can either inherit from it in a MonoBehaviour as documented in the comments or just temporarily modify the SDK code to have it inherit from `IRGBot`.

Here is an example of learning curve seen on Unity's [Platformer Microgame](https://learn.unity.com/project/2d-platformer-template). 
q_learning


src/gg.regression.unity.bots/Runtime/Scripts/GenericBots/Experimental/GameSpeedup.cs
```diff
@@ -0,0 +1,62 @@
...
+namespace RegressionGames.GenericBots.Experimental
+{
+    /// <summary>
+    /// Configures the game to run as fast as possible for faster training/search.
+    /// The speedup is active as soon as this object is constructed.
+    /// When the Dispose() method is invoked, the speedup is disabled.
+    /// The user of this class needs to regularly invoke the Update() method to ensure the speedup
+    /// is retained.
+    /// </summary>
+    public class GameSpeedup : IDisposable
```
vontell, line 13: For some reason I feel like this could be a good small micro blog post (i.e. How to Speed Up Your Unity Game)... I'm gonna write that down

src/gg.regression.unity.bots/Runtime/Scripts/GenericBots/Experimental/QLearningBot.cs
```diff
@@ -0,0 +1,623 @@
...
+    ///
+    /// By default, the bot aims to explore a variety of functionality via a generic exploration reward.
+    /// A custom reward can be specified by overriding the CreateRewardModule() method.
+    ///
+    /// When the bot is first started, it records the initial scene it was started in. The bot runs an episode for a
+    /// fixed amount of time defined by EpisodeDuration, then restarts the game by invoking SceneManager.LoadScene
+    /// on the initial scene it was loaded with. The logic for restarting the game can be modified by overriding the RestartGame() method.
+    ///
+    /// At the end of each episode, the model (Q-table) is saved to ModelFilePath,
+    /// as well as the training state (epsilon value). Therandom action selection rate (epsilon)
```
vontell, line 105:
```suggestion
    /// as well as the training state (epsilon value). The random action selection rate (epsilon)
```

src/gg.regression.unity.bots/Runtime/Scripts/GenericBots/Experimental/QLearningBot.cs
```diff
@@ -0,0 +1,623 @@
...
+            }
+        }
+    
+        /// <summary>
+        /// Generates a discrete action space for the game.
+        /// </summary>
+        protected virtual List<QAction> GenerateActionSpace()
+        {
+            var actionSpace = new List<QAction>();
+            foreach (var act in RGActionManager.Actions)
```
vontell, line 236: Cool to see it in action!

src/gg.regression.unity.bots/Runtime/Scripts/GenericBots/Experimental/QLearningBot.cs
```diff
@@ -0,0 +1,623 @@
...
+                _qTable.Add(entry.Key, actionVals);
+            }
+            return true;
+        }
+
+        public void Start()
+        {
+            if (!RGActionManager.IsAvailable)
+            {
+                RGDebug.LogError("Action manager is currently unavailable");
```
vontell, line 304: This caught me up in one of my early times using the framework. Mind if we add this?
```suggestion
                RGDebug.LogError("Action manager is currently unavailable. Have you run Regression Games > Analyze Actions on your project?");
```

src/gg.regression.unity.bots/Runtime/Scripts/GenericBots/Experimental/QLearningBot.cs
```diff
@@ -0,0 +1,623 @@
...
+                        {
+                            qact.CurrentInstances.Add(inst);
+                        }
+                    }
+                    if (qact.CurrentInstances.Count > 0)
+                    {
+                        _validActions.Add(qact);
+                    }
+                }
+            }
```
vontell, line 374: What exactly is happening here? Is there a reason why ComputeValidActions does not just equal RGActionManager.GetValidActions()?

vontell (changes requested):
This is really cool to see! Mostly just some small edits and a question. Also, you mentioned an example to get this working, do we have an example project with this?

src/gg.regression.unity.bots/Runtime/Scripts/GenericBots/Experimental/GameSpeedup.cs
batu: Very cool, I would have benefited from that blog post! I now learned that there is to it than just setting timescale.

src/gg.regression.unity.bots/Runtime/Scripts/GenericBots/Experimental/QLearningBot.cs
svolokh: Fixed!

src/gg.regression.unity.bots/Runtime/Scripts/GenericBots/Experimental/QLearningBot.cs
svolokh: That does look more helpful! I went ahead and added that here and to the RGMonkeyBot as well.

batu (changes requested):
The improvement looks really strong and there is clear signal that the approach is working.  

A few thoughts:
**Baselines**
I think in these types of plots it is important to have the upper (human performance) and lower (random agent or simplest script) bound  represented so we have an understanding of where our approach falls. 

**In this case would you please be able to plot the performance of a random agent?** Of course it wouldn't be over episodes, so one approach would be to collect n episodes ( 10 < n < 100) and plotting the average as with error bars as a horizontal line, I assume it will be around 22, which is where our agent started but it will be good to check. (If you have the time, a human performance would be good as well!)

**Error Bars**
It is important for us to understand the variance in our approaches, as variance has a disproportionate impact on the user experience. Given this, it would be useful to have multiple seeds plotted. (And have the variance of our own approaches be compared to the variance of the upper and lower bounds.)

This is especially important in **Deep** reinforcement learning. Especially when there is the so called [deadly triad: Bootstrapping, Off-policy learning, function approximation](https://k4ntz.medium.com/why-is-there-a-deadly-triad-issue-and-how-to-handle-it-e1839b1a8b40). When we have the deadly triad we don't have convergence guarantees, in practice, this means there is higher variance. 

I know that we are not using function approximations but I believe we hit the other two of the triad. So it is not essential to have multiple seeds but it would be useful.

Just to illustrate how bad things can be: 
image
This is showing the same algorithm, save everything, but with wildly different results. And this isn't even an off policy method! 

**Plotting Reward vs Actual Metric**
Rewards are useful but we always should plot the actual metric we are interested in as well. Total number of discrete camera positions? Agents hack rewards, all the time, and it is important to find out whether the agent touched all tiles once or many times to reach this reward would be important.

TL:DR / Update:
Non of these are blocking requests, as it doesn't relate to the code. However I would love to see the average reward of the random agent as a comment later on in this PR!

batu (approved):

svolokh:
@batu I think that plot should definitely be doable, I'll just run it against the RGMonkeyBot. Will follow up with the requested graphs.

svolokh:
@vontell I've added the demo platformer I've been testing the camera exploration reward with to the LegacyInputGames repo (in the folder `PlatformerMicrogame`). Running this Q learning bot with it gives an agent that eventually learns to get out of the initial area and move to the right.

RG-nAmKcAz:
@svolokh Too many moving projects.. but we have that same project checked in https://github.com/Regression-Games/PlatformerSample with an a* pathfinding algorithm bot implemented.  Hasn't been updated in a while, but you may want to update the project in some way to indicate it's original purpose, deprecate it, and point to your new one (which I agree is in a better spot)

src/gg.regression.unity.bots/Runtime/Scripts/GenericBots/Experimental/GameSpeedup.cs
RG-nAmKcAz, line 13: @svolokh Make sure to test this out with project steelline.  They use timed animations of the cards shuffling into the game and I ran into timing issues with those in the past recording bot segments.

RG-nAmKcAz (approved):

svolokh:
@RG-nAmKcAz I will take a look. The A* bot will probably do a lot better than this Q learning bot though (the main point here was just to create a generic bot type that can do better than the monkey bot for some games).

vontell (approved):

svolokh:
@batu Here are the requested graphs for number of camera positions visited per episode as the model learns. However, I was not able to meaningfully compare the rewards of the RGMonkeyBot and QLearningBot due to challenges in syncing up their action rate, so I did not include graphs for rewards.

With the default level in the platformer, this is what we get: 
default_lvl

However, the default level has a big pit early on that the basic Q learning agent struggles to jump over. We can see more of a difference if the pits are removed and the agent is allowed to move further.
flat_lvl

RG-nAmKcAz:
@svolokh direct link to the a* bot code location for reference 

https://github.com/Regression-Games/PlatformerSample/tree/main/Assets/RegressionGames/Runtime/Bots/SampleBot_Platformer2D

src/gg.regression.unity.bots/Runtime/Scripts/GenericBots/Experimental/GameSpeedup.cs
svolokh: I tested it out and it doesn't seem to break with the time scaling, but I did notice that with the more performance-intensive game the time scaling results in a very huge deltaTime that isn't great for training the bots. So I've reduced the default training scale to something that is probably more reasonable for most games (from 20 -> 3). It could be worth exploring dynamically adjusting the timeScale based on unscaledDeltaTime in the future.

PR #250, opened 2024-08-09
[REG-1900] Input playback mouse keyboard interleaving

RG-nAmKcAz:
- Interleaves mouse and keyboard inputs from a segment by time

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/Models/InputData.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/Models/InputData.cs
@@ -28,11 +28,14 @@ namespace RegressionGames.StateRecorder.Models
             if (allInputsSorted == null)
             {
                 allInputsSorted = new();
+
+                // add all the keyboard data, and then we will interleave the mouse data into it
                 allInputsSorted.AddRange(keyboard);
 
                 foreach (var mouseInputActionData in mouse)
                 {
                     var allInputsSortedCount = allInputsSorted.Count;
+                    var didAdd = false;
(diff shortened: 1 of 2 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/Models/InputData.cs
```diff
@@ -9,17 +9,60 @@ namespace RegressionGames.StateRecorder.Models
...
+                            // if an existing mouse time is after the new mouse time.. insert the new mouse before this one
+                            if (mouseData.startTime > mouseInputActionData.startTime)
+                            {
+                                allInputsSorted.Insert(i, mouseInputActionData);
+                                break;
+                            }
+                        }
+                    }
+                    // didn't find something to put it before, stick it on the end
+                    allInputsSorted.Add(mouseInputActionData);
```
svolokh, line 60: I believe there's a bug here: This line will get called for every mouse input, even if a place for it is already found above.

svolokh (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/Models/InputData.cs
RG-nAmKcAz: good catch, ty

RG-nAmKcAz: fixed

svolokh (approved):

abeizer (approved):

PR #253, opened 2024-08-12
[REG-1884] Don't send 2 of same key in same frame

RG-nAmKcAz:
- Stops us from sending 2 events for the same key within the same frame
- Adds the current frame # to event send messages for clarity/debugging
- Test: Close in time, or even intentionally making the end before the start don't write in the same frame

image

image

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/InputPlaybackActionData.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/InputPlaybackActionData.cs
@@ -4,6 +4,7 @@ using System.Text;
 using RegressionGames.StateRecorder.JsonConverters;
 using RegressionGames.StateRecorder.Models;
 using UnityEngine;
+using UnityEngine.InputSystem;
 
 namespace RegressionGames.StateRecorder.BotSegments.Models
 {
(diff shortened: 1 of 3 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/InputPlaybackActionData.cs
```diff
@@ -51,10 +52,19 @@ public bool ProcessAction(int segmentNumber, Dictionary<long, ObjectStatus> curr
 
                 var allInputs = inputData.AllInputsSortedByTime();
 
+                HashSet<Key> keysSeenThisUpdate = new();
+                bool stopKeyboard = false;
+
                 foreach (object input in allInputs)
                 {
-                    if (input is KeyboardInputActionData replayKeyboardInputEntry)
+                    if (!stopKeyboard && input is KeyboardInputActionData replayKeyboardInputEntry)
```
vontell, line 60: I am potentially missing something or a little bit confused - wouldn't this stop any other key events from being sent as well? For instance, if my actions are press A, A, then B, wouldn't B fail to ever be executed?

vontell (changes requested):
Question about the behavior of this, but seems sound otherwise!

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/InputPlaybackActionData.cs
vontell: Ah I see... that makes more sense now. Thanks!

vontell (approved):

abeizer (approved):

svolokh (approved):
This makes sense given the way the keyboard recordings are stored. 

Perhaps it could make sense one day to explore capturing the actual underlying down/up events themselves (e.g. with [InputEventTrace](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.5/manual/Events.html)), which could reduce the need for such handling.

PR #256, opened 2024-08-14
[REG-1883] Update action manager and legacy input unit tests to be more precise

svolokh:
This PR updates the unit tests for the action manager and legacy input support to be more precise and wait for the minimum amount of frames needed for actions to take place (previously, the tests were imprecise and waited for 2-4 frames with the hope that it was enough for the event processing to take place).

Changes:
1. Add the ability to manually invoke the update loop of the legacy input support, so that the tests can have more precise control over when the update happens.
2. Minimize the number of frames waited for as much as possible, and add comments indicating why extra frames are waited for. For tests about the new Input System, the tests now specifically wait for an update step to complete rather than waiting an arbitrary number of frames.

---


RG-nAmKcAz (approved):
Very nice

vontell (approved):

PR #259, opened 2024-08-15
[REG-1907] SDK Cleanup to Finalize Blue Flower Tests

RG-nAmKcAz:
- Makes `apiVersion` field optional in Bot Segments
- Allows Zip files to contain both BotSegments and BotSegmentLists interleaved
- Properly sorts zip file entries

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentZipParser.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentZipParser.cs
@@ -87,6 +87,7 @@ namespace RegressionGames.StateRecorder.BotSegments
                     using var sr = new StreamReader(entry.Open());
 
                     var fileText = sr.ReadToEnd();
+                    badSegmentName = entry.FullName;
 
                     try
                     {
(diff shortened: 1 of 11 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentZipParser.cs
```diff
@@ -30,24 +85,61 @@ public static List<BotSegment> ParseBotSegmentZipFromSystemPath(string zipFilePa
...
-                        versionMismatch = frameData.EffectiveApiVersion;
-                        break;
-                    }
+                        // see if this is a bot segment list file
+                        var botSegmentList = JsonConvert.DeserializeObject<BotSegmentList>(fileText);
 
-                    badSegmentName = entry.FullName;
+                        if (botSegmentList.EffectiveApiVersion > SdkApiVersion.CURRENT_VERSION)
+                        {
+                            versionMismatch = botSegmentList.EffectiveApiVersion;
```
vontell, line 98: Just want to make sure I understand this, so this now allows the bot segment list to set the version for the underlying segments so that each segment doesn't need a version, is that a correct statement?

vontell (approved):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentZipParser.cs
RG-nAmKcAz: The effective api version checking system HAS NOT CHANGED.

What I did is removed the apiVersion as a required field on every element.

When saving bot segments during recording they will still be written there and enforced.

When using our UI tools, they will also still be written and enforced.

What this change allows is for hand written files to not need to specify them and they will default to being the 'latest' version for each model present.  Makes developers editing these files way less tedious.  What I did as a 'hybrid' was to put the version I wanted just at the top level of each of my files, which made it much easier to manage/understand when hand editing the files.

(If you were to load one of these files on an older SDK version, it may or may not load depending on which fields are there and you just wouldn't get the clean 'version unsupported' message.)

RG-nAmKcAz: I could see an argument for still making it required on BotSegmentLists and maybe BotSegments at the top level

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentZipParser.cs
```diff
@@ -30,24 +85,61 @@ public static List<BotSegment> ParseBotSegmentZipFromSystemPath(string zipFilePa
...
+                            results.Add(frameData);
+                        }
+                        catch (Exception ex)
+                        {
+                            throw new Exception($"Invalid/missing file data while parsing BotSegment or BotSegmentList from zip file path: {zipFilePath} at entry: {entry.Name}", ex);
+                        }
                     }
 
-                    results.Add(frameData);
+                    badSegmentName = entry.FullName;
```
svolokh, line 142: If a version mismatch error happens above, then the loop would break before `badSegmentName` gets set. Perhaps this should be happening at the top of the loop?

svolokh (changes requested):
Looks good! I noted a possible issue below in the error reporting.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentZipParser.cs
RG-nAmKcAz: yep, that's from before i made this throw.. will fix, good catch

svolokh (approved):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentZipParser.cs
vontell: Got it, that seems fine to me!

addisonbgross (approved):

PR #260, opened 2024-08-19
[REG-1901] Bot Sequences UI List View 

addisonbgross:
image

## What has been done
- Created a new version of `RGOverlay` that gets the honour of having the original name (no v3)
  - The overlay can be opened and closed just like the previous version
  - The list of Bot Sequences can be refreshed by clicking the refresh button. This can be done after creating a new JSON Sequence file
- Created components used within the RGOverlay:
  - `SequenceCard`: Shows the high level information for a Bot Sequence located within the user's project. Will be used to start, edit, or delete the Sequence
  - `CreateSquenceCard`: Will be used to show the 'create a new Bot Sequence' UI
  - Created a bunch of new art assets to make these components look pretty

## Out of scope
- Managing Bot Sequences in any way. We're just loading and showing them for this task
- Sorting or filtering Bot Sequences. I think this is something that should be added after we complete the base feature set, our end to end flows. I'll be keeping it mind when laying out the new UI pages 🧠  
- Tests. It looks like the only way we can test this UI is by creating a test scene and using the Test Runner. There are no existing tests for the RGOverlay AFAIK, and I think it would be good to brainstorm a solid test scene we can use to exercise our UI(s)

## LOOM
https://www.loom.com/share/858aef6f32fd4e1e876f9ec682591ba2


src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceEntry.cs
```diff
@@ -0,0 +1,86 @@
...
+     */
+    [SerializeField]
+    public Component showByDefault;
+
+    [SerializeField]
+    public Component hideByDefault;
+
+    void Start()
+    {
+        if (nameComponent != null)
```
batu, line 43: Will things work if these fields are not set? If they will cause a crash we can consider telling the user things are missing right at this level, instead of having a downstream crash with a harder to debug error message.

batu (approved):
This looks good! I have some non-blocking questions/suggestions.

batu:
Beauty! Very nice looking update. 

- I definitely agree on the using the face over full body Williward.
- The clipped description makes me think whether we need a 'title' field in our boss segments? If I am organizing these segments I see myself using both the title and description.
- Given that we are creating new concepts (Bot Segments isn't something people are familiar with outside of regression) I think we need to be very aggressive in first time user experience. A dedicated FTIU pass might be in order for our flow later down the line, but should we consider having a `Help` button that opens up the relevant documentation page?

batu:
Oh whoops! I used [this](https://github.com/Regression-Games/RGBossRoom/blob/4b272b82c17ce6953b33e603db57390846535f0b/Assets/RegressionGames/Resources/BotSegments/cv_text_action_menu_changeprofilebutton/2.json) json to look at the fields, instead of your actual video. My bad!

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceEntry.cs
addisonbgross: These `*Component` fields are set by us! In this file, the *Component fields have all been manually set in their parent prefab. It is possible that this assignment gets messed up by us when updating the prefab in the future, but in the worst case these fields will just be empty (eg: if the `nameComponent` isn't set, we wont' see the name on the Sequence Card)

Having said all of this, I'm a complete novice with Unity dev. If there's any improvements I can make at all, I'm game!

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSequence.cs
```diff
@@ -36,6 +36,11 @@ public class BotSequence
          */
         public string description;
 
+        /**
+         * <summary>LastModified is the date that the sequence file was last altered</summary>
+         */
+        public DateTime lastModified;
```
vontell, line 42: Do we need to update a version field for this change? I just noticed that I don't see any api version field on this data type... I wonder if there is a reason for that. We can keep this as a note to ask when Zack gets back I suppose.

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public void LoadSequences()
+    {
+        if (Directory.Exists(SEQUENCES_PATH))
+        {
+            // ensure we arent' appending new sequences on to the previous ones
+            ClearExistingSequences();
+
+            IEnumerable<string> sequenceFiles = Directory.EnumerateFiles(SEQUENCES_PATH, "*.json");
+            foreach (string fileName in sequenceFiles)
```
vontell, line 53: Before this, could we map the files to their loaded bot sequences and then sort by last modified in descending order? I think as we have more and more sequences, it will be useful to have the most recently modified one at the front. Thoughts?

```csharp
var sequences = sequenceFiles
            .Select(f => BotSequence.LoadSequenceJsonFromPath(f))
            .OrderByDescending(seq => seq.lastModified)
            .ToList();
```

vontell (approved):
Looking good! Just had a small suggestion

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSequence.cs
addisonbgross: I think we're ok without an `apiVersion` for now. This `BotSequence` file doesn't have any consumers other than this brand new UI work I'm doing

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
addisonbgross: Works for me! Done

RG-nAmKcAz:
@addisonbgross @vontell 

This PR has outstanding issues that needed addressing before merge.  I realize it was already merged but it has bugs that need a follow up fix asap.

The immediate ones I noticed were...
1. RGSequenceEntry extends MonoBehaviour but defines a 'name' field.  This is invalid/ bad-practice as it hides the monobehaviour name field.
2. BotSequence adds lastModified , but assumes the sequences are loaded from a filesystem directory, but in a production runtime these are highly likely to be loaded as a Resource, not from a directory and this API call will fail with an exception.
3. RGSequenceManager assumes that we have file system access to a directory of sequences.  But again here, in production runtime these will be resources in the game, NOT a directory

PR #261, opened 2024-08-19
[REG-1913] Open vocab text query object detection bot segment criteria. 

batu:
[Loom video](https://www.loom.com/share/7d55604f8e2c445daabbf32dab9569c5?sid=613050f0-da98-48c3-9078-fcb358121575).

Implemented the open vocab text query bot segment criteria. 

[Example segments on Boss Room.](https://github.com/Regression-Games/RGBossRoom/pull/67)

There are some follow up tasks that are related to this:
- [Refactor this and CV Image match evaluators.](https://linear.app/regression-games/issue/REG-1928/refactor-cv-object-detection-evaluator)
- [Update Python endpoints to match.](https://linear.app/regression-games/issue/REG-1920/update-python-object-detection-endpoints-to-match-sdk)

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVObjectDetectionEvaluator.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVObjectDetectionEvaluator.cs
@@ -94,7 +94,9 @@ namespace RegressionGames.StateRecorder.BotSegments
             {
                 // Check if we have results or a request in progress.
                 requestInProgress = _requestTracker.ContainsKey(segmentNumber);
-
+                
+                // TODO(REG-1928): Think more about hardening this data reading.
+                
                 // Get the results for the segment.
                 _queryResultTracker.TryGetValue(segmentNumber, out objectDetectionResults);
                 // Get the prior results for the segment.
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVObjectDetectionEvaluator.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVObjectDetectionEvaluator.cs
@@ -328,7 +328,7 @@ namespace RegressionGames.StateRecorder.BotSegments
                             },
                             textQuery: textQuery,
                             queryImage: null,
-                            withinRect: null,
+                            withinRect: withinRect,
                             index: 1
                         ),
                         // Cancel ongoing request in a thread safe manner.
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVObjectDetectionEvaluator.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVObjectDetectionEvaluator.cs
@@ -83,45 +83,84 @@ namespace RegressionGames.StateRecorder.BotSegments
         // Returns a list of non-matched entries
         public static List<string> Matched(int segmentNumber, List<KeyFrameCriteria> criteriaList)
         {
-            List<string> resultList = TryGetResults(
-                segmentNumber,
-                out ConcurrentDictionary<int, List<CVObjectDetectionResult>> objectDetectionResults,
-                out List<string> priorResults,
-                out bool requestInProgress);
+            RGDebug.LogVerbose($"CVObjectDetectionEvaluator - Matched - botSegment: {segmentNumber} - BEGIN");
 
-            // only query when no request in progress and when we have cleared out the prior results
+            ConcurrentDictionary<int, List<CVObjectDetectionResult>> objectDetectionResults = null;
+            List<string> priorResults = null;
+            bool requestInProgress = false;
+
+            // Try to get the results or request in progress.
+            lock (_requestTracker)
+            {
+                // Check if we have results or a request in progress.
+                requestInProgress = _requestTracker.ContainsKey(segmentNumber);
+
+                // Get the results for the segment.
+                _queryResultTracker.TryGetValue(segmentNumber, out objectDetectionResults);
+                // Get the prior results for the segment.
+                _priorResultsTracker.TryGetValue(segmentNumber, out priorResults);
+
+                // Print the results if there are any.
+                if (objectDetectionResults != null)
+                {
+                    var resultsString = $"\n[{string.Join(",\n", objectDetectionResults.Select(pair => $"{pair.Key}:[{(pair.Value == null ? "null" : string.Join(",\n", pair.Value))}]"))}]";
+                    RGDebug.LogDebug($"CVObjectDetectionEvaluator - Matched - botSegment: {segmentNumber} - cvImageResults: {resultsString}");
+                }
+            }
+
+            var resultList = new List<string>();
+
+            // If no results and no request, we'll initiate a new request.
             if (!requestInProgress && objectDetectionResults == null)
             {
-                TryPostRequest(segmentNumber, criteriaList, resultList);
+                resultList = TryPostRequest(segmentNumber, criteriaList, resultList);
             }
-            else if (!requestInProgress)
+            // If we have results and no request in progress, we'll evaluate them.
+            // We check priorResults null here so that we only evaluate a new result 1 time.
+            else if (!requestInProgress && objectDetectionResults != null && priorResults == null)
             {
-                EvaluateResult(segmentNumber, criteriaList, priorResults, objectDetectionResults, resultList);
+                resultList = EvaluateResult(segmentNumber, criteriaList, objectDetectionResults, resultList);
+            }
+            // If a request is in progress, we'll wait for the results.
+            else
+            {
+                resultList.Add("Awaiting CV Image evaluation results ...");
             }
 
             lock (_requestTracker)
             {
                 _priorResultsTracker[segmentNumber] = resultList;
             }
-
             RGDebug.LogVerbose($"CVObjectDetectionEvaluator - Matched - botSegment: {segmentNumber} - resultList: {resultList.Count} - END");
 
             return resultList;
         }
 
-        private static void EvaluateResult(
+        /// <summary>
+        /// Matches the given criteria list against the object detection results for a specific segment.
+        /// </summary>
+        /// <param name="segmentNumber">The number of the segment being evaluated.</param>
+        /// <param name="criteriaList">The list of criteria to match against.</param>
+        /// <param name="priorResults">The list of prior results for the segment.</param>
+        /// <param name="objectDetectionResults">The list of object detection results for the segment.</param>
+        /// <param name="resultList">The list of result strings to be updated.</param>
+        /// <returns>A list of non-matched entries as strings.</returns>
+        /// <remarks>
+        /// This method performs the following steps:
+        /// 1. Attempts to retrieve existing results for the given segment.
+        /// 2. If no request is in progress and no results exist, it initiates a new request.
+        /// 3. If results are available and no request is in progress, it evaluates the results against the criteria.
+        /// 4. Updates the prior results tracker with the current result list.
+        /// </remarks>
+        private static List<string> EvaluateResult(
             int segmentNumber,
             List<KeyFrameCriteria> criteriaList,
-            List<string> priorResults,
             ConcurrentDictionary<int, List<CVObjectDetectionResult>> objectDetectionResults,
             List<string> resultList)
         {
             int criteriaListCount = criteriaList.Count;
-            bool areCountsEqual = objectDetectionResults.Count == criteriaListCount;
             bool allResultsHaveValue = objectDetectionResults.All(r => r.Value != null);
-            // we check priorResults null here so that we only evaluate a new result 1 time
-            if ((priorResults == null && objectDetectionResults != null &&
-                 areCountsEqual && allResultsHaveValue))
+            if ((objectDetectionResults.Count == criteriaListCount && allResultsHaveValue))
             {
                 // Evaluate the results if we have them all for each criteria
                 for (var i = criteriaListCount - 1; i >= 0; i--)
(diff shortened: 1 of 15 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVService/CVObjectDetectionRequest.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        public void WriteToStringBuilder(StringBuilder stringBuilder)
+        {
+
+            bool textQuery = queryText != null;
+            bool imageQuery = queryImage != null;
+
+            if (textQuery && imageQuery)
+            {
+                RGDebug.LogError("You need to use either textQuery or imageQuery. You can not have both.");
```
vontell, line 39: Does our WriteToStringBuilder approach throw errors in other instances? Trying to determine if this should be caught here or elsewhere.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVObjectDetectionKeyFrameCriteriaData.cs
```diff
@@ -0,0 +1,56 @@
...
+using RegressionGames.StateRecorder.BotSegments.JsonConverters;
+using RegressionGames.StateRecorder.JsonConverters;
+
+namespace RegressionGames.StateRecorder.BotSegments.Models
+{
+    [Serializable]
+    public class CVObjectDetectionKeyFrameCriteriaData : IKeyFrameCriteriaData
+    {
+        // update this if this schema changes
+        public int apiVersion = SdkApiVersion.VERSION_9;
```
vontell, line 14: Is this the latest version? There may be a newer one, not sure.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVObjectDetectionKeyFrameCriteriaData.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public string text;
```
vontell, line 16: Let's add a comment on what this is used for. Same for other class variables.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVObjectDetectionKeyFrameCriteriaData.cs
```diff
@@ -0,0 +1,56 @@
...
+        /**
+       * base64 encoded byte[] of jpg image data , NOT the raw pixel data, the full jpg file bytes
+       */
+        // public string imageData; This will be added in the next PR.
```
vontell, line 20: If there is a ticket for this, we usually do the following so we know:

```suggestion
        // TODO(REG-???): Supporting image queries will be added soon
        // public string imageData;
```

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/KeyFrameEvaluator.cs
```diff
@@ -193,8 +197,44 @@ internal bool MatchedHelper(bool firstSegment, int segmentNumber, bool botAction
...
+                    // cvText results will be null until we get the async response back
+                    if (andOr == BooleanCriteria.And)
+                    {
+                        _newUnmatchedCriteria.Add("Waiting for CV Object Detection evaluation results ...");
+                        return false;
+                    }
+                }
+                else
+                {
+                    var cvObjectDetectionResultsCoount = cvObjectDetectionResults.Count;
```
vontell, line 215:
```suggestion
                    var cvObjectDetectionResultsCount = cvObjectDetectionResults.Count;
```

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/KeyFrameEvaluator.cs
```diff
@@ -193,8 +197,44 @@ internal bool MatchedHelper(bool firstSegment, int segmentNumber, bool botAction
...
+                        {
+                            if (andOr == BooleanCriteria.And)
+                            {
+                                _newUnmatchedCriteria.Add(resultEntry);
+                                return false;
+                            }
+                        }
+                    }
+                }
+            }
```
vontell, line 236: Assuming this is basically a copy-paste of the logic for the other CV stuff? If so, 👍

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVObjectDetectionEvaluator.cs
```diff
@@ -0,0 +1,449 @@
...
+using RegressionGames.StateRecorder.BotSegments.Models;
+using RegressionGames.StateRecorder.BotSegments.Models.CVService;
+using RegressionGames.StateRecorder.JsonConverters;
+using UnityEngine;
+
+namespace RegressionGames.StateRecorder.BotSegments
+{
+
+    /**
+     * <summary>Evaluates CV Image criteria using CVServiceManager to send/receive HTTP requests to a python server for doing the actual CV evaluations.
```
vontell, line 14:
```suggestion
     * <summary>Evaluates CV Object criteria using CVServiceManager to send/receive HTTP requests to a python server for doing the actual CV evaluations.
```

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVObjectDetectionEvaluator.cs
```diff
@@ -0,0 +1,449 @@
...
+     * Python 'detects' a source image template in a provided screenshot and then this class evaluates those results against the specified bot segment criteria.</summary>
```
vontell, line 15:
```suggestion
     * Python 'detects' classes of objects in a provided screenshot based on image and text queries, and then this class evaluates those results against the specified bot segment criteria.</summary>
```

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVObjectDetectionEvaluator.cs
```diff
@@ -0,0 +1,449 @@
...
+            var resultList = new List<string>();
+
+            // If no results and no request, we'll initiate a new request.
+            if (!requestInProgress && objectDetectionResults == null)
+            {
+                resultList = TryPostRequest(segmentNumber, criteriaList, resultList);
+            }
+            // If we have results and no request in progress, we'll evaluate them.
+            // We check priorResults null here so that we only evaluate a new result 1 time.
+            else if (!requestInProgress && objectDetectionResults != null && priorResults == null)
```
vontell, line 120: I noticed that this check was slightly different from the check in the other file. Just curious to learn more about that!

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVObjectDetectionEvaluator.cs
```diff
@@ -0,0 +1,449 @@
...
+        /// <param name="objectDetectionResults">The list of object detection results for the segment.</param>
+        /// <param name="resultList">The list of result strings to be updated.</param>
+        /// <returns>A list of non-matched entries as strings.</returns>
+        /// <remarks>
+        /// This method performs the following steps:
+        /// 1. Attempts to retrieve existing results for the given segment.
+        /// 2. If no request is in progress and no results exist, it initiates a new request.
+        /// 3. If results are available and no request is in progress, it evaluates the results against the criteria.
+        /// 4. Updates the prior results tracker with the current result list.
+        /// </remarks>
```
vontell, line 154: Really really really appreciate all of the comments in this file!

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVObjectDetectionEvaluator.cs
vontell: I see, this check is in the other method now. Thanks!

vontell (changes requested):
Honestly, amazing amazing work. I know that code could be difficult to sift through, this PR got in much faster than I thought it would. Really nice work! Just a few small comments really.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVObjectDetectionKeyFrameCriteriaData.cs
batu: That definitely was not the newest one. Updated to the newest of 15.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVService/CVObjectDetectionRequest.cs
batu: No, you are definitely right. This is the wrong place to do it. I think we should do it in the constructor

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/KeyFrameEvaluator.cs
batu: This is a copy paste. I don't fully understand this yet, but I am getting there.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVService/CVObjectDetectionRequest.cs
batu: I thought about this a little more. This is better but not good enough. Because it is likely the constructor will never actually receive both of these values, Especially if they are set as they are now, like so:

```
                    _ = CVServiceManager.GetInstance().PostCriteriaObjectTextQuery(
                        new CVObjectDetectionRequest(
                            screenshot: new CVImageBinaryData()
                            {
                                width = width,
                                height = height,
                                data = imageData
                            },
                            queryText: queryText,
                            queryImage: null,
                            withinRect: null,
                            index: 1
                        ),
```

Where this check needs to live is in parsing from the json. If in the criteria match json they give both text AND image they should be notified then and there. I added a TODO and the corresponding task.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVObjectDetectionKeyFrameCriteriaData.cs
batu: Done

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVService/CVObjectDetectionResult.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Text;
+using RegressionGames.StateRecorder.JsonConverters;
+using UnityEngine;
+
+namespace RegressionGames.StateRecorder.BotSegments.Models.CVService
+{
+    public class CVObjectDetectionResult
+    {
+        public string text;
```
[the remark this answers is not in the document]
batu: This is very wrong. Fixing now, however the fact I missed it is worrying and potentially calls for a hardening.

Basically this is what happened:
1) I assumed compiler would complain about a mismatch between result json and string, when it didn't I didn't double check. This is not the case, as we use 'TryGetValue' to fill in the results, which does its best to load results but doesn't enforce a match.
```c#
public bool TryGetValue(TKey key, out TValue value)
    {
      int entry = this.FindEntry(key);
      if (entry >= 0)
      {
        value = this._entries[entry].value;
        return true;
      }
      value = default (TValue);
      return false;
    }
```
2) The criteria only cares about existence of a match, not whether the data returned is correct (well because we don't have ground truth). So even if the type was very wrong, the behavior worked as expected. This would have fallen over very hard at the action phase.

It is likely there is a reason why this is the way it is, so I am adding a link to the future [REG-1928](https://linear.app/regression-games/issue/REG-1928/refactor-cv-object-detection-evaluator) to discuss hardening

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVService/CVObjectDetectionRequest.cs
batu: Ended up fixing it in this PR. However, I have a question about logging errors:
image

With both 
```
RGDebug.LogError("Both textQuery and imageQuery cannot be provided simultaneously. Use only one.");
throw new ArgumentException("Both textQuery and imageQuery cannot be provided simultaneously. Use only one.");
```
I feel like we are duplicating information? Should RGDebug.Log actually raise an error?

vontell: yeah I'm looking through our code and I am confused about how exactly the pattern would work, I can't find any spot where we throw an error... it seems like we always just terminate, stop loops, etc... 

I suppose putting it into that CVObjectDetectionKeyFrameCriteriaData parsing file is fine... there is an argument to be made that this is not an error for the JSON parsing and is more of an error for the business logic later on... but it gets weird since we are defining this data format and parsing the JSON in the same class, so the class could impose the restrictions I suppose... this seems fine to me!

TLDR; yeah I think this is all good for now, but let's make a note for @RG-nAmKcAz to check this out when he gets back in case he has a pattern in mind that we were following.

vontell: (mark this as resolved when you read it)

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVService/CVObjectDetectionResult.cs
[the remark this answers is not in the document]
vontell: Good catch!

vontell (commented):
Just one more small request related to API version - otherwise, looking good!

vontell (approved):

addisonbgross (approved):
Pending Aaron's comments, this looks reasonable to me!

PR #262, opened 2024-08-20
[REG-1906] Start Sequences from Sequence overlay

addisonbgross:
## What has been done
- Enabled the play button on the Bot Sequence Card to actually start the Sequence
- Ensured that the RGOverlay is in the correct state upon starting a Sequence

## LOOM
https://www.loom.com/share/bfa22fdc15ff44c5b8ec32f5e2914354

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Prefabs/RGOverlay/SequenceCard.prefab
+++ b/src/gg.regression.unity.bots/Runtime/Prefabs/RGOverlay/SequenceCard.prefab
@@ -575,6 +575,7 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   name: 
   description: 
+  playButton: {fileID: 2445726624697383442}
   nameComponent: {fileID: 714608286913792316}
   descriptionComponent: {fileID: 0}
   lastModifiedComponent: {fileID: 2750996074632129499}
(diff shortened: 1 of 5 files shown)


vontell (approved):
Nice! Quick question though (I should have seen this in the other PR, and I probably know this but my early morning brain is having trouble) are sequences automatically detected from the project, or do they all need to be added manually?

addisonbgross:
@vontell Sequences will be loaded automatically when the Unity game starts. There is a little refresh button within the RGOverlay that is used to reload the Sequences if you make a new one, or delete one, from outside of Unity

batu (approved):
This looks good!

batu:
Are the defaults logs we are seeing user facing? If so I think they can be improved. 

Both looking at this PR and working with criteria matching on my end, I find it non trivial to parse what is the current state of our bot segment runs. I opened up a [task](https://linear.app/regression-games/issue/REG-1926/evaluate-whether-our-debug-logs-are-informative-to-our-end-users) to go over them.

addisonbgross:
@batu good thinking with the debug log review! I wasn't sure who our target audience is for those logs, it makes sense that it will be our users. I've subscribed to your Linear tasks, as I'm game to help figure out the best messaging across the SDK

PR #265, opened 2024-08-23
[REG-1935] Add missing UnityEngine import

vontell:
It just so happens that the BotSequences file got by without using any UnityEngine imports when used in the editor, but does rely on the import inside of builds. Added the missing import (this wasn't caught before because the lines that use this import are inside of `!UNITY_EDITOR` pragmas.


batu (approved):
Nice catch

addisonbgross (approved):

PR #266, opened 2024-08-23
[REG-1912] Implement bot segment criteria for open vocab image queries.

batu:
[Loom video.](https://www.loom.com/share/1bd67e94a1624c109b7a9a555251aa4c)
Related [Unity Boss Room example](https://github.com/Regression-Games/RGBossRoom/pull/69).

Implemented object detection criteria match using image query.
---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVServiceManager.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVServiceManager.cs
@@ -72,13 +72,29 @@ namespace RegressionGames
             );
         }
 
-        public async Task PostCriteriaObjectTextQuery(CVObjectDetectionRequest request, 
+        public async Task PostCriteriaObjectDetection(CVObjectDetectionRequest request, 
                                                       Action<Action> abortRegistrationHook,
                                                       Action<List<CVObjectDetectionResult>> onSuccess,
                                                       Action onFailure)
         {
+            
+            string endpoint;
+            if (!string.IsNullOrEmpty(request.textQuery))
+            {
+                endpoint = "/criteria-object-text-query";
+            }
+            else if (request.imageQuery != null)
+            {
+                endpoint = "/criteria-object-image-query";
+            }
+            else
+            {
+                RGDebug.LogError("Invalid CVObjectDetectionRequest: Both textQuery and queryImage are null or empty.");
+                onFailure.Invoke();
+                return;
+            }
             await SendWebRequest(
-                uri: $"{GetCvServiceBaseUri()}/criteria-object-text-query",
+                uri: $"{GetCvServiceBaseUri()}/{endpoint}",
                 method: "POST",
                 payload: request.ToJsonString(),
                 abortRegistrationHook: abortRegistrationHook.Invoke,
@@ -94,6 +110,7 @@ namespace RegressionGames
                 }
             );
         }
+
         
         public async Task PostCriteriaTextDiscover(CVTextCriteriaRequest request, Action<Action> abortRegistrationHook, Action<List<CVTextResult>> onSuccess, Action onFailure)
         {
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVServiceManager.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVServiceManager.cs
@@ -81,18 +81,24 @@ namespace RegressionGames
             string endpoint;
             if (!string.IsNullOrEmpty(request.textQuery))
             {
-                endpoint = "/criteria-object-text-query";
+                endpoint = "criteria-object-text-query";
             }
             else if (request.imageQuery != null)
             {
-                endpoint = "/criteria-object-image-query";
+                endpoint = "criteria-object-image-query";
             }
-            else
-            {
-                RGDebug.LogError("Invalid CVObjectDetectionRequest: Both textQuery and queryImage are null or empty.");
+            else if (request.imageQuery != null && request.textQuery != null){
+                RGDebug.LogError("Invalid CVObjectDetectionRequest: Both textQuery and queryImage are both set. Please set one or the other.");
                 onFailure.Invoke();
                 return;
             }
+            else
+            {
+                RGDebug.LogError("Invalid CVObjectDetectionRequest: Both textQuery and queryImage are null or empty. Please set at least one of them.");
+                onFailure.Invoke();
+                return;
+            }
+            
             await SendWebRequest(
                 uri: $"{GetCvServiceBaseUri()}/{endpoint}",
                 method: "POST",
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVService/CVObjectDetectionRequest.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVService/CVObjectDetectionRequest.cs
@@ -37,7 +37,7 @@ namespace RegressionGames.StateRecorder.BotSegments.Models.CVService
         public CVObjectDetectionRequest(CVImageBinaryData screenshot,
                                         [CanBeNull] string textQuery,
                                         [CanBeNull] string imageQuery,
-                                        [CanBeNull] RectInt? withinRect,
+                                        RectInt? withinRect,
                                         int index = 0)
         {
             this.screenshot = screenshot;
@@ -71,18 +71,10 @@ namespace RegressionGames.StateRecorder.BotSegments.Models.CVService
 
             stringBuilder.Append("{\"screenshot\":");
             screenshot.WriteToStringBuilder(stringBuilder);
-            
-            if (textQuery != null)
-            {
-                stringBuilder.Append(",\"textQuery\":");
-                StringJsonConverter.WriteToStringBuilder(stringBuilder, textQuery);
-            }
-
-            if (imageQuery != null)
-            {
-                stringBuilder.Append(",\"imageToMatch\":");
-                StringJsonConverter.WriteToStringBuilder(stringBuilder, imageQuery);
-            }
+            stringBuilder.Append(",\"textQuery\":");
+            StringJsonConverter.WriteToStringBuilder(stringBuilder, textQuery);
+            stringBuilder.Append(",\"imageToMatch\":");
+            StringJsonConverter.WriteToStringBuilder(stringBuilder, imageQuery);
             stringBuilder.Append(",\"withinRect\":");
             RectIntJsonConverter.WriteToStringBuilderNullable(stringBuilder, withinRect);
             stringBuilder.Append(",\"index\":");
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVService/CVObjectDetectionRequest.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVService/CVObjectDetectionRequest.cs
@@ -14,9 +14,9 @@ namespace RegressionGames.StateRecorder.BotSegments.Models.CVService
         public CVImageBinaryData screenshot;
 
         /// <summary>
-        /// Optional image data to be used as a query for object detection.
+        /// Optional image data to be used as a query for object detection in base64 encoded string format.
         /// </summary>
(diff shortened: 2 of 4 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVServiceManager.cs
```diff
@@ -72,13 +72,29 @@ await SendWebRequest(
...
 
-        public async Task PostCriteriaObjectTextQuery(CVObjectDetectionRequest request, 
+        public async Task PostCriteriaObjectDetection(CVObjectDetectionRequest request, 
                                                       Action<Action> abortRegistrationHook,
                                                       Action<List<CVObjectDetectionResult>> onSuccess,
                                                       Action onFailure)
         {
+            
+            string endpoint;
+            if (!string.IsNullOrEmpty(request.textQuery))
```
RG-nAmKcAz, line 82: Do we document somewhere that text takes precedence over image ?

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVServiceManager.cs
```diff
@@ -72,13 +72,29 @@ await SendWebRequest(
...
+            }
+            else
+            {
+                RGDebug.LogError("Invalid CVObjectDetectionRequest: Both textQuery and queryImage are null or empty.");
+                onFailure.Invoke();
+                return;
+            }
             await SendWebRequest(
-                uri: $"{GetCvServiceBaseUri()}/criteria-object-text-query",
+                uri: $"{GetCvServiceBaseUri()}/{endpoint}",
```
RG-nAmKcAz, line 97: The `endpoint` already has a leading `/` leading to an invalid `//` in the path

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVServiceManager.cs
RG-nAmKcAz: Re-looking at this, I feel like these probably need to be separate criteria to prevent any ambiguity / user errors.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVService/CVObjectDetectionRequest.cs
```diff
@@ -36,8 +36,8 @@ public class CVObjectDetectionRequest
 
         public CVObjectDetectionRequest(CVImageBinaryData screenshot,
                                         [CanBeNull] string textQuery,
-                                        [CanBeNull] CVImageBinaryData imageQuery,
-                                        RectInt? withinRect,
+                                        [CanBeNull] string imageQuery,
+                                        [CanBeNull] RectInt? withinRect,
```
RG-nAmKcAz, line 40: [CanBeNull] seems weird/un-necessary on a nullable type

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVService/CVObjectDetectionRequest.cs
```diff
@@ -81,7 +81,7 @@ public void WriteToStringBuilder(StringBuilder stringBuilder)
             if (imageQuery != null)
```
RG-nAmKcAz, line 81: these != null checks don't really need to exist, just let the StringJsonConverter write out a `null` value for the json field

RG-nAmKcAz (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVServiceManager.cs
batu: Image doesn't take precedence over image. If you specify both or neither, we catch it earlier on. 

Here is [the relevant conversation](https://github.com/Regression-Games/RGUnityBots/pull/261#discussion_r1727120478) about where to catch the both/neither field filled in error. Open to more thoughts here.

But I will add a catch here as well.

RG-nAmKcAz: The fact that we now have 2 code places to check for a potential user error on the same misconfiguration is a red flag.  

We have 2 different apis, and based on these code patterns, we should change these to be 2 different criteria type names in the sequence json to resolve this

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVService/CVObjectDetectionRequest.cs
batu: I like the visual consistency of things that are can be null but yeah it is redundant. Removed

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVServiceManager.cs
batu: I am coming at it from a user centric perspective, not from our internal code structure. 

When I am thinking about using an API, I often think about solving a problem, in this case, I want to detect an object. I see CVObjectDetection type and that tells me I can use this endpoint to detect an object. Now I go into the docs to understand how to use it and what the options are, in this case there are two.

Seeing two options as CVObjectDetectionText and CVObjectDetectionImage, would make me think one detects text and one detects images, which is not what we want.

We can consider CVObjectDetectionViaText and CVObjectDetectionViaImage, or something more clear on the user side, but imho these are too verbose and can be combined.

batu: @RG-nAmKcAz If you feel strongly about this I am happy to change it, but I am not fully convinced that our internal code patterns should dictate user facing decisions.

Happy to talk about where the right place for the image/text both filled check should live

RG-nAmKcAz: my perspective is that we have code trying to avoid an obvious user ambiguity/error path of our own creation, and we could avoid it by being more verbose.

To me, the difference in being verbose vs being clean but ambiguous would always go towards verbose.

-----

To clarify this case, I suggest this.

CVObject (or CVObjectDetection)

should have the following fields in its json data

`type` - An enum of either `Text` or `Image`, but could this also have a 3rd+ enum value later ?
`data` - The data for this as either a json {} or just as a string
- I suggest maybe using a json object if we expect other potential types later that might need more data ?
  - for text - data:{'text':"..."} // named to be consistent with CVText
  - for image - data:{'imageData':"..."}   // named to be consistent with CVImage
- But if just a string
  - data:"..."

batu: I agree with being more verbose! Thats why I prefer 'textQuery/imageQuery' over 'text/image' or CVObjectDetection over CVObject so we are definitely aligned there.

I am definitely closer to this approach over CVObjectDetectionText, so thank you for clarifying.

My only remaining concern is that this approach has a higher surface area for error on the user side:
```
CVObjectDetection {
  type: 
  data: {
     'text': "..."
      or
     'imageData': "..."
}
```

If we go with this approach now they have two places they need to change. They have to both align the type and the correct input key. (and also need to know that the payload needs to be wrapped in the data object)

```
CVObjectDetection {
  type: "..."
  data: "..."
}
```
If we go with this approach, then there is just one place to update (type) but then it is ambiguous in this context what the data type should be.  Of course they can infer it from the type just above, but it is an extra step. They need to get two things correct. (type and data)

```
CVObjectDetection {
  textQuery: "..."
or
  imageQuery: "..."
}
```

I agree that in this case people can easily make the error of adding both of them, which would lead them to an error path. Maybe I see this less problematic than you, as I don't consider well formed, early errors educating our users to be a big issue. In other words, I would prefer making this mistake, realizing that I can't use both, then updating my mental model and still using the simpler approach.

Simpler in terms of number of things to update get correct, which is just one, over the 2 or more in other approaches.

batu: I think we are both pointing out important concerns, should we get other folks take on this design?

RG-nAmKcAz: I asked @vontell to provide an opinion

vontell: Filling in with my two cents over here - when I think about this particular part of the API, I have a hard time not thinking about it from the ML side, which is the following: "We have an image, and we want to detect some object, and we do this by embedding some query". And so from that standpoint, I'd rather us have this as one criteria, since it is all the same algorithm and approach. I think part of having these more advanced techniques available to these developers is also informing them a bit on how things work under the hood, and having it be the same one criteria will enforce the idea that they are using the same algorithm. I think if we also separate it out, we also just trade this ambiguity for some new ambiguity - i.e. a user could ask "CVObjectDetectionViaImage and CVImage both take an image and find it in the screenshot, how do I know which one to use?"

Comparing the two query approaches (i.e. `textQuery` and `imageQuery` vs `type` and `data`), I really don't see one being better than the other... textQuery and imageQuery can be filled at the same time and the user gets an error, and with type, you run the risk of them still defining both but now it selects the type and there is no error, and it uses the right one, but a reader of the JSON may wonder why both are set. I personally would just like to trust the user here in that they at least get the JSON from a sample or our docs and didn't construct it out of thin air.

My belief here is that we are probably just overcomplicating this a bit and overestimating how confused a user might get from the ambiguity? I think we can just make sure to have a very concise piece of documentation, something like below, and also make sure the error messages are quite clear. 

> CVObjectDetection will search for a certain class of objects in an image, using one of either a text query or an image query. This uses the OWLv2 approach for searching an image using natural language.

> ### Querying with Text
> ...
> ### Querying with an Image

If I am being 100% honest, I don't think we will really get any user errors here, because they would be copy and pasting an example, or probably using some tool provided by us to define the image query anyways since copy and pasting a b64 encoded string is not going to be our final solution. If anything, there is likely to be more user error from that process than accidentally including both queries, in which case the feedback is immediate from the API. 

If we found that the error rate in our API of people doing both image and text queries at the same time was some surprising amount, it will be pretty trivial to add two new criteria (CVObjectDetectionViaText and CVObjectDetectionViaImage) and mark this one as deprecated, but I really think this is unlikely. But in terms of moving on, I say we keep it as one criteria for now and either go with 

```
CVObjectDetection {
  textQuery: "..."
or
  imageQuery: "..."
}
```

or if we want to include the possibility of making more configs later and organizing this a bit, we could do

```
CVObjectDetection {
  query: {
      text: "..."
     or
      image: "..."
  }
}
```

TLDR; I think the current approach here is fine, aligns more closely with the algorithm that the developer should at least be aware of, and has a clear error message. I'm in support of when there are (at least in my opinion, I know this might be different amongst ourselves) a small difference like this, we can always iterate and see what the users actually behave like. If it was implemented the other way (i.e. two separate criteria) I would have probably asked to combine it (I'm pretty sure I recommended at the beginning of the task to have it as one anyways). For the query stuff, I see them as similar in different ways... so I'd just go with what we have.

RG-nAmKcAz: respectfully dissenting but not dying on this hill either.. approved

RG-nAmKcAz (approved):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/CVServiceManager.cs
batu: @RG-nAmKcAz Thanks for the approval! 

I would like to learn what aspect of it are you disagreeing it with, so I can incorporate that into my code going forward, only if you have the time. (For example, I decided not to use any early return in this code after learning you dislike it, and after reading up on it I am happy to not use them here.)

Does the disagreement rest on the field 'type' being more explicit over doing the separation via 'textQuery/imageQuery'?

vontell (approved):

PR #267, opened 2024-08-23
[REG-1915] Implement bot segment action for clicking based on open vocab image queries.

batu:
[Loom video.](https://www.loom.com/share/a6657745b2674b7e8c08da81000b8503?sid=da3f217b-1e1d-4a7c-93a9-d38ccc057a13)

[Related Unity Boss Room example.](https://github.com/Regression-Games/RGBossRoom/pull/70)

Implement the bot segment action using image queries.

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVObjectDetectionMouseActionData.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVObjectDetectionMouseActionData.cs
@@ -24,7 +24,7 @@ namespace RegressionGames.StateRecorder.BotSegments.Models
         public int apiVersion = SdkApiVersion.VERSION_17;
 
         [NonSerialized]
-        public static readonly BotActionType Type = BotActionType.Mouse_ObjectDetectionText;
+        public static readonly BotActionType Type = BotActionType.Mouse_ObjectDetection;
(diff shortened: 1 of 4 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/CVObjectDetectionMouseActionData.cs
```diff
@@ -138,7 +138,7 @@ private void RequestCVObjectDetectionTextEvaluation(int segmentNumber)
                                     data = screenshot
                                 },
                                 textQuery: textQuery,
-                                imageQuery: null,
+                                imageQuery: imageQuery,
```
RG-nAmKcAz, line 141: Similar / same comment as
https://github.com/Regression-Games/RGUnityBots/pull/266/files#r1732779487

It feels strange to allow the user to pass both queries, but then prefer the text query endpoint.  I feel like these need to be 2 separate criteria to avoid this potential user configuration error.

RG-nAmKcAz (changes requested):

vontell (approved):
Approving based on the other conversation around the matter of configuring this action. Just to reiterate in a different way from the other explanation, I'd rather have users thinking about this as one algorithm/approach.

RG-nAmKcAz (approved):

PR #264, opened 2024-08-22
[REG-1914] Implement object detection via text query mouse action.

batu:
[Explainer video.](https://www.loom.com/share/ff4be77da2eb4b3f979140748f806a43?sid=696825ea-4017-4b72-b03e-0e1cdd8c51ea)
[Relevant examples PR.](https://github.com/Regression-Games/RGBossRoom/pull/68)

Implements object detection via text query mouse action.

This is going to need a huge refactor as there is A LOT of code duplication. But for sake of simplicity, we are first checking in working changes, then refactoring.

---

Find the pull request instructions [here](https://www.notion.so/regressiongg/Pull-Request-Process-0d57a7eb582a446983edfacafa406f1e?pvs=4)

Every reviewer and the owner of the PR should consider these points in their request (feel free to copy this checklist so you can fill it out yourself in the overall PR comment)

- [ ] The code is extensible and backward compatible
- [ ] New public interfaces are extensible and open to backward compatibility in the future
- [ ] If preparing to remove a field in the future (i.e. this PR removes an argument), the argument stays but is no longer functional, and attaches a deprecation warning. A linear task is also created to track this deletion task.
- [ ] Non-critical or potentially modifiable arguments are optional
- [ ] Breaking changes and the approach to handling them have been verified with the team (in the Linear task, design doc, or PR itself)
- [ ] The code is easy to read
- [ ] Unit tests are added for expected and edge cases
- [ ] Integration tests are added for expected and edge cases
- [ ] Functions and classes are documented
- [ ] Migrations for both up and down operations are completed
- [ ] A documentation PR is created and being reviewed for anything in this PR that requires knowledge to use
- [ ] Implications on other dependent code (i.e. sample games and sample bots) is considered, mentioned, and properly handled
- [ ] Style changes and other non-blocking changes are marked as non-blocking from reviewers

PR #269, opened 2024-08-26
[REG-1937] Bot Sequence UI post-list-view updates

addisonbgross:
## What has been done

- Removed the `lastModified` property from the Sequence Card prefab. Since we're not able to have this info in a production build, we've opted to remove it for now
image

- Renamed the Bot Sequence field `name` to `sequenceName`
- Ensure that we can read Sequences in both the Unity editor and in a build
- BONUS: Removed a random `image` component I had on the main Bot Sequences panel component. This was preventing a running Segment from actually interacting with the UI 🤦🏻


src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
```diff
@@ -33,6 +33,14 @@ private void Awake()
     {
         _sequences = new List<BotSequence>();
 
+        // the path to sequences varies depending on if the overaly is used within the Unity editor
+        // or within an actual build of the game
+#if UNITY_EDITOR
+        _sequencePath = "Assets/RegressionGames/Resources/BotSequences";
+#else
+        _sequencePath = Application.persistentDataPath + "/BotSequences";
+#endif
```
addisonbgross, line 42: @RG-nAmKcAz  does this fix the issue you pointed out?
`3. RGSequenceManager assumes that we have file system access to a directory of sequences.  But again here, in production runtime these will be resources in the game, NOT a directory`

RG-nAmKcAz: Not fully

You'll have to handle resources for production.

You can steal/borrow some code from `BotSequence.LoadJsonResourcesFromDirectory` to handle loading every json from a path (whether as a resource or persistent file).  However, you can't use that method directly as it will only load from the filesystem if any file is found there, as opposed to providing the combined set of filesystem overrides + resources built into the game

RG-nAmKcAz (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
addisonbgross: I thiiiink I have it figured out based on that function you mentioned. Its split into 3 cases:
- In the Unity Editor, use the hardcoded path: `Assets/RegressionGames/Resources/BotSequences`
- In a build, check to see if the persistent data path exists: `Application.persistentDataPath` + `/` + `BotSequences` exists
- In a build with nothing in the persistent data path? Use the `BotSequences` directory

batu:
This looks good! I love PRs that delete more then they add.

Note: the nullable context warnings will be resolved when [this PR ](https://github.com/Regression-Games/RGUnityBots/pull/264) gets merged in.

batu (approved):

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
```diff
@@ -44,12 +45,12 @@ private void Awake()
      */
     public void LoadSequences()
     {
-        if (Directory.Exists(SEQUENCES_PATH))
+        if (Directory.Exists(_sequencePath))
```
RG-nAmKcAz, line 48: 2 things
1. You can't use the Directory API to load Resources in the built game binary.
2. Somewhat related to 1 ... we need to build the set of the directory + the resources when running in the built binary.  The same name 'may' appear in both in which case we should prefer loading the one from the directory, but there will also be some that have not been overridden that only appear as an in game resource.

RG-nAmKcAz (changes requested):

vontell (approved):
Approving assuming that Zacks comments will be eventually resolved. That part can be tricky, @addisonbgross do you have the required context for how to start tackling that?

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
RG-nAmKcAz: More context from slack conversation...


```
for a single sequence loading...
Editor - Loads from the Assets folder using File APIs
Runtime - Loads from the persistent data path (if said file exists), else falls back to loading as a resource - effectively treating the persistent data path as an override of any compiled in sequences


So... For finding all the sequences that exist...
Editor - Use file apis on the Assets folder path (you have this part)
Runtime - Needs to find all the sequences in the persistent data path (you have this part)
                   PLUS - also needs to find all sequences that are resources (you do NOT have this part) ... and only load those that aren't overridden (same file name and sequence name inside json) in the persistent data path
```

addisonbgross:
@vontell should be good, thanks. Zack has helped clear things up

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
```diff
@@ -105,4 +85,96 @@ private void ClearExistingSequences()
...
+        // 2. load Sequences from Resources, while skipping any that have already been fetched from the
+        //    persistentDataPath. We will compare Sequences by their filename (without extension), and by the actual
+        //    Sequence name
+        const string runtimePath = "BotSequences";
+        var jsons = Resources.LoadAll(runtimePath, typeof(TextAsset));
+        foreach (var jsonObject in jsons)
+        {
+            try
+            {
+                var resourceFilename = jsonObject.name;
```
addisonbgross, line 156: @RG-nAmKcAz hopefully this works for comparing loaded Sequences for a build of a game.

I'm grabbing the filename (without the file extension) when reading from persistent data, and excluding any Sequences loaded from Resources if they have the same file name (and Sequence name within the file). From what I read, this `jsonObject.name` call should get a Resources file name without its extension

vontell (approved):

RG-nAmKcAz (approved):

addisonbgross:
@RG-nAmKcAz thanks for the guidance on this one. I've added a note about ensuring the Sequence list is nice and scrollable to [my current task.](https://linear.app/regression-games/issue/REG-1903/create-new-sequence-view) The Sequences are already in a scrollable container, but I haven't done much testing with it yet

PR #273, opened 2024-09-03
[REG-1938] Performance Optimizations

RG-nAmKcAz:
- Performance: Optimize bounds computations for better scaling/performance
- Performance/BugFix: Properly Filter Entities captured in the state to those visible (like Transforms)
- Performance: Use slightly more efficient APIs when forcing application focus
- Performance: Refactor JSON patterns to avoid the slow serializer library altogether in most cases
- Performance: Use delegates to avoid reflection performance cost of component data lookup
- Feature: Add options for ECS state [default:true] and component data capture [default:false]
  - Disabling component data capture makes a large performance impact
---
- Results
  - Bot segment spawns 20 entity creatures and I moved the camera to have them on , off , on screen to show the differences
  - Light blue is the time we spent creating the state JSON 
  - Vertical white line indicates an area of profile when the entities were off screen
  
- Before
BEFORE_skinny

- After
OnVsOffCameraEntities

- After - No Component Data captured
**[THIS IS OUR NEW DEFAULT SETTING]**
NoComponentData

- After - No Entities captured
NoEntities

---


abeizer (approved):

vontell (approved):
Had a great paired walkthrough on this with Abby, makes sense!

PR #276, opened 2024-09-08
[REG-1903] Create new Bot Sequences in Unity

addisonbgross:
image

## What has been done
- Users can now use the Sequence Editor to create a brand new Sequence, composed of Segments
- Users need to add a `name` and at least 1 Segment before a new Sequence can be created. The description is optional
- Added and updated some icons

## Out of Scope
- Testing (the next task)
- Editing and deleting Sequences (a future task)
- Scrollable list of Segments composing a Sequence
  - I had burnt too much time on trying to figure this out. The code should be simple-ish, but I was really struggling to get the Canvas parts correct. [A task has been created to cover this functionality](https://linear.app/regression-games/issue/REG-1952/within-the-sequence-editor-ensure-that-the-list-of-segments-composing)

## LOOM
https://www.loom.com/share/2b89b55e1e5547eba981ea6dad2f45a2


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
```diff
@@ -232,5 +234,170 @@ public void WriteToStringBuilder(StringBuilder stringBuilder)
...
+            {
+                var entry = new BotSequenceEntry
+                {
+                    path = fileName,
+                };
+
+                try
+                {
+                    var segment = JsonConvert.DeserializeObject<BotSegmentList>(json);
+                    entry.description = segment.description;
```
RG-nAmKcAz, line 373: We should check the version support when loading these files in.  Segments/SegmentLists/etc

```csharp
                if (segment.EffectiveApiVersion > SdkApiVersion.CURRENT_VERSION)
                {
                    throw new Exception($"Bot segment list file contains a segment which requires SDK version {segment.EffectiveApiVersion}, but the currently installed SDK version is {SdkApiVersion.CURRENT_VERSION}");
                }
```

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
```diff
@@ -232,5 +234,170 @@ public void WriteToStringBuilder(StringBuilder stringBuilder)
...
+         * Recursively look through directories for Segement files, and load them (in a build of the game)
+         * </summary>
+         * <param name="path">
+         * Directory to search for Segments
+         * </param>
+         */
+        private static List<BotSequenceEntry> LoadSegmentsInDirectoryForRuntime(string path)
+        {
+            var results = new List<BotSequenceEntry>();
+            var directories = Directory.EnumerateDirectories(path);
```
RG-nAmKcAz, line 317: Short Term
This API is not usable in the runtime environment for resources.  (or any File/Directory API)

I would expect you to be able to load these using only the resources api but only from that top level path.


Longer term:
I will create a task -> We'll need to make this a build time operation so we have more control and store the listing of our segments/sequences/etc into a scriptable object that we can then utilize here later instead of digging for resources / folders at runtime.

RG-nAmKcAz (changes requested):
- Overall UI functionality - A+
- I have some minor nits over UI styling/spacing, but I suspect you will be editing those as you work on the scroll pane, so not really relevant here.
- Blocker (Possibly future task if needed)- Need to handle sequence version checks on load.  I think (though this is just one opinion) that we could try to load these, but show them in the UI in some way that they aren't selectable and say incompatible version.
- Blocker - Remove use of File/Directory APIs for resource loads
  -  Also, seeing this makes me realize we'll need to handle those cases differently... I'm creating a task for that.

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceEditor.cs
```diff
@@ -0,0 +1,314 @@
...
+         */
+        public void OnSearchInputChange(string text)
+        {
+            if (string.IsNullOrEmpty(text))
+            {
+                CreateAvailableSegments(_segmentEntries);
+            }
+            else
+            {
+                _filteredSegmentEntries = _segmentEntries.Where(s => s.entryName.Contains(text.Trim())).ToList();
```
vontell, line 252: From your loom video, it might be good to .lowercase() these strings so you can search without remembering capitalization

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
```diff
@@ -232,5 +234,170 @@ public void WriteToStringBuilder(StringBuilder stringBuilder)
...
+            {
+                segments = LoadSegmentsInDirectoryForRuntime(persistentDataPath);
+            }
+        
+            // 2. load Segments from Resources, and ensure that no Segments have duplicate names
+            const string runtimePath = "BotSegments";
+            segments = segments.Concat(LoadSegmentsInDirectoryForRuntime(runtimePath)).ToList();
+#endif
+
+            return segments;
```
vontell, line 268: We definitely need to make sure to document this well, I was confused last week when trying out this new overlay because I couldn't go and look for my sequences like with the file explorer.

vontell (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
```diff
@@ -232,5 +234,170 @@ public void WriteToStringBuilder(StringBuilder stringBuilder)
...
+                }
+
+                // recurse into the next level of directories
+                results = results.Concat(
+                    LoadSegmentsInDirectory(directory)
+                ).ToList();
+            }
+
+            return results;
+        }
```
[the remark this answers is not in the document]
addisonbgross: good catch. I tested it out, and the segments are currently not being reloaded from disk when pressing the play button again. I've created a task to resolve this: https://linear.app/regression-games/issue/REG-1970/ensure-bot-segments-sequences-are-reloaded-from-disk-when-pressing-the

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
addisonbgross: Do we want to have both the new UI and file explorer for choosing segments?

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
addisonbgross: To clarify, regarding the short term: Should I expect that the Bot Segments I want to load will only be at the top level path? 

When loading Segments in editor mode, the Segments are nested into different folders

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceEditor.cs
addisonbgross: good catch, added 👍🏻

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
RG-nAmKcAz: Short (from planning).. runtime builds just won't work well until I do the build time parsing task.  So i would just get the directory api / loop out of this PR so it doesn't crash/do weird stuff.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
addisonbgross: As per our cycle planning meeting: we're just going to leave the file explorer button where it is for now (on the recording toolbar, outside of the new UI)

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
addisonbgross: I've added a check for both the Segment and SegmentList types. I've compared like `segment.EffectiveApiVersion >= SdkApiVersion.CURRENT_VERSION`, I think equal versions is ok

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
addisonbgross: I removed the use of the Directory function, so now we're just searching the `BotSegments` path only. It seems to load all of the Segments/SegmentLists in a build of RGBossRoom. So that's nice!

addisonbgross:
@RG-nAmKcAz I'm hoping to have some time during this cycle to circle back to aesthetics. I'll be sure to ask you for that input then! There's some wonky 9-sliced images in there bothering me 👀

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
```diff
@@ -346,25 +350,44 @@ private static BotSequenceEntry ParseSegment(string json, string fileName)
...
 
+                // First we will try to parse the file as a SegmentList, if that fails we will try to parse
+                // the file as a standard Segment
                 try
                 {
-                    var segment = JsonConvert.DeserializeObject<BotSegmentList>(json);
-                    entry.description = segment.description;
-                    entry.entryName = segment.name;
+                    var segmentList = JsonConvert.DeserializeObject<BotSegmentList>(json);
+                    if (segmentList.EffectiveApiVersion >= currentVersion)
```
RG-nAmKcAz, line 364: This should be `>`

We are checking the bad case where the sequence/segment/etc is TOO new for the current runtime.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
RG-nAmKcAz: This should be >

We are checking the bad case where the sequence/segment/etc is TOO new for the current runtime.

RG-nAmKcAz (changes requested):
Versioning checking criteria needs to be `>`

RG-nAmKcAz (approved):
Approved pending aaron's feedback

batu:
Overall this looks cleeeeeeeeaaaan! Non blocking nits from the video:

- Is an empty sequence meaningful? If not, should the create button be available only after one segment is added? -- 😮 Oh wait! This already works this way, but I missed your verbal comments at the start of the video. Maybe the suppress the color change of the button until one segment is added?
- I am believe the 'pull out to delete' gesture will be intuitive to most people--but likely not all. When we are demonstrating how to use this feature we should make a point to demonstrate that as well/keep an eye out for this fail case. If we observe it we can show a trash bin icon when we are dragging a segment.
- I don't think it is necessary in this PR, but I would really like to be able to double click (or click a button) on a segment and have it open the .json in the editor.

batu:
I was going to ask whether we should have different icons depending on action/match but each bot segment can contain actions and matches. This means it is hard to know what a segment 'does' at a glance. This is not an actionable thought but I think this can be a 'design' smell that we should reevaluate at some point

vontell (approved):

PR #279, opened 2024-09-11
[REG-1965] Handle AI Service auth tokens

RG-nAmKcAz:
- Passes auth tokens to AI Service
- Refactors service manager patterns to singletons to remove monobehaviour links and remove warnings when using the settings menus
- Updates prefabs to remove behaviour references
- Refactors/CleansUp CV naming to be AI

---


src/gg.regression.unity.bots/Runtime/Scripts/RGServiceManager.cs
```diff
@@ -49,7 +40,7 @@ public bool IsAuthed()
             return _rgAuthToken != null;
         }
 
-        public bool LoadAuth()
```
addisonbgross, line 52: Nit: It could be nice to have an overloaded use of this method, to match how it gets used a bunch below. Like:
```
public bool LoadAuth()
{
    return LoadAuth(out _);
}
```

addisonbgross (approved):
Aside from one tiny nit, this LGTM

abeizer (approved):

src/gg.regression.unity.bots/Runtime/Prefabs/RGOverlayCanvas.prefab
```diff
@@ -2893,7 +2893,6 @@ GameObject:
   - component: {fileID: 2671455581720235337}
```
vontell, line 2893: Thank you for updating both prefabs!

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/AIServiceManager.cs
```diff
@@ -51,98 +53,110 @@ private String GetCvServiceBaseUri()
...
-                    var response = JsonConvert.DeserializeObject<CVImageResultList>(s);
-                    onSuccess.Invoke(response.results);
-                },
-                onFailure: (f) =>
-                {
-                    RGDebug.LogWarning($"Failed to evaluate CV Image criteria: {f}");
-                    onFailure.Invoke();
-                }
-            );
+            if (RGServiceManager.GetInstance().LoadAuth(out var authToken))
```
vontell, line 56: If they are not authed, what happens? It seems like neither action gets called which could cause hanging somewhere? Maybe we should either throw an exception or we could call onFailure. But at a minimum I'd like the user to at least see a message somewhere that they can't use the AI API without being logged in.

vontell (changes requested):
Nice! Just a small note about error messages for the case of not being logged in.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/AIServiceManager.cs
RG-nAmKcAz: LoadAuth internally logs a warning if the APIKey isn't configured, same as it does for the existing RGService calls.  If an invalid key is configured, you will get 401 errors from the AIserver calls.

vontell (approved):

PR #280, opened 2024-09-11
[REG-1922] Add base tests for the new Bot Sequences UI

addisonbgross:
image

## What has been done
- Added tests to exercise the public methods of the scripts composing the new Bot Sequences UI
- Added a couple of helper methods and reorganized a bit of code in the existing scripts, to aid in testing
- Added a couple helper methods to the RGTestUtils
- The mocking for this wasn't too painful, but if we are feeling spicy in the future we could use something more robust like [Moq](https://github.com/devlooped/moq)

## Out of Scope
- Exhaustive tests. There are some interactions that are tricky if not impossible to test using this Unit Testing-ish approach. Unity has a method of [using recorded gameplay to run UI tests](https://docs.unity3d.com/Packages/com.unity.automated-testing@0.8/manual/RecordedTesting.html), but I think it would be a cool idea to try this out with our own tools!

## Open Question
- Is this a good location in the project hierarchy for these tests?


src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceEditor.cs
```diff
@@ -152,14 +152,19 @@ public void CreateAvailableSegments(IList<BotSequenceEntry> segments)
...
         {
-            var childCount = AvailableSegmentsList.transform.childCount - 1;
-            for (var i = childCount; i >= 0; i--)
+            var childCount = AvailableSegmentsList.transform.childCount;
+            for (var i = 0; i < childCount; ++i)
             {
                 Destroy(AvailableSegmentsList.transform.GetChild(i).gameObject);
             }
+
+            return childCount;
```
addisonbgross, line 167: I don't know why I wrote this so strangely before 🤔 
it works just the same

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
```diff
@@ -57,7 +54,9 @@ public void SaveSequenceLoadedInEditor()
             }
             
             sequenceEditor.SetActive(false);
-            LoadSequences();
+            
+            var sequences = ResolveSequenceFiles();
+            InstantiateSequences(sequences);
         }
```
addisonbgross, line 60: Changed this to the `Start` method so that I could have more control of initialization during tests. The script functions the same when used for real

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
```diff
@@ -234,4 +201,31 @@ private IList<BotSequence> ResolveSequenceFiles()
...
+                }
+                catch (Exception exception)
+                {
+                    Debug.Log($"Error reading Bot Sequence {fileName}: {exception}");
+                    return (null, null);
+                }
+            })
+            .Where(s => s.Item2 != null)
+            .ToDictionary(s => s.Item1, s => s.Item2);
+    }
```
addisonbgross, line 230: This function is unchanged. I just moved it below the public functions

vontell (approved):
Appreciate all these tests! Excited to have more testing in our SDK, and this is a good set of examples for the rest of the team.

abeizer (approved):

PR #282, opened 2024-09-13
[REG-1904] Delete Sequence In Unity

addisonbgross:
image

## What has been done
- Added a confirmation dialog when deleting a Sequence
- Can delete Sequences!
- Added tests

## Out of scope
- Deleting files in a Runtime build. According to my research, this is not possible. It seems we can only unload files from memory, but not delete them in a build of a game

## LOOM
https://www.loom.com/share/0d55d279543d4c46adddf016fc50afc2


src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGDeleteSequence.cs
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+
+        /**
+         * <summary>
+         * Set the path and name of the Sequence the user is wanting to delete
+         * </summary>
+         * <param name="sequence">The Sequence Entry to delete</param>
+         */
+        public void Initialize(RGSequenceEntry sequence)
```
vontell, line 56: If we can't delete project-based sequences from a Runtime build, maybe we can have this dialog show this? i.e. if we can see that the sequence is part of the build, disable the confirm button and have text that says something like "Sequences created as part of the Unity Editor can only be deleted when running in the Editor"?

vontell (changes requested):
Just a suggestion/discussion on how we want to handle the attempt to delete editor-created sequences in runtime builds, let me know what you think and we can either come up with a plan or I can approve.

abeizer (changes requested):
The button colors should be red for confirming delete (destructive), and grey for cancel (no effect). We should save green for things like create/save/update

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGDeleteSequence.cs
RG-nAmKcAz: Ideally:

We should be able to grey out/disable the delete option for segments/sequences built into the project; when we are NOT in the editor.  Having a hover effect to explain that pre-built sequences/segments in the project can only be deleted in the Editor would give the user good feedback.

addisonbgross:
@vontell @RG-nAmKcAz disabled the Sequence delete buttons + added a tooltip when running in a build

https://github.com/user-attachments/assets/dd398b78-3411-4711-91d0-1dd743d0be3f

addisonbgross:
@abeizer changed to use grey for the cancel button, and red for the confirm button

https://github.com/user-attachments/assets/a3cedf5f-4d9e-4f53-9403-c868dd9100a9

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGDeleteSequence.cs
RG-nAmKcAz: @addisonbgross your change for disabling delete seems fine for this pass

in the future, we should consider whether we want to allow them to remove 'override' sequences in runtime builds

anything that isn't prebuilt into the project (anything loaded from the persistent data path) is deletable at runtime.. and in fact, may ONLY exist in that specific runtime and wouldn't be deletable from the editor.

that said.. do this as a follow up task because the changes i'm making for how runtime sequences are loaded will massively conflict with a change like this

vontell (approved):

vontell:
Will approve for now, but the reason I was thinking of showing it in the overlay is because there is chance we won't get hover tooltips on mobile devices, I should have added additional context in my original comment . But this should serve the purposes we have for now!

abeizer (approved):

PR #283, opened 2024-09-13
[Addison IF] Better sprite rendering

addisonbgross:
## What has been done
- Went through the icons used in our UI and optimized the following:
  - Anisotropic filtering levels: some of our pre-blurred art assets were being blurred further with more anisotropic filtering. Most of the icons should appear a bit clearer/crisper
  - Removed any mipmapping for sprite assets. Apparently Unity is no good at generating mipmaps for sprites used in the UI, so we're just using raw image scaling for differing screen sizes. Much less blurred 👍🏻 
  - Adjusting any pixel scaling. Some of our prefabs use the same assets at different scales. By adjusting these values we can make sure that edges are crisp at any size
  - Set most of our sprites to use the `full rect` mesh type. Long story short, using the default mesh type of `tight` can result in the warping of images due to Unity attempting to remove any extraneous transparent pixels from images. The `tight` setting is intended for performance improvements and since our UI is not a very intense situation, we can use the higher quality image setting of `full rect`
  
## Comparisons

Check out the crispness of the borders of things in these images. See that the **after** images and icons don't have the same degree of blurring or strange warping at the corners. Fantastic

### Before
Sequence List - Before
Sequence Editor - Before

### After
Sequence List - After
Sequence Editor - After

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Images/AddIcon.png.meta
+++ b/src/gg.regression.unity.bots/Runtime/Images/AddIcon.png.meta
@@ -33,7 +33,7 @@ TextureImporter:
   maxTextureSize: 2048
   textureSettings:
     serializedVersion: 2
-    filterMode: 2
+    filterMode: 0
     aniso: 1
     mipBias: 0
     wrapU: 1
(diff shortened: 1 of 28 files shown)


vontell (approved):

RG-nAmKcAz (approved):
While not blocking today, we should not avoid mipmaps altogether if we plan on supporting mobile.  These images we have are simple and scale well, but that requires loading that full image into the GPU.  In other cases (patterned textures) we will run into issues where the scaling causes artifacting without the mip maps

PR #286, opened 2024-09-16
[REG-1981] Implement the Object Detection threshold parameter in sdk.

batu:
[Loom video](https://www.loom.com/share/9775948da78742b5b444b8b2551d4a69?sid=105fafc7-1a8d-4256-a1b7-0b27553c738b).

This PR adds the threshold parameter to the sdk. This is an optional threshold to accept a returned match from the object detection model. Returned matches with a confidence score less than this threshold are ignored.

The added tests are as simple as they get. Currently only checking for errors at query construction, and that an error is logged when both text and image queries are added. I think we should add more comprehensive tests that check the actual result of sending a query. However, decided that this is out of scope for this PR and created a [ticket](https://linear.app/regression-games/issue/REG-1982/add-more-end-to-end-sdk-tests-for-computer-vision) for future work.

The a future PR will update the docs.

---


vontell (approved):
LGTM!

PR #293, opened 2024-09-17
[REG-1904] Edit Sequences in Unity

addisonbgross:
## What has been done
- Existing Sequences can be loaded into the Sequence Editor for updating
- Some small refactoring to how we manage Drop Zone states + instantiate prefabs
- Fixed the Sequence reload button. A mis-merge on my part resulted in a invalid action on the reload button
- Better resizing of Delete Sequence Dialog

## LOOM
https://www.loom.com/share/7452c3992944455ba5952b7c010edcfd


src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceEditor.cs
```diff
@@ -168,13 +206,24 @@ public int ClearAvailableSegments()
...
+            if (!string.IsNullOrEmpty(_existingSequencePath))
+            {
+                CurrentSequence.segments.Clear();
+
+                // If the Sequence being edited is renamed, we must delete the original Sequence
+                // first. Else we will end up with a duplicate
+                if (CurrentSequence.name != NameInput.text)
+                {
+                    BotSequence.DeleteSequenceAtPath(_existingSequencePath);
+                }
```
addisonbgross, line 223: A bit more info (beyond my Loom) on why the existing Sequence gets deleted before saving a new one, if the name changed. We generate the Sequence's file name based on the Sequence's name, so a changed name would result in a different file name. This is why I delete the original

src/gg.regression.unity.bots/Runtime/Prefabs/RGOverlay/DeleteSequenceDialog.prefab
```diff
@@ -738,19 +737,6 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   m_HorizontalFit: 0
   m_VerticalFit: 0
---- !u!114 &9089944729806318628
```
RG-nAmKcAz, line 741: Which behaviour is removed here ?

addisonbgross: I had a unrelated script attached `RGSaveSequenceButton` on the Delete button. Not useful

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceEditor.cs
```diff
@@ -168,13 +206,24 @@ public int ClearAvailableSegments()
...
-            var addedSegments = _dropZone.GetChildren();
+            if (!string.IsNullOrEmpty(_existingSequencePath))
+            {
+                CurrentSequence.segments.Clear();
+
+                // If the Sequence being edited is renamed, we must delete the original Sequence
+                // first. Else we will end up with a duplicate
+                if (CurrentSequence.name != NameInput.text)
+                {
+                    BotSequence.DeleteSequenceAtPath(_existingSequencePath);
```
RG-nAmKcAz, line 222: Instead of deleting and recreating... we should at least consider passing the correct path to `BotSequence.SaveSequenceAsJson` instead.  At least when we're editing, we already have the correct path to save to from loading the BotSequence.

I'm not saying we have to do this as it also has downsides in that the filename will become disconnected from the sequence name... but at least ponder its plusses and minuses

addisonbgross: Good feedback 👍🏻 
I share you concern about having file names and sequence names disconnected. Your point about a crash between the delete and create is valid, I'll make a Linear task to make this process more robust

addisonbgross: and I'll update to do the create first, delete second

addisonbgross: After updating the code, I'm not sure if a ticket to improve this is worth it. In the worst case, we save first and the game crashes before we can delete the original. The user will be left with a strange duplicate, but it seems pretty rare. What do you think Zack?

RG-nAmKcAz: that seems fine.. we didn't lose their work, they can manually remove the old un-deleted entry

RG-nAmKcAz (approved):

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGDropZone.cs
```diff
@@ -208,6 +208,25 @@ public bool IsEmpty()
             return _emptyStatePrefabInstance != null;
         }
 
+        /**
+         * <summary>
+         * Show or hide the empty state assigned to this drop zone
+         * </summary>
+         * <param name="isEmpty">If the empty state should be shown or hidden</param>
```
batu, line 215: Mega nit pick: The parameter name in conjunction with the param doc confused me, as it describes how the boolean is going to be acted on instead of what it is, I think? Maybe instead of isEmpty it also is called setEmpty but this is splitting hairs at this point, or maybe I misunderstood

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSaveSequenceButton.cs
```diff
@@ -1,16 +1,79 @@
...
+            
+            if (ButtonIcon == null)
+            {
+                Debug.LogError("RGSaveSequenceButton is missing its ButtonIcon");
+            }
+            
+            if (ButtonText == null)
+            {
+                Debug.LogError("RGSaveSequenceButton is missing its ButtonText");
+            }
```
batu, line 49: Nice early error handling.

batu (approved):
This looks good!

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGDropZone.cs
addisonbgross: Good call, a hair worth splitting. I was feeling a bit confused as well, I've changed it to `setEmpty` 👍🏻

PR #296, opened 2024-09-19
[REG-1984] Cleans up virtual mouse device... doesn't solve UI real mouse interference

RG-nAmKcAz:
- Disables the 'real' mouse device before we click and re-enables it after we next unclick
  - This may solve the issue where moving the real mouse during a click causes the click not to fire, but ONLY for behaviours listening for clicks.
  - This DOES NOT solve the issue for UI elements as the UI system is processing the raw input system mouse movement events irrespective of whether that device is enabled or not.  It also assumes a single mouse pointer device and thus the movements of the real mouse effectively update the position of the virtual mouse as the UI input system sees things. Poor design on unity's part imho

- Cleans up when our VirtualMouse exists to only be while a replay is playing.  Before this was active all the time and might have interfered with their actual game logic for devices.
---

### DIFF
--- a/src/RGUnityBots/Assets/Tests/Runtime/RGActionManagerTests.cs
+++ b/src/RGUnityBots/Assets/Tests/Runtime/RGActionManagerTests.cs
@@ -29,7 +29,7 @@ namespace Tests.Runtime
             Assert.IsTrue(parsedRange.RangeEquals(rng));
             Assert.IsTrue(rng.RangeEquals(parsedRange));
         }
-        
+
         [Test]
         public void TestValueRanges()
         {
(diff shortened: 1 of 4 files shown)


src/RGUnityBots/Assets/Tests/Runtime/RGActionManagerTests.cs
```diff
@@ -714,7 +715,7 @@ public IEnumerator TestActionManager()
                 }
                 RGActionManager.StopSession();
             }
-            
+
```
vontell, line 718: Lots of whitespace change, I wonder if we all have different formatters setup. We should message out what we have so the PRs don't have this going back and forth between two different formats.

vontell (approved):

src/RGUnityBots/Assets/Tests/Runtime/RGActionManagerTests.cs
RG-nAmKcAz: Remember ashley's whole thing about auto formatting that basically stopped development for a week while we had to go undo it.. let's not get there again over meaningless whitespaces.  I would suggest you 'hide whitespace' when reviewing in github.

abeizer (approved):

PR #297, opened 2024-09-19
[REG-2000] Move tests to RGUnityBots to avoid breaking customer builds

RG-nAmKcAz:
- Having tests in our core SDK module required customer projects to have NUnit included in their build dependencies/packages... This was breaking games like Match3 and others that did not include this

---


vontell (approved):
Will approve since I get the issue, but what we had before is what is [recommended by Unity itself](https://docs.unity3d.com/Manual/cus-tests.html) - I wonder if they have a different approach to avoiding those issues.

RG-nAmKcAz:
While I don't agree with Unity on this one, I do note that we did NOT follow that guide correctly in the first place.

I will test an alternate fix before merging this.

PR #303, opened 2024-09-19
[REG-1950] Better manage event system lifecycle + fix legacy 2D mouse clicks

RG-nAmKcAz:
- Updates our code for managing our legacy input overrides... tries to keep these out of the way unless we're actively playing a replay
- Now handles detecting pointer over objects correctly based on our overrides
- Fixes an issue where 2D mouse over object hits weren't being detected properly

- This PR DOES NOT make the Match3 fully work with recording / replay... there are many more unique-ness-es about the game that require using behaviour segments instead of recording.

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs
@@ -1,4 +1,4 @@
-﻿#if ENABLE_LEGACY_INPUT_MANAGER 
+﻿#if ENABLE_LEGACY_INPUT_MANAGER
 using System;
 using System.Collections;
 using System.Collections.Generic;
@@ -16,7 +16,7 @@ namespace RegressionGames.RGLegacyInputUtility
         AUTOMATIC,
         MANUAL
     }
-    
+
     class RGLegacyAxisInputState
     {
         public float rawValue;
@@ -39,19 +39,19 @@ namespace RegressionGames.RGLegacyInputUtility
     struct RGLegacySimulatedInputEvent
     {
         public RGLegacySimulatedInputEventType eventType;
-        
+
         // Key event
         public KeyCode keyCode;
         public bool isKeyPressed;
-        
+
         // Mouse movement event
         public Vector3 newMousePosition;
         public Vector3? newMouseDelta;
-        
+
         // Mouse scroll event
         public Vector2 newMouseScrollDelta;
     }
-    
+
     /**
      * Class that provides APIs for simulating device inputs, and also
      * provides the necessary wrappers for the UnityEngine.Input APIs that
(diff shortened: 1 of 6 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs
```diff
@@ -320,64 +329,85 @@ public static void Update()
...
                     }
-                        
+
                     // 2D raycast
                     {
                         Vector3 mouseWorldPt = camera.ScreenToWorldPoint(new Vector3(_mousePosition.x, _mousePosition.y, camera.nearClipPlane));
-                        RaycastHit2D hit2D = Physics2D.Raycast(mouseWorldPt, Vector2.zero, distance: Mathf.Infinity, 
-                            layerMask: cameraRaycastMask);
-                        if (hit2D.collider != null)
+                        var collider = Physics2D.OverlapPoint(mouseWorldPt, cameraRaycastMask);
```
vontell, line 376: I'll make sure to do some extensive testing in my UI testing UAT to make sure this works as expected 👍

vontell (approved):
Thank you for these many bug fixes, good catch! As mentioned in our conversation in Slack, I'll make a ticket in the backlog for us to consider making a suite of automated game tests. I actually don't think it would be too difficult, just some busy work...

src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs
RG-nAmKcAz: Reminder, this only affects Legacy InputManager games

addisonbgross (approved):

src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs
```diff
@@ -320,64 +329,85 @@ public static void Update()
...
                     {
                         continue;
                     }
-                        
-                    int cameraRaycastMask = camera.cullingMask & camera.eventMask;
-                        
+
+                    var cameraRaycastMask = camera.cullingMask & camera.eventMask;
+
+                    (GameObject, GameObject) result = (null, null);
```
batu, line 362: Flagging a github warning here

batu (approved):

src/gg.regression.unity.bots/Runtime/Scripts/RGLegacyInputUtility/RGLegacyInputWrapper.cs
batu, line 362: flagging a `The variable 'result' is assigned but its value is never used` warning here.

PR #304, opened 2024-09-20
[REG-1952] Scrollable Segment list for Sequences

addisonbgross:
## What has been done
- A Sequence's list of Segments is now scrollable
  - The scroll bar will be shown when the Segment list grows beyond its container's height
  - When dragging/reordering a Segment within the list, and the scroll bar is showing, we can scroll up or down by placing the cursor near the top or bottom of the list
- Better detection of cards leaving the drop zone. Removing Segments from a Sequence is smoother now
- Added a utility class for the Overlay tests

## LOOM
https://www.loom.com/share/47256eaf0cd64a70879d261c61493076


RG-nAmKcAz (approved):

batu (approved):
Looks great! Feature request: add the sound effect that we hear in the video 1:18 to everythong.

PR #305, opened 2024-09-20
[REG-2013] Record an END segment with any remaining inputs, logs, etc

RG-nAmKcAz:
- Captures any outstanding logs/inputs/etc in a final state capture on end recording
  - Only works on STOP recording from overlay button... if the application exits, we get notified too late by unity to be able to run a coroutine or wait for another frame :/ .  We still work in the case of application exit, we just don't get this extra data.
- Excludes any mouse events since the last button down IF this was a stop from our overlay button... this avoids replaying this sequence causing it to start recording again by clicking that button at the end
```json
{
"sessionId":"907c0645f3724b0b9454badd067b0fc2",
"referenceSessionId":null,
"apiVersion":13,
"tickNumber":2,
**"keyFrame":["END_RECORDING"],**
"time":7.7670521,
"timeScale":1,
"screenSize":{"x":1502,"y":796},
"performance":{
"previousTickTime":5.5324536,
"apiVersion":4,
"framesSincePreviousTick":260,
"fps":116,
```

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
@@ -130,9 +130,9 @@ namespace RegressionGames.StateRecorder
             _this._loggingObserver = GetComponent<LoggingObserver>();
         }
 
-        private void OnDestroy()
+        private void OnApplicationQuit()
         {
-            StopRecording();
+            StopRecordingNoCoroutine();
         }
 
         private void Start()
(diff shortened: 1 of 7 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -490,11 +493,45 @@ public void StopRecording()
...
+         */
+        private void StopRecordingNoCoroutine()
+        {
+            var wasRecording = IsRecording;
+            IsRecording = false;
+            StopRecordingCleanupHelper(wasRecording);
+        }
+
+        /**
+         * When triggered from the toolbar button, we need to exclude the last mouse click events so we don't capture the click on our own interface buttons
```
vontell, line 523: Good catch

vontell (approved):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -490,11 +493,45 @@ public void StopRecording()
...
+         */
+        public void StopRecording(bool toolbarButtonTriggered = false)
```
batu, line 525: very meticulous

batu (approved):

PR #310, opened 2024-09-24
[REG-1991] Resolve missing SDK fonts

addisonbgross:
image

## What is the problem
- A client is trying to use our SDK in their game, but our fonts are all messed up. The fonts are missing references to shaders

## What has been done
- Added TextMeshPro directly into our SDK (3.2 MB). This ensures that any shader/material references are correct, and that client's don't need to install TextMeshPro into their project (and hope that it doesn't clobber their existing fonts setup) in order to work with our SDK. Our Regression font now references the local TextMeshPro files

## Slack convo
https://regressiongames.slack.com/archives/C07M4T84ZNH/p1726597377076829

## LOOM
https://www.loom.com/share/e9253cf53bca4b40ba1f536c0da69cfd


batu (approved):
Great work! 3mb seems like a good tradeoff for this. 

I had some questions about distributing TMP, but it seems like this is fair game, according to [a self-proclaimed employee at Unity](https://www.reddit.com/r/unity/comments/g82cdz/textmesh_pro_license_question/).

RG-nAmKcAz (changes requested):
I want to think on this, at first glance, this feels like an anti pattern that could potentially cause more problems than it solves by introducing duplicate copies of textmeshpro into customer projects.

RG-nAmKcAz:
@addisonbgross I think you linked this to the wrong issue (1992 instead of 1991).  I realized this after writing up a big comment pondering how git-lfs had anything to do with font issues... then realizing something was wrong and checking your full task list.
 
That aside, since our SDK pre-reqs textmeshpro (correct pattern for dependency inclusion) the fonts should resolve.  Given that they aren't working.. in Lieu of including all of TMP in our build and potentially causing other dependency conflict issues, can we stick to pre-reqing textmeshpro, but just include our own fonts/materials/etc that we need in our sdk ?

vontell:
Just wanted to follow up with a similar comment regarding including this - from our conversation in our Linear ticket, I thought our solution was to provide a fallback font if they don't have TMP installed? I also would be a little worried about including TMP in our project since that could cause conflicts in the future.

addisonbgross:
@vontell turns out that the pink font wasn't being caused by a missing fallback font. Sergey showed me a screenshot and I noticed that materials were missing from our text/font. I'm going to try and just extract the pieces we need from TMP, to avoid including it all

abeizer:
For clarification, did they receive the prompt to add TMP assets to their game when they installed our SDK? If it will pop-up consistently, and our docs state you need the TMP assets, then I second Zack's comment. If they didn't see that pop-up then imo that's another issue

addisonbgross:
@abeizer 
I assume they've used the prompt, as I tested out another approach (unsuccessfully) that would require it. An issue that's happening with installing those assets into their game, is that it clobbered the existing fonts they had set up. Our SDK fonts were correct, but all of the fonts in their game had missing assets

addisonbgross:
@vontell @RG-nAmKcAz @abeizer 
I have the following updates:
- I removed all of the TMP stuff, aside from is minimally required for our fonts to function
  - The key to preventing the TMP "install the essential assets" popup from appearing was to ensure that the `TMP_Settings` file was in place. This way if a user has TMP in their project already, they can import those essentials without mixing anything up with our fonts
- I updated both versions of the `RGOverlayCanvas` to use our font exclusively. There were some refs to materials and shaders that were outside of our package. I'm not sure why this resulted in such a large drop in the size of the prefabs 🤔 
- I tested out these changes with Sergey, and things are working for him

RG-nAmKcAz (approved):

RG-nAmKcAz:
@addisonbgross Please test this with bossroom and PW starsreach before merging

vontell:
Nice work! Glad you tried this out with Sergey, ditto to Zack's comment, I am approving assuming that works.

vontell (approved):

abeizer (approved):

addisonbgross:
Here's some screenshots from Stars Reach, Steel Line and RG Boss Room. You can see that old Tanks! game in my original Loom video.

In-game text is present, and our overlay text is present as well 👍🏻 

image (4)
image (5)
image (6)
image (7)
image

addisonbgross:
Just a note that the last 3 commits were me undoing my changes to the overlay prefabs, then re-applying them after a merge from the `main` branch

PR #315, opened 2024-09-27
Zack/reg 1999

RG-nAmKcAz:
INSERT_PR_DETAILS_HERE

---


batu:
This is a 3-tickets-in-one PR! This also closes: https://linear.app/regression-games/issue/REG-1997/ability-to-duplicate-sequences

src/gg.regression.unity.bots/Editor/Scripts/BotSegments/IRGBotSequencesFinder.cs
```diff
@@ -58,21 +58,21 @@ private static void CreateBotSegmentsAsset(object o)
...
                     catch (Exception)
                     {
                         try
                         {
-                            var segmentList = JsonConvert.DeserializeObject<BotSegmentList>(fileContents);
+                            var segmentList = JsonConvert.DeserializeObject<BotSegmentList>(fileContents, JsonUtils.JsonSerializerSettings);
                             botSequences.segmentLists.Add(outputPath);
                         }
-                        catch (Exception)
+                        catch (Exception ex)
```
batu, line 71: Flagging github warnings.

src/gg.regression.unity.bots/Editor/Scripts/BotSegments/IRGBotSequencesFinder.cs
```diff
@@ -58,21 +58,21 @@ private static void CreateBotSegmentsAsset(object o)
...
                         {
                             try
                             {
-                                var segment = JsonConvert.DeserializeObject<BotSegment>(fileContents);
+                                var segment = JsonConvert.DeserializeObject<BotSegment>(fileContents, JsonUtils.JsonSerializerSettings);
                                 botSequences.segments.Add(outputPath);
                             }
                             catch (Exception)
```
batu, line 78: (nit) It would be nice if we could tie `AssetDatabase.Refresh();` after copying.

However, I saw that we already do a refresh just below here, and I couldn't find a better place in the context of this PR. Typing this in case there is another possible refresh point that I missed.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/JsonConverterContractResolver.cs
```diff
@@ -57,10 +61,30 @@ public class JsonConverterContractResolver : DefaultContractResolver
             { typeof(MeshRenderer), new MeshRendererJsonConverter() },
             { typeof(SkinnedMeshRenderer), new SkinnedMeshRendererJsonConverter() },
             { typeof(NavMeshAgent), new NavMeshAgentJsonConverter() },
-            { typeof(RGGameMetadataJsonConverter), new RGGameMetadataJsonConverter() }
 
-        };
+            { typeof(BehaviourActionData), new BehaviourActionDataJsonConverter() },
+            { typeof(BotAction), new BotActionJsonConverter() },
+            { typeof(CVImageBinaryData), new CVImageBinaryDataJsonConverter() },
```
batu, line 67: What does this block of code do? Is this a bug fix given a lot these are not related to copying and we are just adding them?

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/JsonUtils.cs
```diff
@@ -165,5 +165,26 @@ public static void WriteBehaviourStateToStringBuilder(StringBuilder stringBuilde
...
+
+        public static readonly JsonSerializerSettings JsonSerializerSettings = new()
+        {
+            Formatting = Formatting.None,
+            ReferenceLoopHandling = ReferenceLoopHandling.Ignore,
+            ContractResolver = JsonConverterContractResolver.Instance,
+            Error = delegate(object _, ErrorEventArgs args)
+            {
+                // just eat certain errors
+                if (args.ErrorContext.Error is MissingComponentException || args.ErrorContext.Error.InnerException is UnityException or NotSupportedException or MissingComponentException)
```
batu, line 177: What is the implication of eating NotSupportedException? Is this for specific settings of serialization or actually reading/writing?

batu (approved):
The code looks good. Just added a few questions for my own comprehension.

batu:
I am not a fan of the name `Generated_Recording`.

I think the fact that it is the 'Latest' recording is more important to communicate to the user in the name, rather than 'Generated'.

When I see 'generated' it immediately makes me ask more questions: Generated from what, when, why etc. and I think it is better if we avoid moments of confusion. Writing that it is auto generated in the description, and the fact that it is there without any human intervention showing that it is autogenerated I think is sufficient to convey the 'generated' aspect of this future.

Seperately, I would prefer using whitespace over "_" given that we support it.

batu:
I like the idea of sorting the latest recording to the top, but I agree that it is not important for this PR.

Thoughts on adding the recording button to all sequences that only contain recordings?

abeizer (approved):
RE: your loom -> I am strongly opinionated about sorting the generated recording to the front of the sequences list as a follow-up. The position should be predictable since recording will always create/update this specific entry

abeizer:
> I am not a fan of the name `Generated_Recording`, imho it should be something like `Latest Recording`.
> 
> I think the fact that it is the 'Latest' recording is more important to communicate to the user in the name, rather than 'Generated'.
> 
> When I see 'generated' it immediately makes me ask more questions: Generated from what, when, why etc. and I think it is better if we avoid moments of confusion. Writing that it is auto generated in the description, and the fact that it is there without any human intervention showing that it is autogenerated I think is sufficient to convey the 'generated' aspect of this future.
> 
> Latest conveys much more information about the sequence's contents.
> 
> Seperately, I would prefer using whitespace over "_" given that we support it.

I second this. "Latest Recording" is more meaningful to me than the fact that it was "generated" for me.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/JsonUtils.cs
RG-nAmKcAz: These are here and have been for a while (Not new to this PR), just moved/renamed code.

They handle converting unity Objects to json, specifically/normally around monobehaviours or other objects we don't have explicit converters for.  When reading all fields of game objects and behaviours in unity, many of them throw exceptions trying to access the field as they aren't meant to be read in user space, or not on the render thread, or other unity specific conditions.

This block prevents those conditions from stopping the other fields on the object being written

src/gg.regression.unity.bots/Editor/Scripts/BotSegments/IRGBotSequencesFinder.cs
RG-nAmKcAz: Copying a sequence happens while the game is running and thus AssetDatabase.Refresh() would only work as an API when running in the editor.  Even in the Editor though, there isn't a need to update the database because we aren't loading files from it.

In the editor runtime, we load the files as files never looking at the generated asset from the code above.

In the production runtime;  we load the asset as a 'resource' as well as loading the other in game segment files as 'resources' and overrides from the filesystem as files.

Ultimately this sequence file is only updated at compile time because it only impacts non-editor runtimes.

And.. in those runtimes, copying a sequence cannot update game binaries (writes to disk) so there wouldn't be a need to update this asset file even if it were possible there.

RG-nAmKcAz:
Latest recording is now renamed and always sorted to the front


image

src/gg.regression.unity.bots/Editor/Scripts/BotSegments/IRGBotSequencesFinder.cs
batu: Thank you for the explanation!

PR #319, opened 2024-10-01
[REG-2061] When creating a bot sequence from the Latest Recording sort the segments numerically. 

batu:
[Loom Video](https://www.loom.com/share/1fe8e487f74e4a78abf47a8cc374563f?sid=dd98d33b-a5b0-4f45-b952-76110476a312)

Currently when we are creating a BotSequence from the copied bot segments the segments are sorted alphanumerically. This means that 2.json will come after 19.json, breaking the sequence.

This PR hotfixes it by ordering it numerically. 

We should do the same sorting order wherever we are showing segments.

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
@@ -295,7 +295,7 @@ namespace RegressionGames.StateRecorder
                 .Select(a=>a.Substring(a.LastIndexOf('/')+1));
 
             // Order numerically instead of alphanumerically to ensure 2.json is before 10.json.
-            segmentFiles = BotSegmentDirectoryParser.OrderJsonFiles(segmentFiles);
+            segmentFiles = BotSegmentDirectoryParser.OrderJsonFiles(segmentFiles).ToList();
             
             string segmentResourceDirectory = null;
             string sequenceJsonPath = null;
@@ -339,6 +339,7 @@ namespace RegressionGames.StateRecorder
             {
                 path = segmentResourceDirectory + "/" + a
             }).ToList();
+
 
             var botSequence = new BotSequence()
             {
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
@@ -289,13 +289,13 @@ namespace RegressionGames.StateRecorder
         private async Task MoveSegmentsToProject(string botSegmentsDirectoryPrefix)
         {
             // get all the file paths normalized to /
-            var segmentFiles = Directory.EnumerateFiles(botSegmentsDirectoryPrefix)
+            var segmentFiles = Directory.GetFiles(botSegmentsDirectoryPrefix)
                 .Where(a=>a.EndsWith(".json"))
                 .Select(a=>a.Replace('\\','/'))
                 .Select(a=>a.Substring(a.LastIndexOf('/')+1));
 
             // Order numerically instead of alphanumerically to ensure 2.json is before 10.json.
-            segmentFiles = BotSegmentDirectoryParser.OrderJsonFiles(segmentFiles).ToList();
+            segmentFiles = BotSegmentDirectoryParser.OrderJsonFiles(segmentFiles);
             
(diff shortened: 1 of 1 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -278,7 +278,12 @@ await CreateAndUploadGameplaySession(
         private async Task MoveSegmentsToProject(string botSegmentsDirectoryPrefix)
         {
             // get all the file paths normalized to /
-            var segmentFiles = Directory.EnumerateFiles(botSegmentsDirectoryPrefix).Where(a=>a.EndsWith(".json")).Select(a=>a.Replace('\\','/')).Select(a=>a.Substring(a.LastIndexOf('/')+1));
+            var segmentFiles = Directory.EnumerateFiles(botSegmentsDirectoryPrefix)
+                .Where(a=>a.EndsWith(".json"))
```
RG-nAmKcAz, line 282: Use a call to `BotSegmentDirectoryParser.OrderJsonFiles` instead of this new code block

RG-nAmKcAz (changes requested):

batu:
Changed to using `BotSegmentDirectoryParser`.

RG-nAmKcAz (approved):

vontell (approved):

batu:
@RG-nAmKcAz @vontell can you please take another look? Some other change affected this PR, and in order to get it to work I had to add `.ToList()` right after sorting.

Here is a [loom video](https://www.loom.com/share/f5e784880f774757a59e9816dcd749d0) showing the behavior.

batu:
I don't want to merge this before understanding what is going on, but my testing shows that it works as intended now. So feel free to merge if you think this looks good

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -288,8 +289,14 @@ await CreateAndUploadGameplaySession(
...
             // get all the file paths normalized to /
-            var segmentFiles = Directory.EnumerateFiles(botSegmentsDirectoryPrefix).Where(a=>a.EndsWith(".json")).Select(a=>a.Replace('\\','/')).Select(a=>a.Substring(a.LastIndexOf('/')+1));
-
+            var segmentFiles = Directory.EnumerateFiles(botSegmentsDirectoryPrefix)
+                .Where(a=>a.EndsWith(".json"))
+                .Select(a=>a.Replace('\\','/'))
+                .Select(a=>a.Substring(a.LastIndexOf('/')+1));
+
+            // Order numerically instead of alphanumerically to ensure 2.json is before 10.json.
+            segmentFiles = BotSegmentDirectoryParser.OrderJsonFiles(segmentFiles).ToList();
```
RG-nAmKcAz, line 298: calling .ToList() would make no difference in ordering and adds a layer of inefficiency, marking as requested changes until there is an explanation on why it wasn't working without this

RG-nAmKcAz (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
vontell: I just spent a good few minutes looking at this - to me it smelled like something to do with the fact that these LINQ functions defer execution until the elements are actually needed, but I couldn't find anything glaring in the code (for instance, we actually force the execution of the LINQ functions inside of OrderJsonFiles because of some print statements that print the elements coming in, so they are already being forced).

From the video, the strange part is that the logs just stop.... I would double check if an exception is being thrown somehow and if it's being suppressed by some surrounding try catch.

RG-nAmKcAz: Linq functions are also notorious for eating exceptions

batu:
@RG-nAmKcAz PTAL. We understand the problem and the fix now. 

Because of lazy evaluation we would evaluate `Directory.EnumerateFiles(botSegmentsDirectoryPrefix)` AFTER the move operation, which would invalidate the path. 

Changed `Directory.EnumerateFiles` to `Directory.GetFiles` for instant execution of loading files. Added a comment

RG-nAmKcAz (approved):

PR #324, opened 2024-10-02
[REG-1951] Segments List + Play Segment in the SDK UI

addisonbgross:
https://github.com/user-attachments/assets/8c1111a6-8fa8-4aff-a400-356b7f070da8

## What has been done
- Users can view a list of Segments and play individual ones
- Users can switch tabs between the list of Sequences and the list of Segments
- Better Tooltip anchor positioning
- Created a `SegmentCardLarge` that is similar to the `SequenceCard`. This provides a view of a Segment that is better suited for interaction and displaying more content than the existing `SegmentCard`  (as seen in the Sequence Editor)
- Additional testing

## LOOM
https://www.loom.com/share/e53230f066a547de8e5d514927d5b236

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
@@ -14,10 +14,8 @@ public class RGSegmentEntry : MonoBehaviour
     public string segmentName;
 
     public string description;
-
-    /**
-     * <summary>Null if this is a resource load, or path if this is a file.</summary>
-     */
+    
+    // will be a file path when in-editor, or a resources path when in a packaged build
     public string path;
     
     /**
@@ -82,46 +80,41 @@ public class RGSegmentEntry : MonoBehaviour
      * When the play button is clicked, start the Segment and close the RGOverlay
      * </summary>
      */
-    public void OnPlay()
+    private void OnPlay()
     {
         var botManager = RGBotManager.GetInstance();
-        if (botManager != null)
+        if (botManager == null)
         {
-            var toolbarManager = FindObjectOfType<ReplayToolbarManager>();
-            toolbarManager.selectedReplayFilePath = null;
-            
-            botManager.OnBeginPlaying();
+            Debug.LogError("RGSegmentEntry cannot find the RGBotManager in its OnPlay function");
+            return;
+        }
+        
+        var toolbarManager = FindObjectOfType<ReplayToolbarManager>();
+        if (toolbarManager == null)
+        {
+            Debug.LogError("RGSegmentEntry cannot find the ReplayToolbarManager in its OnPlay function");
+            return;
+        }
+        toolbarManager.selectedReplayFilePath = null;
+        
+        botManager.OnBeginPlaying();
 
-            var segmentList = new List<BotSegmentList>();
-            segmentList.Add(CreateBotSegmentListForPath(path, out var sessId));
-            var sessionId = sessId;
-            sessionId ??= Guid.NewGuid().ToString();
-            
-            var playbackController = FindObjectOfType<BotSegmentsPlaybackController>();
-            playbackController.Stop();
-            playbackController.Reset();
-            playbackController.SetDataContainer(new BotSegmentsPlaybackContainer(segmentList.SelectMany(a => a.segments), sessionId));
-            playbackController.Play();
-        }
-    }
-
-    private BotSegmentList CreateBotSegmentListForPath(string path, out string sessionId)
-    {
-        var result = BotSequence.LoadBotSegmentOrBotSegmentListFromPath(path);
-        if (result.Item3 is BotSegmentList bsl)
+        var segmentList = new List<BotSegmentList>
         {
-            sessionId = bsl.segments.FirstOrDefault(a => !string.IsNullOrEmpty(a.sessionId))?.sessionId;
-            return bsl;
-        }
-        else
+            BotSequence.CreateBotSegmentListForPath(path, out var sessId)
+        };
+        var sessionId = sessId ?? Guid.NewGuid().ToString();
+        
+        var playbackController = FindObjectOfType<BotSegmentsPlaybackController>();
+        if (playbackController != null)
         {
-            var segment = (BotSegment)result.Item3;
-            sessionId = segment.sessionId;
-            var segmentList = new BotSegmentList(path, new List<BotSegment> { segment });
-            segmentList.name = segment.name;
-            segmentList.description = segment.description;
-            segmentList.FixupNames();
-            return segmentList;
+            Debug.LogError("RGSegmentEntry cannot find the BotSegmentsPlaybackController in its OnPlay function");
+            return;
         }
+        
+        playbackController.Stop();
+        playbackController.Reset();
+        playbackController.SetDataContainer(new BotSegmentsPlaybackContainer(segmentList.SelectMany(a => a.segments), sessionId));
+        playbackController.Play();
     }
 }
--- a/src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
@@ -106,7 +106,7 @@ public class RGSegmentEntry : MonoBehaviour
         var sessionId = sessId ?? Guid.NewGuid().ToString();
         
         var playbackController = FindObjectOfType<BotSegmentsPlaybackController>();
-        if (playbackController != null)
+        if (playbackController == null)
         {
             Debug.LogError("RGSegmentEntry cannot find the BotSegmentsPlaybackController in its OnPlay function");
             return;
--- a/src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
@@ -90,11 +90,8 @@ public class RGSegmentEntry : MonoBehaviour
         var botManager = RGBotManager.GetInstance();
         botManager.OnBeginPlaying();
 
-        // create the list of segments to play (in our case there will only be 1)
-        var segmentList = new List<BotSegmentList>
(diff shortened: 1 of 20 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            BotSequence.CreateBotSegmentListForPath(path, out var sessId)
+        };
+        var sessionId = sessId ?? Guid.NewGuid().ToString();
+        
+        var playbackController = FindObjectOfType<BotSegmentsPlaybackController>();
+        if (playbackController == null)
+        {
+            Debug.LogError("RGSegmentEntry cannot find the BotSegmentsPlaybackController in its OnPlay function");
+            return;
```
addisonbgross, line 143: This function paraphrases how the `BotSequence` class plays a Sequence. I have this feeling that this logic would be good somewhere else.

I'm thinking that playing a Segment should be a static method in the `BotSegment` class. Thoughts and feelings?

RG-nAmKcAz:
I still feel like these uis should be a searchable list/table instead of tiles, but reviewing this as it is currently

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
```diff
@@ -0,0 +1,152 @@
...
+ * </summary>
+ */
+public class RGSegmentEntry : MonoBehaviour
+{
+    public string segmentName;
+
+    public string description;
+    
+    // will be a file path when in-editor, or a resources path when in a packaged build
+    public string path;
```
RG-nAmKcAz, line 24: keep this a 'filePath' and 'resourcePath' like the sequences are

we will need that information later for properly indicating to users which segments are loading from a local override vs in the build

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * <summary>
+     * Ensures all needed prefabs and fields are properly set. Shows an icon if this segment is of the SegmentList type
+     * </summary>
+     */
+    public void Start()
+    {
+        if (string.IsNullOrEmpty(path))
+        {
+            Debug.LogError($"RGSegmentEntry is missing its path");
```
RG-nAmKcAz, line 52: (non-blocking) At some point we should do away with these checks and error log messages.  These are SDK breaking conditions that should really throw exceptions, not just log a message and continue

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
```diff
@@ -0,0 +1,152 @@
...
+     * <summary>
+     * When the play button is clicked, start running the Segment and close the RGOverlay
+     * </summary>
+     */
+    private void OnPlay()
+    {
+        var botManager = RGBotManager.GetInstance();
+        if (botManager == null)
+        {
+            Debug.LogError("RGSegmentEntry cannot find the RGBotManager in its OnPlay function");
```
RG-nAmKcAz, line 117: In no (non-unit test) situation should this ever happen :/ 

I would rather this throw the exception for the null botManager later in the code so this doesn't pass our dev testing with us possibly missing one of these error messages.

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            Debug.LogError("RGSegmentEntry cannot find the ReplayToolbarManager in its OnPlay function");
+            return;
+        }
+        toolbarManager.selectedReplayFilePath = null;
+        
+        // set the recording toolbar's state for playing this segment
+        botManager.OnBeginPlaying();
+
+        // create the list of segments to play (in our case there will only be 1)
```
RG-nAmKcAz, line 132: (comment wording issue) will there really only be 1 if the path is a segmentList ?

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
```diff
@@ -0,0 +1,152 @@
...
+        var segmentList = new List<BotSegmentList>
```
RG-nAmKcAz, line 133:
```suggestion
        var segmentList = BotSequence.CreateBotSegmentListForPath(path, out var sessId);

```

RG-nAmKcAz: This suggestion can't apply directly, but we don't need the extra new List allocation here

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
RG-nAmKcAz: Re:  // play the segment

Botsequence has this same logic flow in its `Play` method.

I would NOT suggest that you create a sequence for the segments just to re-use this. 

 **Given that playing a bot segment/list directly is only exposed from this UI, I feel like the code is perfect right where it is.**

If you are insistent on moving it somewhere else, it would go in BotSegmentList in a new `Play` method, to match the pattern on `BotSequence`.

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
```diff
@@ -52,15 +58,55 @@ public void LoadSequences()
...
     {
         DontDestroyOnLoad(gameObject);
     }
 
     public void Start()
     {
-        LoadSequences();
+        if (sequencesPanel == null)
+        {
+            Debug.LogError("RGSequenceManager is missing its sequencesPanel");
```
RG-nAmKcAz, line 96: again, these should be hard exceptions if this is null, our SDK is mis-configured/broken if this is happening

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGTabButton.cs
```diff
@@ -0,0 +1,189 @@
...
+        public void Start()
+        {
+            var background = GetComponent<Image>();
+            if (background != null)
+            {
+                _background = background;
+            }
+            else
+            {
+                Debug.LogError("RGTabButton is missing its background");
```
RG-nAmKcAz, line 54: last comment on this, but we shouldn't need all these checks / log statements for misconfigured game objects

src/gg.regression.unity.bots/Tests/Scripts/RGOverlay/RGSegmentEntry.test.cs
```diff
@@ -0,0 +1,68 @@
...
+            entry.descriptionComponent = RGTestUtils.CreateTMProPlaceholder();
+            entry.playButton = _uat.AddComponent<Button>();
+            
+            // create tooltip child
+            var tooltip = new GameObject();
+            tooltip.AddComponent<RGTooltip>();
+            tooltip.transform.SetParent(entry.transform);
+
+            // create segment list indicator w/Image
+            var segmentListIndicator = new GameObject();
```
RG-nAmKcAz, line 36: all uses of `new GameObject()` in these unit tests must follow the pattern...

```csharp
new GameObject(){
                transform =
                {
                    parent = parentObject.transform
                }
            };
```

where `parentObject` is something in the transform hierarchy of `_uat`.   


Otherwise, destroying `_uat` DOES NOT cleanup the other game objects  and leads to unit test interference between tests

RG-nAmKcAz (changes requested):

abeizer (approved):

vontell (approved):
Didn't have anything in addition to Zack's comments, I'll mark as approved assuming his remarks are addressed.

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
addisonbgross: good call. These 2 have been separated

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
addisonbgross: Fair point. I've removed any of these checks I've added in this PR from the relevant `Start` methods. I also removed these from the Sequence editor

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
addisonbgross: reworded. Good call

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
addisonbgross: nice one. Changed

src/gg.regression.unity.bots/Tests/Scripts/RGOverlay/RGSegmentEntry.test.cs
addisonbgross: ah I'm glad you noticed this. I've updated my added tests

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSegmentEntry.cs
addisonbgross: Your logic is sound, the code will stay put

RG-nAmKcAz (approved):

RG-nAmKcAz:
After testing this in the build, opened follow up task for dis-ambiguation 

https://linear.app/regression-games/issue/REG-2075/segment-cards-need-path-information

PR #326, opened 2024-10-03
[REG-2068] Handle ambiguous available segments in SequenceEditor UI by showing path

RG-nAmKcAz:
- Handles dis-ambiguating segments that have the same 'name' but different paths on the editor UI
  - If the segment has a name, then the name and path will be shown
  - If the segment DOES NOT have a name, then the path will be shown in place of the name, but the path sub text will be hidden
- Makes needed updates to the prefab to make decription/overlow expansion respect this new layout
- Stops setting the 'resourcePath' as the name on segments/segmentLists loaded from sequences and tracks the resourcePath separately.  This allows us to know if a user supplied name was truly set or not.

Long descriptions example with ...
image

Empty name example
image

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Fonts/Regression.asset
+++ b/src/gg.regression.unity.bots/Runtime/Fonts/Regression.asset
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:6b1db76e101060aa98c6c6d0eaa0b0190f830a94c558a914a25cd9cb03e7e2d9
-size 1081928
+oid sha256:fac3f5b1478c0e1f4a6cad4a1670e1cdb0d4e029e150ad382b063de6314582f5
+size 2185363
--- a/src/gg.regression.unity.bots/Runtime/Prefabs/RGOverlay/SegmentCard.prefab
+++ b/src/gg.regression.unity.bots/Runtime/Prefabs/RGOverlay/SegmentCard.prefab
@@ -35,8 +35,8 @@ RectTransform:
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
   m_AnchorMin: {x: 0, y: 1}
   m_AnchorMax: {x: 0, y: 1}
-  m_AnchoredPosition: {x: 121, y: -53}
-  m_SizeDelta: {x: 230, y: 50}
+  m_AnchoredPosition: {x: 123.5, y: -58}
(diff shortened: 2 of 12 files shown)


addisonbgross (approved):

batu:
This is an important feature but I think the overall aesthetics of our tool takes a step back. How difficult would it be do only show this extra information when it need disambiguation?

RG-nAmKcAz:
> This is an important feature but I think the overall aesthetics of our tool takes a step back. How difficult would it be do only show this extra information when it need disambiguation?

That is what is already happening.  The information is only shown when a name is set.  Without this shown, finding your segment on disk or in the project required going 1 by 1 opening files to read names.

If you're asking if we can do this only when there are duplicate names... sure we probably could, but that didn't help me when I wanted to find that segment on disk.

abeizer:
Functionally this is helpful, but aesthetically it makes things hard to visually parse.

Can we make the text smaller, possibly non-italic (if that looks good), and try a muted tone so that the name stands out the most and the path is there if the user wants to read it but it doesn't take center stage on the card? Not sure if you and @addisonbgross have talked about visual tweaks as a follow-up.

Also is there a way to see the entire path when it's sub-text? In your screenshot there's a bunch of `bot-segments/computer-vision/computer...` but if I can't see the end of the path then that won't always help me differentiate between segments with the same name.

RG-nAmKcAz:
I have already opened a follow up task as this is needed on Segments tab as well.
https://linear.app/regression-games/issue/REG-2075/segment-cards-need-path-information

Right now.. this is completely un-usable in my environment.  If we don't merge this PR that is fine, but we need some similar change here immediately to resolve this.

As for UI styling... I'm not going to argue this beyond saying, I already spent about an hour with various colors, intensities, font sizings and layout spacings to get it to where I personally could easily see it.  I realize that others may not agree with those styling or color choices.  To get this to please everyone will involve re-thinking how those cards are styled altogether to fit this in nicely whereas this was meant as a PR for ... its not usable, get it working now.

RG-nAmKcAz:
Here it is update without italics.
image

The ellipsis ... is mostly because our layout is too small compared to the screen space, I have a task open for next cycle to redo the Editor pane to takeup more of the screen space so we can widen the middle and right sections and show more details

https://linear.app/regression-games/issue/REG-2071/increase-the-dimensions-of-the-sequence-editor

RG-nAmKcAz:
Gray update

image

abeizer (approved):

PR #327, opened 2024-10-03
[REG-2063] Enable running unity with -rgsequencepath startup option

RG-nAmKcAz:
- Adds -rgsequencepath command line option
  - This allows you to start Unity , automatically run an RG bot sequence, and then exit the unity runtime
  - Also supports -rgsequencetimeout option to put a number of seconds limit on your runtime before it closes Unity and reports a failure
  - Supports RC=0 for successful run and RC=1 for timeout failure and RC=# for various arg parsing errors
- Fixes some cases for -batchmode where keyboard and mouse devices needed to handle being null, and/or needed a virtual keyboard device registered to function correctly 

Usage samples

- Run the game with the full on screen graphics
  - Supports ALL of our SDK features
  - `./com.unity.multiplayer.samples.coop.exe -rgsequencepath "Assets/RegressionGames/Resources/BotSequences/TypeADF.json" -rgsequencetimeout 7`

- Run the game without rendering
  - Does NOT support anything involving graphics including (screenshots, anything CV related)
  - `./com.unity.multiplayer.samples.coop.exe -batchmode -logFile C:/Users/zack/bossroom_build/headless.log -rgsequencepath "Assets/RegressionGames/Resources/BotSequences/TypeADF.json"`

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
@@ -21,7 +21,7 @@ namespace RegressionGames
         internal static readonly int Rc_OverlayNotInScene = 3;
         internal static readonly int Rc_SequenceLoadFailure = 4;
 
-        internal static readonly string CommandLineArgument = "-rgSequencePath";
+        internal static readonly string CommandLineArgument = "-rgsequencepath";
 
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.AfterSceneLoad)]
         public static void Initialize()
--- a/src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
@@ -1,9 +1,12 @@
 ﻿using System;
 using System.Collections;
+using System.Linq;
+using System.Reflection;
 using RegressionGames.StateRecorder.BotSegments;
 using RegressionGames.StateRecorder.BotSegments.Models;
 using RegressionGames.Types;
 using UnityEngine;
+using UnityEngine.InputSystem;
 using Object = UnityEngine.Object;
 
 // ReSharper disable InconsistentNaming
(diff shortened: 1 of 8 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
```diff
@@ -0,0 +1,255 @@
...
+        internal static readonly int Rc_SequenceTimeoutNotInt = 7;
+
+        internal static readonly string SequencePathArgument = "-rgsequencepath";
+        internal static readonly string SequenceTimeoutArgument = "-rgsequencetimeout";
+
+        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSplashScreen)]
+        public static void FirstLoadChecks()
+        {
+            // do this here to fail fast on bad args
+            ParsePathArgument();
```
batu, line 34: Good stuff.

src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
```diff
@@ -0,0 +1,255 @@
...
+        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.AfterSceneLoad)]
+        public static void Initialize()
+        {
+            var resourcePath = ParsePathArgument();
+            if (resourcePath != null)
+            {
+                BotSegmentsPlaybackController playbackController = Object.FindObjectOfType<BotSegmentsPlaybackController>();
+                if (playbackController == null)
+                {
+                    RGDebug.LogError($"RGOverlay must be in the first scene loaded for your game to utilize {SequencePathArgument} command line argument");
```
batu, line 47: Do we see this in the console when running in headless mode? If not, I think we should definitely make a note in docs to point people to the correct log file to debug sudden closure of their app

src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
```diff
@@ -0,0 +1,255 @@
...
+                    Application.Quit(Rc_OverlayNotInScene);
+                }
+
+                playbackController.gameObject.AddComponent<RGHeadlessSequenceRunnerBehaviour>();
+
+                //ensure a keyboard exists for the input playback
+                if (IsBatchMode())
+                {
+                    var currentKeyboard = Keyboard.current;
```
batu, line 56: If they already have a keyboard would we also need to attempt to set flags?

batu (approved):
This looks good! Added a few questions for my comprehension

src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
RG-nAmKcAz: We definitely do not today as we don't create a virtual keyboard for our runs outside of this headless environment.

It's a good philosophical discussion we should experiment with though in the future

src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
RG-nAmKcAz: In console mode you get no STDOUT from Unity.  Everything is put into the -logfile path you specify, or defaults to the player.log file if not running in -batchmode

That said, I'm not very interested in re-documenting how Unity works for users 👎 

We absolutely should/will document that this feature exists once we've had a cycle or two of testing on it, 
https://linear.app/regression-games/issue/REG-2074/document-rgsequencepath-command-line-options

vontell (approved):
Nice! No additional comments from my side. I wonder if we want to add some tests, although in this case I don't really mind since this was mostly just to make sure things can work in headless mode, I imagine the testing around the remote execution of a sequence will be time better spent.

PR #329, opened 2024-10-03
[REG-2064] Manually write files on OSX platforms

vontell:
I'm not very happy with this PR... I cannot determine if it fixes the original issue because while we were able to reproduce the bug on Wednesday afternoon, as of Thursday I was **not** able to reproduce the issue... Also, this works as expected when in the editor, but in the build, the sequence is not created correctly (the segments and sequence are not created). I have a feeling that the build issue is unrelated to this particular PR.

I am opening this as a draft and would like to chat with @batu and @addisonbgross about this perhaps?

UPDATE (Oct 8th, Mon afternoon) - I've now updated this to use the File.Copy function, but once again, I won't be able to determine if this fix works until we try it out with the user.

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
@@ -333,7 +333,7 @@ namespace RegressionGames.StateRecorder
                     {
                         var sourcePath = botSegmentsDirectoryPrefix + "/" + segmentFile;
                         var destinationPath = segmentResourceDirectory + "/" + segmentFile;
-                        await File.WriteAllBytesAsync(destinationPath, await File.ReadAllBytesAsync(sourcePath));
+                        File.Copy(sourcePath, destinationPath);
                     }
                     // Then delete the original directory
                     Directory.Delete(botSegmentsDirectoryPrefix, true);
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
@@ -332,7 +332,7 @@ namespace RegressionGames.StateRecorder
                 // If on a Mac, for some weird reason, .Move will fail for project paths that have spaces. We predict
                 // that downstream, something is happening where the space is not properly handled in Unix commands. 
                 // Unfortunately, we could only reproduce this once, but this change should fix it. On Mac, we instead
-                // read and write the files by hand, and avoid any move operations.
+                // use the File.Copy function.
                 if (Application.platform == RuntimePlatform.OSXEditor || Application.platform == RuntimePlatform.OSXPlayer)
                 {
                     Directory.CreateDirectory(segmentResourceDirectory); // Since we aren't moving, we need to create the directory
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
@@ -322,8 +322,28 @@ namespace RegressionGames.StateRecorder
                 // So we can safely create this without interfering with the .Move command right below.
                 Directory.CreateDirectory(Path.GetDirectoryName(segmentResourceDirectory));
 
-                // move the directory (this also deletes the source directory)
-                Directory.Move(botSegmentsDirectoryPrefix, segmentResourceDirectory);
+                // If on a Mac, for some weird reason, .Move will fail for project paths that have spaces. We predict
+                // that downstream, something is happening where the space is not properly handled in Unix commands. 
+                // Unfortunately, we could only reproduce this once, but this change should fix it. On Mac, we instead
+                // read and write the files by hand, and avoid any move operations.
+                if (Application.platform == RuntimePlatform.OSXEditor || Application.platform == RuntimePlatform.OSXPlayer)
+                {
+                    Directory.CreateDirectory(segmentResourceDirectory); // Since we aren't moving, we need to create the directory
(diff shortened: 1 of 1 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -322,8 +322,28 @@ private async Task MoveSegmentsToProject(string botSegmentsDirectoryPrefix)
                 // So we can safely create this without interfering with the .Move command right below.
                 Directory.CreateDirectory(Path.GetDirectoryName(segmentResourceDirectory));
 
-                // move the directory (this also deletes the source directory)
-                Directory.Move(botSegmentsDirectoryPrefix, segmentResourceDirectory);
+                // If on a Mac, for some weird reason, .Move will fail for project paths that have spaces. We predict
+                // that downstream, something is happening where the space is not properly handled in Unix commands. 
+                // Unfortunately, we could only reproduce this once, but this change should fix it. On Mac, we instead
+                // read and write the files by hand, and avoid any move operations.
+                if (Application.platform == RuntimePlatform.OSXEditor || Application.platform == RuntimePlatform.OSXPlayer)
```
RG-nAmKcAz, line 329: You should only need OSXPlayer as this code is run in the context of the game playing, not an editor script

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -322,8 +322,28 @@ private async Task MoveSegmentsToProject(string botSegmentsDirectoryPrefix)
...
+                {
+                    Directory.CreateDirectory(segmentResourceDirectory); // Since we aren't moving, we need to create the directory
+                    foreach (var segmentFile in segmentFiles)
+                    {
+                        var sourcePath = botSegmentsDirectoryPrefix + "/" + segmentFile;
+                        var destinationPath = segmentResourceDirectory + "/" + segmentFile;
+                        await File.WriteAllBytesAsync(destinationPath, await File.ReadAllBytesAsync(sourcePath));
```
RG-nAmKcAz, line 336: why not use `File.Copy` ?

RG-nAmKcAz:
@vontell .. In the production build, this copy operation doesn't happen; so if that's what you're seeing.  That is expected/intentional.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
vontell: Well that's where I get worried... if Directory.Move isn't working, would File.Copy itself not work? I went with what I thought was the safest option... I am fine with File.Copy but I have no idea if it suffers from the same issue. Technically this too might not be safe.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
vontell: Ah thank you

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
RG-nAmKcAz: I would definitely use File.Copy as it uses streams and copies small byte buffers at a time instead of reading the full file into memory.  Not that these files are huge, but it is better practice.

It sounds like your real concern is that the recreate is un-reliable.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
vontell: @RG-nAmKcAz After some testing, when running in the editor as the game is playing, the platform is indeed `RuntimePlatform.OSXEditor`, not `RuntimePlatform.OSXPlayer`, so I have to keep this here. It is OSXEditor even if not an editor script.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
vontell: Changed it to File.Copy, but yes my concern is not only that the recreate is unreliable, it's not possible at all 😆 I am making this change but I will not know if this fixes anything until we have the user try it out.

I tried to dig into the Mono source code, but as expected it eventually calls into some native libraries so I can't really inspect **why** Directory.Move wouldn't work, and how different it is from File.Copy. But anyways, this is at least different enough to try out.

vontell:
> @vontell .. In the production build, this copy operation doesn't happen; so if that's what you're seeing. That is expected/intentional.

Can you elaborate on this? Because this code here (line 309) does have a path to save the latest recording to the persistentDataPath if running in a production build - can that be removed? Since I saw that, I assumed that it was possible (and I can't seem to find the conditional that would stop this method from running in builds).

The bug I was seeing was actually something else I believe (I can't get sequences to be edited or saved in builds) so I'll move that convo to Slack / another ticket.

RG-nAmKcAz:
> > @vontell .. In the production build, this copy operation doesn't happen; so if that's what you're seeing. That is expected/intentional.
> 
> Can you elaborate on this? Because this code here (line 309) does have a path to save the latest recording to the persistentDataPath if running in a production build - can that be removed? Since I saw that, I assumed that it was possible (and I can't seem to find the conditional that would stop this method from running in builds).
> 
> The bug I was seeing was actually something else I believe (I can't get sequences to be edited or saved in builds) so I'll move that convo to Slack / another ticket.

It doesn't copy INTO the project.  It copies to the persistent data path.  That is also where bot sequences are saved from the editor in production builds.  IOW.  The issues you're seeing there are probably related.

RG-nAmKcAz (approved):

addisonbgross (approved):

PR #332, opened 2024-10-07
feat: add resource path to large segment cards

addisonbgross:
image

## What has been done
- Added a resource path text component to the large Segment Cards
  - This will show beneath the card's name field. If the name field is missing, we will use the resource path in its stead
- Added the resource path text to the dragging card prefab
- Increased the height of the potential drop spot prefab

## LOOM (It cuts before I'm done talking, but its ok)
https://www.loom.com/share/66adb820a05b4fbe88991f499813945a


addisonbgross:
@RG-nAmKcAz good thinking. I did a slight adjustment to both the `SequenceCard` and `SegmentCardLarge` prefabs. I also moved that red, recording dot on Sequence cards so that it's behind the card's name. I noticed that it could block the name text a bit

image
image

vontell (approved):

PR #333, opened 2024-10-09
[REG-2105] Allow absolute paths to be used in segment and sequence loading

vontell:
There was a bug where sequences and segments could not be loaded from the application data path in builds. I am not super familiar with all of the logic here so wanted to take a minute to explain my observations and decisions here:

* I was getting errors like this in **production builds**: `Error reading Bot Sequence /Users/aaronvontell/Library/Application Support/DefaultCompany/normal/RegressionGames/Resources/BotSequences/Latest_Recording.json: System.Exception: Invalid path.  Path must be relative, not absolute in order to support editor vs production runtimes interchangeably.`
* In the editor, paths look like `Assets/RegressionGames/Resources/BotSequences/asdasd.json`
* When in a build, the paths look like `/Users/aaronvontell/Library/Application Support/com.DefaultCompany.This--should-break-/RegressionGames/Resources/BotSequences/Latest_Recording.json`
* The code for loading these sequences had an explicit catch for JSONs coming from an absolute path, and wouldn't allow it.
* The fix I found was to remove that check, **but** this seemed weird... because we are basically taking the absolute path, trimming off most of it in the `ToResourcePath` function, and then reconstructing it... and the LoadJsonResource function knows how to handle both scenarios. 

An alternative to my change is to only call ToResourcePath inside of the `UNITY_EDITOR` scenarios and then use the path directly inside of the else block, so that the path doesn't need to be reconstructed. However, I wanted to check with @RG-nAmKcAz and @addisonbgross that this is still fine considering the "override" behavior, or if there is additional logic I need to consider.

At least on mac, I confirmed that this allows us to both create and see sequences that are in the application data path, and everything behaves as expected in builds and in editor. Note, however, that the sequences and segments inside the application data path do not show in the editor (which I believe is expected, but wanted to confirm).

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections;
+using System.IO;
 using System.Linq;
 using System.Reflection;
 using RegressionGames.StateRecorder.BotSegments;
@@ -124,6 +125,18 @@ namespace RegressionGames
                 {
                     if (i + 1 < argsLength)
                     {
+
+                        var path = args[i + 1];
+                        
+                        // Sequences that are run via -rgsequencepath must be relative, as we only allows sequences
+                        // included in the Resource or persistent path directories.
+                        if (Path.IsPathRooted(path))
+                        {
+                            RGDebug.LogError($"{SequencePathArgument} command line argument requires a relative path");
+                            Application.Quit(Rc_SequencePathMissing);
+                            return null;
+                        }
+                        
                         return args[i + 1];
                     }
                     // else
--- a/src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
@@ -24,6 +24,7 @@ namespace RegressionGames
         internal static readonly int Rc_SequenceTimeoutNeedsPath = 5;
         internal static readonly int Rc_SequenceTimeoutMissing = 6;
         internal static readonly int Rc_SequenceTimeoutNotInt = 7;
+        internal static readonly int Rc_SequencePathNotRelative = 8;
 
         internal static readonly string SequencePathArgument = "-rgsequencepath";
         internal static readonly string SequenceTimeoutArgument = "-rgsequencetimeout";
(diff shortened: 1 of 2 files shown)


RG-nAmKcAz:
@vontell On Windows, everything works without this change.  Can you include the full call stack for the error you're getting so we can see what codepath on mac is producing an unexpected path into that call.

Taking these checks out is not valid and will prevent sequences from loading properly in a few other scenarios that require relative paths.

RG-nAmKcAz (approved):
This works on Windows because the path is starting with C:\  not \ or /

Furthermore, looking into how this prompt was interacting with command line sequence runner options, I think this will be safe there as well and should support full or relative paths.

This check existed before I started auto normalizing everything to a resourcePath and then running it through the loading evaluator to make sure we always loaded the override copy, not accidentally something else.

addisonbgross (approved):

vontell:
I see! Just want to confirm this before I merge then since the last time I thought about this was last week:

* My current fix is all good
* It won't work for arbitrary paths for the new command line parameter, which is fine (and we should document that the command line option should only be used for sequences contained in the build / persistent data path).

If that is correct, please provide a thumbs up and I will merge!

RG-nAmKcAz:
@vontell I would add a check to the option loader asserting that it is a relative path for now.  Similar to the existing code, but maybe with an extra bit to save us from c: , e:, etc

Doesn't start with '/' or '\' and doesn't have a second char of ':'

vontell:
@RG-nAmKcAz will do 👍

vontell:
Apologies for the delay on this. @RG-nAmKcAz please see my latest commit for that change - confirmed that the sequence path works properly for relative paths and does print the error message in the logs when it is absolute (the Path utility I use here should properly handle both Windows disks and unix systems)

src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
```diff
@@ -124,7 +125,19 @@ internal static string ParsePathArgument()
...
-                        return args[i + 1];
+
+                        var path = args[i + 1];
+                        
+                        // Sequences that are run via -rgsequencepath must be relative, as we only allows sequences
+                        // included in the Resource or persistent path directories.
+                        if (Path.IsPathRooted(path))
+                        {
+                            RGDebug.LogError($"{SequencePathArgument} command line argument requires a relative path");
+                            Application.Quit(Rc_SequencePathMissing);
```
RG-nAmKcAz, line 136: This should be a unique RC

RG-nAmKcAz (approved):
Changes look good.  Just need a unique RC for that case

src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
vontell: Good catch, added

PR #334, opened 2024-10-10
[REG-2102] Support Remote WorkAssignments with -rgremoteworker

RG-nAmKcAz:
- Creates remote work execution patterns
- TODO: Test End to End function
- TODO: Update network API paths

---


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/JsonConverters/AnimatorJsonConverter.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace RegressionGames.StateRecorder.JsonConverters
 {
-    public class AnimatorJsonConverter : Newtonsoft.Json.JsonConverter, ITypedStringBuilderWriteable<Animator>, IBehaviourStringBuilderWritable
+    public class AnimatorJsonConverter : Newtonsoft.Json.JsonConverter, ITypedStringBuilderConverter<Animator>, IBehaviourStringBuilderWritable
```
RG-nAmKcAz, line 9: You can ignore everything below this file... just updates where I renamed the Interface to be more descriptive as I needed another interface to do exactly what this old name said it did....

src/gg.regression.unity.bots/Runtime/Scripts/RemoteOrchestration/RGRemoteWorkerAssignmentManager.cs
```diff
@@ -0,0 +1,364 @@
...
+using RegressionGames.StateRecorder.BotSegments.Models;
+using RegressionGames.StateRecorder.JsonConverters;
+using RegressionGames.Types;
+using UnityEngine;
+using Object = UnityEngine.Object;
+
+// ReSharper disable InconsistentNaming
+namespace RegressionGames.RemoteOrchestration
+{
+    public static class RGRemoteWorkerAssignmentManager
```
RG-nAmKcAz, line 19: This file is the meat/potatoes of the changes

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSequence.cs
```diff
@@ -35,6 +35,13 @@ public class BotSequence
          */
         public string description;
 
+        /**
+         * Not persisted or written to json, populated on load
+         */
+        public string resourcePath;
+
+        public static BotSequence ActiveBotSequence = null;
```
RG-nAmKcAz, line 43: The changes in these areas get a big ugly, but I tried to minimize the 'full re-write' that could have happened here.

We have code now to manage each of these possible playback cases and their 'status' on each entity in their chain as much as possible...

- Remote Work . which runs a botsequence , which uses the playback controller to run its segments
- Botsequence , which uses the playback controller to run its segments
- Zip, which uses the playback controller to run its segments
- Single Segment, which uses the playback controller to run its segment

So...
- the playbackcontroller can tell us if there is are active segments, but can't tell us if they are in a sequence
- a botsequence can now tell us if there is an active sequence, so we can give more details than just true/false segments are running
- a work assignment, can tell us its details, and also checks for sequence/segments active

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/IStringBuilderConverter.cs
```diff
@@ -0,0 +1,36 @@
+using System.Text;
+using JetBrains.Annotations;
+
+namespace RegressionGames.StateRecorder
+{
+    public interface IStringBuilderConverter
+    {
+
+        public void WriteToStringBuilderNullable(StringBuilder stringBuilder, [CanBeNull] object val)
```
RG-nAmKcAz, line 9: New interface that does what `IStringBuilderWriteable` used to do.  This is used for converters of json and is better named.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/IStringBuilderWriteable.cs
```diff
@@ -1,36 +1,22 @@
 using System.Text;
-using JetBrains.Annotations;
+using System.Threading;
 
 namespace RegressionGames.StateRecorder
 {
     public interface IStringBuilderWriteable
```
RG-nAmKcAz, line 6: This now does what its name says it does and there is a new IStringBuilderConverter interface encapsulating the old behaviour

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSequence.cs
```diff
@@ -442,6 +451,18 @@ public static (string, string, string) LoadJsonResource(string path)
...
         public void Play()
         {
+            if (ActiveBotSequence != null)
+            {
+                throw new Exception("Cannot start 2 bot sequences simultaneously.  Another bot sequence is already running.");
+            }
+            ActiveBotSequence = this;
+
+            var playbackController = UnityEngine.Object.FindObjectOfType<BotSegmentsPlaybackController>();
+            if (playbackController.GetState() != PlayState.NotLoaded)
```
vontell, line 461: I'm guessing we don't need to do the null check here because `BotSegmentsPlaybackController` definitely exists if we get to this point?

vontell (approved):
Nice! Very readable, appreciate the quick comment laying out the cases. Really no comments that require changes from me, so going to mark as approved.

src/gg.regression.unity.bots/Runtime/Scripts/RemoteOrchestration/Models/WorkAssignment.cs
```diff
@@ -0,0 +1,155 @@
...
+                        Stop();
+                        return;
+                    }
+                }
+
+                // check if the sequence is done
+                var activeBotSequence = RGRemoteWorkerBehaviour.GetActiveBotSequence();
+                if (activeBotSequence == null)
+                {
+                    var playbackController = UnityEngine.Object.FindObjectOfType<BotSegmentsPlaybackController>();
```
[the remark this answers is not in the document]
RG-nAmKcAz: I found that some games that pro-actively destroy objects on return to their first scene (match3 game I think it was...)
I had to fix a related / similar bug in existing code that batu found in his testing.

Therefore, this is 'safe' even if that overlay gets destroyed and recreated.

src/gg.regression.unity.bots/Runtime/Scripts/RemoteOrchestration/RGRemoteWorkerAssignmentManager.cs
```diff
@@ -0,0 +1,364 @@
...
+                // check to send registration if past the interval - rate limited to avoid request spam if server offline
+                if (_lastRegistrationTime + HeartbeatInterval < timeNow)
+                {
+                    _lastRegistrationTime = timeNow;
+                    var registrationRequest = new SDKClientRegistrationRequest(RGRemoteWorkerAssignmentManager.Guid)
+                    {
+                        availableSequences = _availableBotSequences
+                    };
+                    _registrationInProgress = true;
+                    _ = RGServiceManager.GetInstance().SendRemoteWorkerRegistration(
```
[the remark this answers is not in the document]
RG-nAmKcAz: I don't have a limit today because we don't know how long the communication will be offline.

src/gg.regression.unity.bots/Runtime/Scripts/RemoteOrchestration/RGRemoteWorkerAssignmentManager.cs
```diff
@@ -0,0 +1,364 @@
...
+                            activeWorkAssignment = incomingWorkAssignment
+                        };
+                        // send report of CONFLICT immediately, don't mess with the active work assignment
+                        _ = RGServiceManager.GetInstance().SendRemoteWorkerHeartbeat(heartbeatRequest, HeartbeatResponseHandler, () => { });
+                    }
+                    else
+                    {
+                        // nothing else is running currently... GOOD... we mark this work assignment as ready to start on next update
+                        ActiveWorkAssignment = incomingWorkAssignment;
+                        ActiveWorkAssignment.status = WorkAssignmentStatus.WAITING_TO_START;
```
abeizer, line 263: I've been assuming the server will set the status as "WAITING_TO_START" at the time a work assignment is scheduled. This seems redundant, which doesn't hurt anything, but want to confirm whether an additional status like "SCHEDULED" would be useful

src/gg.regression.unity.bots/Runtime/Scripts/RemoteOrchestration/Models/ActiveSequence.cs
```diff
@@ -0,0 +1,28 @@
...
+using UnityEngine.Animations;
+
+// ReSharper disable InconsistentNaming
+namespace RegressionGames.RemoteOrchestration.Models
+{
+    [Serializable]
+    public class ActiveSequence
+    {
+        public string name;
+        public string resourcePath;
```
abeizer, line 14: Should also have `description` here, for the webapp's sake. 9 times out of 10, I expect the active sequence to match one from the availableSequences list one-to-one, but we've mentioned that users can change the name and description after registration, which won't be reflected in that list

abeizer (approved):
Good stuff 👍 

Comments are non-blocking / feelers for possible enhancements

src/gg.regression.unity.bots/Runtime/Scripts/RemoteOrchestration/Models/ActiveSequence.cs
abeizer: I've been thinking about how the webapp will respect discrepancies between the registered `availableSequences` and sequences that are modified post-registration. If we're viewing the clients at some moment in time, then obviously we can show trust the active sequence as the "correct version". But when we schedule a workAssignment for a client, we won't know about any post-registration changes to that list, meaning any historical view of completed assignments won't match up with what the dashboard says the client actually ran.

This might be a nothing-burger / future issue. But we could send the `availableSequences` in the heartbeat to keep them up-to-date. A WorkAssignment would then capture the entire sequence object from the available list at the time it's scheduled, so we get a more accurate snapshot of what "version" the client actually ran for that assignment.

RG-nAmKcAz: I think nothing-burger.  In order for you to get one after registration.  You'd have a to create a new sequence in a game instance already running with -rgremoteworker.  This would mean that a user is manually controlling a system meant for automation.  And not just controlling it, using it as a live dev system.  I think it will be fine for any new sequence not to show up in the remote console until the next game play restart

RG-nAmKcAz: As for description, I can add this in I guess, but currently I don't see any UI flows that show anything other than the name and path.

addisonbgross (approved):
What a journey. Great job on the comments, they made all of these changes easier to digest. I left just one, minuscule nit of a comment

src/gg.regression.unity.bots/Runtime/Scripts/RemoteOrchestration/Models/ActiveSequence.cs
abeizer: Bringing up adding/removing was sort of tangential, I apologize. 

The core of my point is... 
why are we passing an object instead of just the resourcePath? -> Because we said the name (and description) could change. 
What does that mean for the UI? -> Viewing active clients will show the "correct" value based on `activeSequence` but a client's `WorkAssignment`s only have visibility to the list. Thus if the `availableSequences` list never reflects changes to name/description then the `WorkAssignment` be forced to display "inaccurate" information (ie. the name and description of the sequence at time of registration rather than after being modified post-registration).

Does this clarify what I was trying to get at?

RG-nAmKcAz: - As for passing an object.  I was doing this mostly to clarify the case where something that is NOT a BotSequence (segmentZip/segmentList/segment/etc) is running.  Since that will NOT have a resourcePath, but WILL have an activeSequence entry in the status so that the server realizes something is running

```csharp
                // else - a zip file or other bot segments are running outside of a sequence
                return new ActiveSequence()
                {
                    name = "BotSegments are active outside of a BotSequence",
                    resourcePath = "",
                    sequence = null
                };
```
- "(ie. the name and description of the sequence at time of registration rather than after being modified post-registration)"  ... this is what i just tried to respond to above.  Modifications post registration on one of these systems are incredibly unlikely and if so, not a big deal to not be reflected until next start of the game.

src/gg.regression.unity.bots/Runtime/Scripts/RemoteOrchestration/RGRemoteWorkerAssignmentManager.cs
```diff
@@ -0,0 +1,364 @@
...
+                        incomingWorkAssignment.details = new ConflictDetails(ActiveWorkAssignment.id, incomingWorkAssignment.id);
+                        var heartbeatRequest = new SDKClientHeartbeatRequest()
+                        {
+                            clientId = this.ClientId,
+                            activeSequence = GetActiveBotSequence(),
+                            activeWorkAssignment = incomingWorkAssignment
+                        };
+                        // send report of CONFLICT immediately - if we have a running sequence this will tell them that
+                        _ = RGServiceManager.GetInstance().SendRemoteWorkerHeartbeat(heartbeatRequest, HeartbeatResponseHandler, () => { });
+                        if (ActiveWorkAssignment != null)
```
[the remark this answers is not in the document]
RG-nAmKcAz: good call out,

I actually refactored this a bit to avoid an unlikely network race condition in that update

src/gg.regression.unity.bots/Runtime/Scripts/RemoteOrchestration/Models/ActiveSequence.cs
RG-nAmKcAz: `description` added

abeizer: Ah okay I understand now, thank you

PR #335, opened 2024-10-10
[REG-2110] Implement the GetRuntimeLogs command in the Runespawn server.

batu:
[Python side component.](https://github.com/Regression-Games/Runespawn/pull/10)

I tried to be mindful of unsubscribing events, but I don't know if what I ended up with is the right way to do it.

I ended up simply writing the code myself instead of using our built in log detailing as it was as simple as: 
`_logs.Add($"[{type}] {logString}");`

I also made KeyboardEventSender default log level from info to debug, as it would clog up the LLM context.

---

### DIFF
--- a/src/gg.regression.unity.experimental.runespawn/Editor/Scripts/EditorCommandServer.cs
+++ b/src/gg.regression.unity.experimental.runespawn/Editor/Scripts/EditorCommandServer.cs
@@ -39,6 +39,8 @@ public class EditorCommandServer
         EditorApplication.update += OnEditorUpdate;
         AssemblyReloadEvents.beforeAssemblyReload += OnBeforeAssemblyReload;
         EditorApplication.quitting += OnEditorQuitting;
+
+        // TODO (REG-2113): Use LoggingObserver to parse runtime logs.
         Application.logMessageReceived += PlayModeController.OnLogMessageReceived;
     }
 
--- a/src/gg.regression.unity.experimental.runespawn/Editor/Scripts/EditorCommandServer.cs
+++ b/src/gg.regression.unity.experimental.runespawn/Editor/Scripts/EditorCommandServer.cs
@@ -19,7 +19,8 @@ public class EditorCommandServer
         Stop,
         GetScripts,
         Compile,
-        GetCompilationMessages
+        GetCompilationMessages,
+        GetRuntimeLogs
     }
 
     // Configuration: Read from environment variables with defaults
@@ -38,6 +39,7 @@ public class EditorCommandServer
         EditorApplication.update += OnEditorUpdate;
         AssemblyReloadEvents.beforeAssemblyReload += OnBeforeAssemblyReload;
         EditorApplication.quitting += OnEditorQuitting;
+        Application.logMessageReceived += PlayModeController.OnLogMessageReceived;
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


batu:
"There was an error while trying to activate the Unity license." is the build error

src/gg.regression.unity.experimental.runespawn/Editor/Scripts/PlayModeController.cs
```diff
@@ -10,6 +11,50 @@ public static class PlayModeController
...
+    private static List<string> _errors = new List<string>();
+
+    /// <summary>
+    /// Static constructor for PlayModeController.
+    /// Initializes event subscriptions and prevents multiple subscriptions.
+    /// </summary>
+    static PlayModeController()
+    {
+        // Prevent multiple subscriptions.
+        EditorApplication.playModeStateChanged -= OnPlayModeStateChanged;
```
RG-nAmKcAz, line 24: I would test this thoroughly..  If you ever get a case where you try to remove a listener that doesn't exist it will throw a Unity exception.

RG-nAmKcAz (approved):
This works, but I'm a mildly concerned that we aren't using `LoggingObserver` from a thinking about the future perspective.

I guess Runespawn doesn't cause us to record state yet for later analysis in our WebClient, but I would expect one day it will.

Before we get there, we need to be observing and processing log messages only one time; but this class listening and LoggingObserver listening from ScreenRecorder would cause us to double process each log statement and use more Ram and more CPU than we should.

I would NOT bother fixing this now, but I would add a follow up task for handling it.  Doing it will require some changes to LoggingObserver to track the number of consumers registered so that we can track which messages in the buffer haven't been flushed to each consumer yet.  If/When we get there, I'd like to work on that task.

RG-nAmKcAz:
@vontell Can you look into the Unity License error in the builds ?
"There was an error while trying to activate the Unity license."

batu:
I added a [TODO](https://linear.app/regression-games/issue/REG-2113/use-loggingobserver-to-parse-runtime-logs) and also created a [Productionize Runespawn](https://linear.app/regression-games/issue/REG-2112/productionize-runespawn) task where I will adding all the productionization tasks so we don't forget any of them.

src/gg.regression.unity.experimental.runespawn/Editor/Scripts/PlayModeController.cs
batu: I tested it in this context and it works. Let me if I might be missing something

```csharp
 public class BotTester : MonoBehaviour
    {
       UnityEvent m_MyEvent;
       void Start()
       {
            Debug.Log("BotTester: Started.");

            Application.logMessageReceived -= SomeRandomFunction;
            Debug.Log("BotTester: Removed subscription.");

            m_MyEvent = new UnityEvent();
            m_MyEvent.RemoveListener(SomeOtherRandomFunction);
            Debug.Log("BotTester: Removed listener.");
       }

        private void SomeRandomFunction(string condition, string stackTrace, LogType type){}
        private void SomeOtherRandomFunction(){}

    }
```

src/gg.regression.unity.experimental.runespawn/Editor/Scripts/EditorCommandServer.cs
```diff
@@ -38,6 +39,9 @@ static EditorCommandServer()
         EditorApplication.update += OnEditorUpdate;
         AssemblyReloadEvents.beforeAssemblyReload += OnBeforeAssemblyReload;
         EditorApplication.quitting += OnEditorQuitting;
+
+        // TODO (REG-2113): Use LoggingObserver to parse runtime logs.
```
vontell, line 43: Thanks for adding this note!

vontell (approved):
Similar comment to Zack but I think you addressed it. Also the build issue should be fixed, I'll restart your build and see if it passes.

PR #337, opened 2024-10-18
[REG-2129] SDK Side: Implement the ability for Runespawn to be able to play bot sequences.

batu:
This PR implements the SDK side of being able to start bot sequences via runespawn. This is important for cases where we don't want the runespawn code to execute immediately, but rather run a sequence first. (Such as loading into level in Match3)

[Python side PR.](https://github.com/Regression-Games/Runespawn/pull/14)


src/gg.regression.unity.experimental.runespawn/Editor/Scripts/PlayModeController.cs
```diff
@@ -68,6 +72,25 @@ public static void StartPlayMode(TcpClient client)
             EditorApplication.delayCall += () =>
             {
                 Application.runInBackground = true;
+
+                _botSequencePath = botSequencePath;
```
batu, line 76: I dislike the usage of global variables here. 

I also tried using a lambda to make a closure (is this the correct term?) / bake the botSequencePath into the delegate, however it made unsubscribing finnicky.   

```csharp
                EditorApplication.playModeStateChanged += (PlayModeStateChange state) =>
                {
                    RunBotSequenceOnPlayModeEnter(state, botSequencePath);
                };
```

Thoughts?

src/gg.regression.unity.experimental.runespawn/Editor/Scripts/EditorCommandServer.cs
```diff
@@ -163,7 +163,18 @@ private static void HandleClient(object obj)
 
             if (Enum.TryParse(commandStr, true, out CommandType commandType))
             {
-                EditorCommand editorCommand = new EditorCommand(commandType, client);
+                EditorCommand editorCommand;
+                if (commandType == CommandType.Start)
+                {
+                    // Read the optional bot sequence path
+                    string botSequencePath = Utilities.ReadLine(stream);
```
vontell, line 170: Does this throw an error if there is no line? Or does it return null?

src/gg.regression.unity.experimental.runespawn/Editor/Scripts/PlayModeController.cs
```diff
@@ -68,6 +72,25 @@ public static void StartPlayMode(TcpClient client)
...
+                // If a bot sequence path is provided, run the sequence after entering play mode
+                if (!string.IsNullOrEmpty(botSequencePath))
+                {
+                    if (!System.IO.File.Exists(botSequencePath))
```
vontell, line 80: @RG-nAmKcAz may have a more specific recommendation, but we should use some method or utility in our bot sequence code for determining if the botSequencePath exists, because it's actually much more complicated than if it exists on file system. There is a case that since this is only running in editor for now, this should technically always work, but I'd rather be more precise. I'd take a look at the existing bot sequence code to see how we check and use the path.

vontell: Maybe you could just load it before this with the try catch?

src/gg.regression.unity.experimental.runespawn/Editor/Scripts/PlayModeController.cs
vontell: I think it's fine in this case, but make sure to reset the variable when the sequence is done, has failed, or when a new change request has come in maybe? I'm not too sure about our static variable usage since we usually go with an instanced singleton class, but also since you already have static _logs and _errors, I don't see this being any different from those.

vontell (changes requested):

src/gg.regression.unity.experimental.runespawn/Editor/Scripts/PlayModeController.cs
RG-nAmKcAz: `BotSequence.LoadSequenceJsonFromPath`

RG-nAmKcAz: We can save the reference to the parsed bot sequence itself instead of saving the path and trying to load it later when it tries to start

RG-nAmKcAz (approved):
Approved pending resolution of the feedback from @vontell

src/gg.regression.unity.experimental.runespawn/Editor/Scripts/PlayModeController.cs
batu: Switched to loading the sequence itself and saving that.

src/gg.regression.unity.experimental.runespawn/Editor/Scripts/EditorCommandServer.cs
batu: Good catch, my bad for missing this. Added the check here.

vontell (approved):

PR #339, opened 2024-10-21
[REG-2127] Implement SimulationInterface.

batu:
This PR is to allow Runespawn to use the mouse simulation API by updating the prompt.

It exposes two public functions in MouseEventSender.cs:

[Python Side PR.](https://github.com/Regression-Games/Runespawn/pull/15)

```csharp
public static void SendMouseEvent(Vector2 mouseScreenPosition,
                                  bool leftButton=false,
                                  bool middleButton=false,
                                  bool rightButton=false,
                                  bool forwardButton=false,
                                  bool backButton=false,
                                  Vector2 scroll = default)
```

and 

```csharp
public static void SendMouseEvent(Transform mouseTarget,
                                  bool leftButton=false,
                                  bool middleButton=false,
                                  bool rightButton=false,
                                  bool forwardButton=false,
                                  bool backButton=false,
                                  Vector2 scroll = default)
```

This is to make it easier for the LLM to write code that clicks on transforms.

Example video:

https://github.com/user-attachments/assets/f0007fa5-d928-411c-a07c-af4373b71141

Example Code:
[BotTester.txt](https://github.com/user-attachments/files/17466130/BotTester.txt)


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseEventSender.cs
```diff
@@ -366,7 +366,63 @@ public static void SendMouseEvent(int replaySegment, MouseInputActionData mouseI
...
+        {
+            Vector2 screenPosition = Camera.main.WorldToScreenPoint(mouseTarget.position);
+            #if ENABLE_LEGACY_INPUT_MANAGER
+            SendMouseEventLegacy(screenPosition, Vector2.zero, scroll, leftButton, middleButton, rightButton, forwardButton, backButton);
+            #else
+            var screenWidth = Screen.width;
+            var screenHeight = Screen.height;
+            var normalizedPosition = new Vector2(screenPosition.x / screenWidth, screenPosition.y / screenHeight);
+            SendRawPositionMouseEvent(0, normalizedPosition, leftButton, middleButton, rightButton, forwardButton, backButton, scroll);
+            #endif
```
vontell, line 397: I'd just have this call your other function? It seems like the code is identical except that it gets the position of the transform first, which you could pass into the other function.

vontell (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseEventSender.cs
```diff
@@ -366,7 +366,63 @@ public static void SendMouseEvent(int replaySegment, MouseInputActionData mouseI
...
+        /// <param name="scroll">Simulates mouse scroll wheel movement.</param>
+        public static void SendMouseEvent(Transform mouseTarget,
+                                          bool leftButton=false,
+                                          bool middleButton=false,
+                                          bool rightButton=false,
+                                          bool forwardButton=false,
+                                          bool backButton=false,
+                                          Vector2 scroll = default)
+        {
+            Vector2 screenPosition = Camera.main.WorldToScreenPoint(mouseTarget.position);
```
RG-nAmKcAz, line 389: This needs a bit more processing.  The transform position of an object is seldomly the 'center' of that object.

We have code for determining that center already in `MouseEventSender.SendMouseEvent` that you can lift / use as a reference

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseEventSender.cs
```diff
@@ -366,7 +366,63 @@ public static void SendMouseEvent(int replaySegment, MouseInputActionData mouseI
...
+        /// Simulates a mouse event at the specified screen position. Bottom left is (0,0) and top right is (Screen.width, Screen.height).
+        /// </summary>
+        /// <param name="mouseScreenPosition">The screen coordinates where the mouse event will be simulated.</param>
+        /// <param name="leftButton">Simulates left mouse button press if true.</param>
+        /// <param name="middleButton">Simulates middle mouse button press if true.</param>
+        /// <param name="rightButton">Simulates right mouse button press if true.</param>
+        /// <param name="forwardButton">Simulates forward mouse button press if true.</param>
+        /// <param name="backButton">Simulates back mouse button press if true.</param>
+        /// <param name="scroll">Simulates mouse scroll wheel movement.</param>
+        public static void SendMouseEvent(Vector2 mouseScreenPosition,
```
RG-nAmKcAz, line 410: This already exists as `SendRawPositionMouseEvent`, let's avoid creating another method for it.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseEventSender.cs
```diff
@@ -366,7 +366,63 @@ public static void SendMouseEvent(int replaySegment, MouseInputActionData mouseI
...
+                                          bool forwardButton=false,
+                                          bool backButton=false,
+                                          Vector2 scroll = default)
+        {
+            #if ENABLE_LEGACY_INPUT_MANAGER
+            SendMouseEventLegacy(mouseScreenPosition, Vector2.zero, scroll, leftButton, middleButton, rightButton, forwardButton, backButton);
+            #else
+            var screenWidth = Screen.width;
+            var screenHeight = Screen.height;
+            var normalizedPosition = new Vector2(mouseScreenPosition.x / screenWidth, mouseScreenPosition.y / screenHeight);
```
RG-nAmKcAz, line 423: I don't understand this logic.  It looks like it is computing a fractional percentage of the screen dimensions for the mouse instead of just sending the coordinates ?

Send Raw Position mouse event already expects the mouse position in terms of the width/height.  This code doesn't seem like it creates the correct mouse inputs

RG-nAmKcAz (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseEventSender.cs
batu: For coding practices 100%. The reason why it is just as a repeat method with the same name is to make it easier for the LLM. I wanted it to be 'There is a SendMouseEvent function, you can either pass in a transform or pixel position, either does something sensible.`

@RG-nAmKcAz @vontell On second thought it might be potentially better to create a separate LLMInterface script that can contain both this and KeyboardInput. Then we can just give this public API to runespawn, and it will be much easier to maintain. I will create a new task for the next PR to combine the two if no one objects.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseEventSender.cs
batu: I spent a good amount of time but wasn't able to do this cleanly.

Initially I tried calculating the 
- bestObject.screenSpaceBounds
- whether something is the worldSpaceObject
- and possibleObjects, but it took me a long time to understand what possibleObjects are

so I switched to calling FindBestClickObject.

However, that lead me to requiring `transforms`. Which I believe I can get via objectFinders, which I can get through `var FindObjectsByType<ObjectFinder>(FindObjectsSortMode.None);` but that requires coupling this static class to a monobehavior.FindObjectsByType call and now think I am doing something wrong.

Can you give me a bit more context on what would be the best way to do this?

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseEventSender.cs
batu: Fixed. It was wrong. When I was reading the code I understood it correctly. It would be helpful if we at least add input/return type docs especially for things like this.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseEventSender.cs
batu: @RG-nAmKcAz pinging here.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseEventSender.cs
RG-nAmKcAz: We aren't in a rush to release this, so I would like it cleaned up in this PR where the feedback was given instead of deferred to later.  I don't want to introduce new public methods to the SDK that effectively duplicate the functionality of existing methods.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseEventSender.cs
RG-nAmKcAz: ... taking a look...

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseEventSender.cs
batu: @RG-nAmKcAz Okay, I will create the llm interface in this PR.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseEventSender.cs
RG-nAmKcAz: @batu

Ultimately all that code I sent you to look at in the prior comments... uses the transforms found for the current frame.

Specifically it uses the `screenSpaceBounds` for the status of each found transform.  Said `screenSpaceBounds` are set in this code from `TransformObjectFinder`.



```csharp
var bounds = SelectBoundsForTransform(mainCamera, screenWidth, screenHeight, theTransform);
tStatus.screenSpaceBounds = bounds.Item1;
tStatus.screenSpaceZOffset = bounds.Item2;
tStatus.worldSpaceBounds = bounds.Item3;
```


So ultimately, your code should be .... 

```csharp
var boundsInfo = TransformObjectFinder.SelectBoundsForTransform(Camera.main, Screen.width, Screen.height, mouseTarget);
var ssBounds = boundsInfo.Item1;
if (ssBounds.HasValue)
{
    Vector2 screenPosition = ssBounds.Value.center;
    // do your action
}
else
{
    // error where object transform isn't on screen
}
```

batu: sweet thank you

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseEventSender.cs
```diff
@@ -29,6 +29,7 @@ public static class MouseEventSender
 
         private static InputDevice _virtualMouse;
 
+        private static ObjectFinder[] _objectFinders;
```
vontell, line 32: Looks like this isn't used?

vontell (changes requested):
One small nit

batu:
This PR is totally redone. Now, it implements SimulationInterface which exposes all the functions that our LLM written agent should use.

The video shows:
1. Typing "Hello".
2. Click on an transform. (color change)
3. Click on a screen space. (Spawn ball)
4. Click and drag. (Adding force)

https://github.com/user-attachments/assets/720fe074-dfe9-4ea5-9bcc-1c379f1f1b8e

src/gg.regression.unity.experimental.runespawn/Editor/Runtime/SimulationInterface.cs
```diff
@@ -0,0 +1,133 @@
...
+    public static void SendMouseEvent(Vector2 mouseScreenPosition,
+                                      bool leftButton=false,
+                                      bool middleButton=false,
+                                      bool rightButton=false,
+                                      bool forwardButton=false,
+                                      bool backButton=false,
+                                      Vector2 scroll = default)
+    {
+        #if ENABLE_LEGACY_INPUT_MANAGER
+        MouseEventSender.SendMouseEventLegacy(mouseScreenPosition, Vector2.zero, scroll, leftButton, middleButton, rightButton, forwardButton, backButton);
```
RG-nAmKcAz, line 101: `SendRawPositionMouseEvent` already calls this. If you test this on a legacy input game, you will probably see double mouse events as currently coded

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/MouseEventSender.cs
```diff
@@ -366,11 +365,10 @@ public static void SendMouseEvent(int replaySegment, MouseInputActionData mouseI
             }
 
             SendRawPositionMouseEvent(replaySegment, normalizedPosition, mouseInput.leftButton, mouseInput.middleButton, mouseInput.rightButton, mouseInput.forwardButton, mouseInput.backButton, mouseInput.scroll);
-
         }
 
 #if ENABLE_LEGACY_INPUT_MANAGER
-        private static void SendMouseEventLegacy(Vector2 position, Vector2 delta, Vector2 scroll,
+        public static void SendMouseEventLegacy(Vector2 position, Vector2 delta, Vector2 scroll,
```
RG-nAmKcAz, line 371: This should stay private, it is called internally from sendrawposition mouseevent

src/gg.regression.unity.experimental.runespawn/Editor/Runtime/SimulationInterface.cs
```diff
@@ -0,0 +1,133 @@
...
+    /// Initializes the simulation interface.
+    /// </summary>
+    /// <param name="monoBehaviour">The MonoBehaviour instance to use for coroutine execution.</param>
+    /// <remarks>
+    /// This method performs the following initializations:
+    /// 1. Starts the RGLegacyInputWrapper simulation.
+    /// 2. Initializes the KeyboardEventSender.
+    /// 3. Initializes the virtual mouse for MouseEventSender.
+    /// </remarks>
+    public static void Initialize(MonoBehaviour monoBehaviour)
```
RG-nAmKcAz, line 21: This should also have some sort of teardown method to undo the setup steps from this Initialize method

RG-nAmKcAz (changes requested):

vontell (commented):
Nothing in addition to Zack's comments, will take a look again once those are addressed.

src/gg.regression.unity.experimental.runespawn/Editor/Runtime/SimulationInterface.cs
batu: @RG-nAmKcAz I tested it with both `Both and Input Manager (Old)`. Is this sufficient to switch? It sends the correct number clicks with that change in both.

Either way switched to only using `SendRawPositionMouseEvent`

src/gg.regression.unity.experimental.runespawn/Editor/Runtime/SimulationInterface.cs
batu: Added a teardown method, mostly mirroring it from BotSegmentsPlaybackController.Stop(). 

Currently, we regain control of mouse and keyboard after calling Teardown.

RG-nAmKcAz (approved):

batu:
Tested mouse on Match3, and keyboard simulation on MicroPlatformer.

batu:
@vontell Addressed Zack's comments. PTAL when you have a mo

vontell (approved):

PR #342, opened 2024-10-23
[REG-2132] Eat our own dogfood testing

RG-nAmKcAz:
- Creates a new test scene to test our own SDK during the build

---


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentsPlaybackController.cs
```diff
@@ -224,7 +224,8 @@ public void Stop()
             }
 
             _nextBotSegments.Clear();
-            _isPlaying = PlayState.NotLoaded;
+            _playState = PlayState.NotLoaded;
+            BotSequence.ActiveBotSequence = null;
```
vontell, line 228: I think this may have been the change that fixed my issue. Thanks!

src/gg.regression.unity.bots/Tests/Scripts/GameScene/GameSceneTests.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        [OneTimeSetUp]
+        public void OneTimeSetUp()
+        {
+            SavePreviousViewAspect();
+            //force this to match our recording resolution
+            Screen.SetResolution(1920, 1080, false);
+
+            // set to full HD (1920x1080)
+            SetEditorAspectRatio(3);
```
vontell, line 35: Glad you got this to work! Ive been trying this in my CI/CD setup tutorial and was having difficulty getting it working.

src/gg.regression.unity.bots/Tests/Scripts/GameScene/GameSceneTests.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        [UnitySetUp]
+        public IEnumerator SetUp()
+        {
+            var botManager = Object.FindObjectOfType<RGBotManager>();
+            if (botManager != null)
+            {
+                // destroy any existing overlay before loading new test scene
+                Object.Destroy(botManager.gameObject);
+            }
```
vontell, line 48: I saw this in a few places, just out of curiosity why does this need to be removed? I would think that since we are loading a new scene, this would be removed anyways but I'm guessing you found this was not the case?

RG-nAmKcAz:
@vontell @abeizer , I am disabling the new test for the moment to get the other fixes in this PR mergeable

I will continue on monday with figuring out why the test case fails on the build system (which may be a nightmare as I can't see the video output :/)

vontell (approved):
Really no changes requested from me, really appreciate you going through and trying this out. As for why GameCI integration is showing failing tests... happy to help debug this, but I'm also happy with punting that in favor of getting this PR in, since that could be a time consuming task.

Definitely inspired for us to prioritize more dogfooding and also potentially kicking off some projects related to testing the SDK across a variety of games to validate stability. Nice work!!!!

src/gg.regression.unity.bots/Tests/Scripts/GameScene/GameSceneTests.cs
RG-nAmKcAz: dontdestroyonload - persists across scene loads, even in testing

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentsPlaybackController.cs
vontell: Context for others seeing this comment, the issue I was seeing is that after I ran or stopped a sequence, it wouldn't allow me to start a new sequence because it thought I still had an active sequence.

PR #344, opened 2024-10-29
[REG-2144] Simplify recorded BotSegment criteria + performance optimizations

RG-nAmKcAz:
- By default, now captures only the changed or clicked on options in recorded bot segment endCriteria
- Adds an internal/dev option for testing out recording minimum mouse state around clicks
  - This currently breaks games that lock the mouse to the camera, like First Person shooter style
- Improves performance and memory allocation issues around log capture 
- Improves performance and memory allocation issues around screenshot capture

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
@@ -61,6 +61,12 @@ namespace RegressionGames.StateRecorder
         [Tooltip("Directory to save state recordings in.  This directory will be created if it does not exist.  If not specific, this will default to 'unity_videos' in your user profile path for your operating system.")]
         public string stateRecordingsDirectory = "";
 
+        [Tooltip("Minimize the amount of criteria in bot segment recordings.  This limits the endCriteria in recorded bot segments to only include objects with count deltas and objects that were under click locations.")]
+        public bool minimizeRecordingCriteria = true;
+
+        [Tooltip("RG-INTERNAL/EXPERIMENTAL: Minimize the amount of mouse movement data capture in bot segment recordings.  This currently breaks replay on any game where camera movement is locked to the mouse, like First Person Shooters.")]
+        public bool minimizeRecordingMouseMovements = false;
+
         private double _lastCvFrameTime = -1;
 
         private int _frameCountSinceLastTick;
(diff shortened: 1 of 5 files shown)


RG-nAmKcAz:
@batu, this PR isn't marked for review yet

I'll answer your points anyway though
- Match3 - I don't test with match 3 at all.  That game is so variable even in the menus its infuriating
- Bossroom - There is a draft PR for bossroom with this change (also not yet marked for review)

Really though, there isn't a before / after as the core processing on replay hasn't changed at all.  This only impacts what we record in the first place

abeizer (approved):

batu (approved):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenRecorder.cs
```diff
@@ -61,6 +61,12 @@ public class ScreenRecorder : MonoBehaviour
         [Tooltip("Directory to save state recordings in.  This directory will be created if it does not exist.  If not specific, this will default to 'unity_videos' in your user profile path for your operating system.")]
         public string stateRecordingsDirectory = "";
 
+        [Tooltip("Minimize the amount of criteria in bot segment recordings.  This limits the endCriteria in recorded bot segments to only include objects with count deltas and objects that were under click locations.")]
+        public bool minimizeRecordingCriteria = true;
+
+        [Tooltip("RG-INTERNAL/EXPERIMENTAL: Minimize the amount of mouse movement data capture in bot segment recordings.  This currently breaks replay on any game where camera movement is locked to the mouse, like First Person Shooters.")]
+        public bool minimizeRecordingMouseMovements = false;
```
vontell, line 68: Another way to put this might be "useful where the way that a mouse moves from one click event to another doesn't matter". No need to add that unless it resonates a lot with you

vontell (approved):
I like the ideas here a lot. When you say performance improves, do you mean the resilience of the approach to handle different situations? I imagine this will help a bit with games where there might be a lot of game objects that we don't necessarily care about for a particular test.

Anyhow, like what I see here, LGTM!

RG-nAmKcAz:
@vontell the performance bullets were 'extra' fixes that are in this PR that were annoying me while developing this

the log capture feature had re-introduced some frame stutter , especially during replay so I wanted to clean that up... 

while profiling I also noticed an area we could improve the image capture performance further

PR #346, opened 2024-10-31
[REG-2145] Add test APIs for using bot segments, actions, and criteria

RG-nAmKcAz:
- Adds APIs to our RGTestUtils to allow users to think about their Unity tests in terms of our BotActions and BotCriteria
  - Allows running a bot segment or bot segment list (in addition to the existing sequence support we had)
  - (I put these in to unlock the use of CV or other powerful criteria types in regular unity tests)
    - Allows running a bot action as a standalone test step
    - Allows waiting for bot criteria as a standalone test step
- Adds test with actions and criteria (this was hard to get right with the setup needed based on which input system and which event system was in use)
- Moves some test code to a more appropriate location 
- Hardens some code around inputData to make this easier and allow null/non-set fields instead of requiring empty objects 
- TODO: Will need to create a follow up task to document these things for both legacy and new input/event systems

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/KeyFrameCriteria.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/KeyFrameCriteria.cs
@@ -1,11 +1,13 @@
 ﻿using System;
 using System.Text;
+using System.Threading;
 using RegressionGames.StateRecorder.JsonConverters;
+// ReSharper disable InconsistentNaming
 
 namespace RegressionGames.StateRecorder.BotSegments.Models
 {
     [Serializable]
-    public class KeyFrameCriteria
+    public class KeyFrameCriteria : IStringBuilderWriteable
     {
         // api version of this top level schema, update if we add/change fields
         public int apiVersion = SdkApiVersion.VERSION_7;
@@ -15,7 +17,6 @@ namespace RegressionGames.StateRecorder.BotSegments.Models
         public IKeyFrameCriteriaData data;
 
         public int EffectiveApiVersion => Math.Max(apiVersion, data?.EffectiveApiVersion() ?? SdkApiVersion.CURRENT_VERSION);
-
 
         // Replay only - used to track if transient has ever matched during replay
         [NonSerialized]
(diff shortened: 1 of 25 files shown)


src/gg.regression.unity.bots/Tests/TestFramework/Scripts/RGTestUtils.cs
```diff
@@ -1,29 +1,36 @@
...
 {
-    /// <summary>
-    /// Utilities for running tests within the Unity Test Runner using Regression Games features.
-    /// </summary>
+    /**
+     * <summary>
+     * Utilities for running tests within the Unity Test Runner using Regression Games features.
+     * </summary>
+     */
     public class RGTestUtils
```
RG-nAmKcAz, line 25: Github folds this file by default, but this file is where the bulk of the meaningful changes are

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/KeyFrameCriteria.cs
```diff
@@ -1,11 +1,13 @@
 ﻿using System;
 using System.Text;
+using System.Threading;
```
vontell, line 3: Was this added accidentally? I don't see any change here that would need this import but maybe I am missing something.

vontell (approved):
This is awesome! I've been hoping for something like this for a while so great to see this get in here. Thanks for adding the tests too, serves as a good example.

One small question about an import, so won't block on that, but am curious about that

src/gg.regression.unity.bots/Tests/Scripts/BotSegments/BotSegmentsTests.cs
```diff
@@ -0,0 +1,193 @@
...
+            RGUtils.SetupOverrideEventSystem();
+
+            GameObject eventSystem = GameObject.Find("EventSystem");
+            var eventSys = eventSystem.GetComponent<EventSystem>();
+            RGLegacyInputWrapper.UpdateMode = RGLegacyInputUpdateMode.AUTOMATIC;
+            RGLegacyInputWrapper.StartSimulation(eventSys);
+        }
+
+        [UnityTest]
+        public IEnumerator TestRGUtilsBotActionsAndCriteria()
```
RG-nAmKcAz, line 48: This is the new sample test using these APIs

addisonbgross (approved):

PR #350, opened 2024-11-06
[REG-2005] part 1 - Mark 'isOverride' on segments/sequences as they are loaded

RG-nAmKcAz:
- The loading code changes to mark `isOverride` on objects as they are loaded ... enables REG-2005 UI work

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
@@ -174,10 +174,10 @@ namespace RegressionGames
         /// <summary>
         /// Plays back a bot sequence, and then returns the save location of the recording.
         /// </summary>
-        /// <param name="botSequenceInfo">(filePath,resourcePath,BotSequence) tuple of the bot sequence</param>
+        /// <param name="botSequenceInfo">(filePath,resourcePath,BotSequence,isOverride) tuple of the bot sequence</param>
         /// <param name="setPlaybackResult">A callback that will be called with the results of this playback</param>
         /// <param name="timeout">How long in seconds to wait for this sequence to complete, less than or == 0 means wait forever (default=0)</param>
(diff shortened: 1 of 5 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
```diff
@@ -174,10 +174,10 @@ internal static int ParseTimeoutArgument()
         /// <summary>
         /// Plays back a bot sequence, and then returns the save location of the recording.
         /// </summary>
-        /// <param name="botSequenceInfo">(filePath,resourcePath,BotSequence) tuple of the bot sequence</param>
+        /// <param name="botSequenceInfo">(filePath,resourcePath,BotSequence,isOverride) tuple of the bot sequence</param>
         /// <param name="setPlaybackResult">A callback that will be called with the results of this playback</param>
         /// <param name="timeout">How long in seconds to wait for this sequence to complete, less than or == 0 means wait forever (default=0)</param>
-        internal static IEnumerator StartBotSequence((string,string, BotSequence) botSequenceInfo, Action<PlaybackResult> setPlaybackResult, int timeout = 0)
+        internal static IEnumerator StartBotSequence((string,string, BotSequence, bool) botSequenceInfo, Action<PlaybackResult> setPlaybackResult, int timeout = 0)
```
addisonbgross, line 180: Are we passing in the `isOverride` param just for consistency? It doesn't get used in this function

addisonbgross (approved):
Approved, just 1 non-blocking question

src/gg.regression.unity.bots/Runtime/Scripts/RGHeadlessSequenceRunner.cs
RG-nAmKcAz: the tuple we have where this is called is the 4 part tuple.. so passing that in instead of creating a new one was cleaner

vontell (approved):

abeizer (approved):

PR #352, opened 2024-11-06
[REG-2005] Display Override Indicator on Sequence + Segment cards, and in the Sequence Editor

addisonbgross:
image

## What has been done
- Added a visual override indicator to display when a Sequence or Segment seen in the UI is from a local file. IOW we're running a build of a game, and we have a Sequence or Segment in the persistent data path that overrides a Sequence or Segment that was built into the game
  - Added to the `SegmentCardLarge`, `SequenceCard`, and `SequenceEditor` prefabs
  - The override indicator itself is its own prefab. The red banner gets its clipped shape by adding a `RectMask2D` component to the parent object, and adjusting the mask's padding a bit
- Removed some superfluous null checks in the initialization of a `RGBotSequenceEntry`' script. Let's make those configuration errors loud!
- Renamed the tooltips on the Sequence + Segment prefabs to be more specific
- Added tests for the override indicator, and fixed up any other tests that were affected by my changes

## LOOM
https://www.loom.com/share/ce9516d061f8482dbfa84604ef7cdbb2

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Prefabs/RGOverlay/OverrideIndicator.prefab
+++ b/src/gg.regression.unity.bots/Runtime/Prefabs/RGOverlay/OverrideIndicator.prefab
@@ -84,7 +84,7 @@ GameObject:
   m_PrefabAsset: {fileID: 0}
   serializedVersion: 6
   m_Component:
-  - component: {fileID: 2753814926951463455}
+  - component: {fileID: 7388321243248179541}
   m_Layer: 0
   m_Name: OverrideIndicator
   m_TagString: Untagged
@@ -92,8 +92,8 @@ GameObject:
   m_NavMeshLayer: 0
   m_StaticEditorFlags: 0
   m_IsActive: 1
---- !u!4 &2753814926951463455
-Transform:
+--- !u!224 &7388321243248179541
+RectTransform:
   m_ObjectHideFlags: 0
   m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInstance: {fileID: 0}
@@ -109,6 +109,11 @@ Transform:
   m_Father: {fileID: 0}
   m_RootOrder: 0
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
+  m_AnchorMin: {x: 0.5, y: 0.5}
+  m_AnchorMax: {x: 0.5, y: 0.5}
+  m_AnchoredPosition: {x: 0, y: 0}
+  m_SizeDelta: {x: 100, y: 100}
+  m_Pivot: {x: 0.5, y: 0.5}
 --- !u!1 &7423362218327648588
 GameObject:
   m_ObjectHideFlags: 0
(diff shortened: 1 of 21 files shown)


RG-nAmKcAz:
I'm thinking the text areas circled in this picture need to be context sensitive to create vs edit and whether this is an override.  I initially noticed the second paragraph seemed weird when you already were editing an override, then i read the first paragraph and saw that is was describing create but not really update/edit.

image

RG-nAmKcAz (commented):
See comment above about updating the text on the create/update dialogue and making it context sensitive to create/update (and isOverride)

vontell (approved):
Very nice! I liken that there are good descriptions and tooltips to explain what is going on. LGTM!

addisonbgross:
@RG-nAmKcAz 3 states of instruction text have been added for create, edit, and edit + override

RG-nAmKcAz:
@addisonbgross where i can easily see these new screens ?

addisonbgross:
@RG-nAmKcAz in these handy screenshots

Create:
image

Edit:
image

Edit + override:
image

RG-nAmKcAz:
@addisonbgross 

Just to make sure I'm seeing what I'm expecting.

'Edit' will be shown when editing/updating a sequence that IS an override already ?

and

'Edit + override' will be shown when editing/updating an existing sequence that is about to create an override ?

batu (approved):
This looks good to me. Only nit is that I would much prefer "Overridden"  vs "Override" as initially the text suggested that it is an action I could take, but the UI style suggested it was "overridden" so had a mini confusion there.

RG-nAmKcAz:
Override seems more correct to me in this context.

addisonbgross:
@RG-nAmKcAz 
- Create is shown whenever creating a new Sequence (Editor or Build)
- Edit is shown when editing a Sequence in the Editor only
- Edit  + override is shown when editing a Sequence in a Build only

RG-nAmKcAz:
@addisonbgross 

ty, that makes sense...

- clarification for the last one though is the 'edit' text above this / also shown ?... showing the screenshots for the whole window would have been very helpful here for context


- and question/text-suggestion
Edit + override is shown when editing a Sequence in a Build only <- The text for this one is a bit mis-leading if you are editing something that is already an override.
  - Maybe we change `will create a copy of this Sequence in a local file` to `will save this Sequence as a local file`

addisonbgross:
@RG-nAmKcAz 
- For the Edit case, we're just showing the text in that screenshot. We have 3 different messages that we show, 1 for each case
- Good call on the text update. Done

RG-nAmKcAz:
@addisonbgross ... I think we should show both the 'edit' and 'edit+override' texts for the 'edit+override' case so that we still have the basic instructions there.

RG-nAmKcAz (approved):
Looks good once that failing test is updated

addisonbgross:
@RG-nAmKcAz I ended up skipping a test that requires loading a Sequence, as I could not find a way to mock the file system interaction without performing a sizable refactor of how we load Sequences/Segments. I think the best option is to use a mocking library in our tests.

The test I skipped exercises a function that sets the Sequence Editor into 'edit' mode: https://github.com/Regression-Games/RGUnityBots/blob/main/src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceEditor.cs#L151-L185

I made a Linear task for integrating a mock library. I'd be game to take it on next cycle: https://linear.app/regression-games/issue/REG-2179/integrate-a-mocking-library-into-our-sdk-tests

Please let me know if this seems like a prudent course of action

RG-nAmKcAz:
@addisonbgross Merge it :D

PR #354, opened 2024-11-12
[REG-2176] Skip graphics capture if graphics not available

vontell:
Sorry this took so long, took a while to iterate since building the game fully takes 15 minutes or so, and my approach of doing an early conditional break was acting up, so went with the more safe approach of just wrapping it in an `if` statement. Tested this in BossRoom and seems to work, and in our user's game, the game itself still doesn't work, but the graphics error is no longer thrown. There may be more changes to come, TBD, but this gets us partway there.


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenshotCapture.cs
```diff
@@ -320,51 +320,58 @@ private IEnumerator UpdateGPUData(int frame, Action<(byte[], int, int)?> onCompl
                 }
 
                 var theGraphicsFormat = _screenShotTexture.graphicsFormat;
-
-                try
+                
+                // If we are running in -nographics mode, the async task fails, causing an exception inside
+                // the AsyncGPUReadback.Request that is difficult to catch. This ensures that the screenshot
+                // is only taken when we have graphics
+                if (SystemInfo.graphicsDeviceType != GraphicsDeviceType.Null)
```
vontell, line 327: Review this with whitespace changes turned off, all that changed here is wrapping it in this if statement

RG-nAmKcAz: This code is in the wrong place.  In the current location, the GPUReadbackRequests structure will fill up with key-null associations and not get cleaned up until the end of the recording when WaitForCompletion is called.  It would be better to position this code so that the readback requests are never even tracked, more up towards `WaitForEndOfFrame` in this method.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/GameFacePixelHashObserver.cs
```diff
@@ -107,7 +107,11 @@ private void UpdateGameFacePixelHash()
                 if (cohtmlViewTexture != null)
                 {
                     var wasActive = Interlocked.CompareExchange(ref _requestInProgress, string.Empty, null);
-                    if (wasActive == null)
+                    
+                    // If we are running in -nographics mode, the async task below fails, causing an exception inside
+                    // the AsyncGPUReadback.Request that is difficult to catch. This ensures that the image data
+                    // is only read when we have graphics
+                    if (wasActive == null && SystemInfo.graphicsDeviceType != GraphicsDeviceType.Null)
```
RG-nAmKcAz, line 114: This needs to be closer to the start of the function.  We don't even need to bother with the cohtmlviewtexture lookup if we don't have graphics output

RG-nAmKcAz (changes requested):

batu (approved):
Nothing to add, pending Zack's comments

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenshotCapture.cs
vontell: Sorry for taking so long to respond to this - TLDR; I was worried that putting it earlier would have implications in other parts of the code that expect the readback requests to have data. But I _think_ where I put it now is fine, but it would be good to have confirmation.

The other holdup is now everything works as expected, I can run a sequence in headless mode in BossRoom for instance... but all of the zip folders are empty, including the data folders. This is something I need to look into now.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenshotCapture.cs
```diff
@@ -290,82 +290,102 @@ private IEnumerator UpdateGPUData(int frame, Action<(byte[], int, int)?> onCompl
...
-            // wait for end of frame or graphics data will be incorrect
-            yield return new WaitForEndOfFrame();
-
-            // get this in there so we don't get duplicate requests out for the same frame
-            if (GPUReadbackRequests.TryAdd(frame, null))
+            
+            // If we are running in -nographics mode, the async task fails, causing an exception inside
+            // the AsyncGPUReadback.Request that is difficult to catch. This ensures that the screenshot
+            // is only taken when we have graphics
+            if (SystemInfo.graphicsDeviceType != GraphicsDeviceType.Null)
```
vontell, line 297: (Just thinking aloud to myself here) On lines 284-285, I was worried that these actions wouldn't be completed properly if I placed the check here, but it seems like that's not the case... and should be fine

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenshotCapture.cs
vontell: (Note: I plan on debugging this myself, I just haven't had a chance to yet. I am going to look into the CompletedActions list and see if those are being completed correctly - if not, that might result in a failure that effects the process in ScreenRecorder.cs which writes the tick data dependent on those callbacks / results.)

RG-nAmKcAz: if that is what you run into, i'd still vote for doing this check early and just calling that completed action right away

RG-nAmKcAz:
@vontell I see you re-tagged me for review, but that seems to conflict with your comments above about still testing this.  I'd prefer to only re-review this one time so please request again once all testing/changes are done.

abeizer (changes requested):
Commenting so slack bot stops pinging me until this is ready

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenshotCapture.cs
vontell: Turns out I was correct in my worry and incorrect in whether this would be fine XD TLDR; because of how I wrote the original code, the onComplete handlers never get called, and so while screenshots are not taken, the handlers that would have gotten called which handle writing the ticks also never get called. The main change is now in the `GetCurrentScreenshotWithCallback`, which will just immediately invoke the completion handler, fixing the bug I was seeing. I've left this check here though as a safety measure since I wanted to keep this check near the code that actually does the readback request too.

vontell:
Ok this should finally be ready for review. Took a while to get to the cleanest solution, but I believe this would be it... TLDR; my original check would stop both the screenshot read and onCompletion handlers from being called ever, meaning that the action handlers that write the ticks to disk also never get called. I've left that check there as a safety precaution, but the main change is that now the onCompletion handlers are called right away in the case when graphics are not available.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenshotCapture.cs
```diff
@@ -255,13 +255,27 @@ public void WaitForCompletion()
...
         public void GetCurrentScreenshotWithCallback(long segmentNumber, Action<(byte[], int, int)?> onCompletion)
         {
             var frame = UnityEngine.Time.frameCount;
             lock (SyncLock)
             {
+
+                // When we don't have graphics available, we immediate invoke the callback with a null value. With how
+                // we use this function now, this needs to happen so the tick data writing task correctly saves the
+                // resulting data.
+                if (SystemInfo.graphicsDeviceType == GraphicsDeviceType.Null && onCompletion != null)
```
RG-nAmKcAz, line 273: this if block should be outside the synclock.. we don't need to acquire the lock as we aren't modifying any of the tracking structures

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/ScreenshotCapture.cs
RG-nAmKcAz: Please remove this code block.  It is un-necessary code with your other check in the correct place and actually is confusing.  IF you insist on leaving this code, then it should move to the start of this function before the synclock and would be similar to the code in the calling function where it checks and instantly invokes the completion action.  It would be cleaner to just remove this though

RG-nAmKcAz (changes requested):

vontell:
@RG-nAmKcAz edits made 👍

vontell:
Also just tested on Windows, this works there as well

abeizer (approved):

PR #357, opened 2024-11-19
[FIX] Avoid index out of bounds error when escaping chars

addisonbgross:
## What is the problem?
- When attempting to record gameplay in this [Vampire Survivors Clone](https://matthiasbroske.github.io/vampire-survivors.html), an error occurs that prevents the recording from happening

## What is the solution?
- When writing our game state to JSON during gameplay recording, we get an index out of bounds error when checking if individual chars in the game state need to be escaped
- This game uses Mandarin characters, even when you use the English language setting, and these Mandarin characters were causing the index out of bounds error
- We solve this error by being able to check if a char needs to be escaped, and handling any standard Unicode character (ie: codes \u0000 to \uFFFF)
  - The `char.MaxValue` == `\uFFFF`

## LOOM-ing
https://www.loom.com/share/e4302d036acf448f9a1bc3ae3ce42393


RG-nAmKcAz (approved):

batu (approved):
fascinating

PR #362, opened 2024-11-26
[REG-2194] Improved SDK UI on mobile

addisonbgross:
In an effort to improve our demos to potential clients building mobile games, I've made the UI a bit larger when running on a mobile device. This is definitely not a comprehensive solution, but it is a small win.  

## What has been done
- We can detect when a game is running on a mobile device, or in the Unity Device Simulator
- When on mobile or using the Unity Device Simulator:
  - Scale everything in our UI to be larger 
  - Disable the edit, copy, and delete actions for Sequences
  - Hide the 'create sequence button'. Creating or editing Sequences on mobile is too cramped
- When the Unity viewport is in a portrait orientation:
  - Scale everything in our UI to be larger 

## Screenshots
These were taken on my old OnePlus 7. Tested with both our [Match3Example](https://github.com/Regression-Games/Match3Example) game, and the [Vampire Survivors Clone](https://github.com/matthiasbroske/VampireSurvivorsClone) I was testing out recently

Screenshot_2024-11-25-16-19-09-95_433f1a8e1986f46fa5f9eab6651a65c2
Screenshot_2024-11-25-16-12-56-27_2c16d6eb5a2ead1fd6ae640023a34bb5
Screenshot_2024-11-25-16-21-17-14_433f1a8e1986f46fa5f9eab6651a65c2

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
@@ -175,10 +175,6 @@ public class RGSequenceManager : MonoBehaviour
     public void SetPortraitView()
     {
         GetComponent<CanvasScaler>().referenceResolution = new Vector2(800, 600);
-        
-        // the create sequence button is always the first child in the sequences panel
-        var createSequenceButton = sequencesPanel.transform.GetChild(0);
-        createSequenceButton.gameObject.SetActive(false);
     }
 
     /**
--- a/src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
@@ -172,7 +172,9 @@ public class RGSequenceManager : MonoBehaviour
      */
     public void UpscaleUI()
     {
-        GetComponent<CanvasScaler>().referenceResolution = new Vector2(800, 600);
+        var scaler = GetComponent<CanvasScaler>();
+        var currentResolution = scaler.referenceResolution;
+        scaler.referenceResolution = new Vector2(currentResolution.x * 0.6f, currentResolution.y * 0.6f);
     }
 
     /**
--- a/src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
@@ -169,7 +169,6 @@ public class RGSequenceManager : MonoBehaviour
      * <summary>
      * When the screen is in a portrait orientation:
      * - Scale the entire canvas so that all components are larger
-     * - Hide the create Sequence button
      * </summary>
      */
     public void SetPortraitView()
--- a/src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
@@ -6,6 +6,7 @@ using JetBrains.Annotations;
 using RegressionGames;
 using RegressionGames.StateRecorder;
 using UnityEngine;
+using UnityEngine.UI;
 // ReSharper disable once RedundantUsingDirective - used in #if #else - do not remove
 using RegressionGames.StateRecorder.BotSegments.Models;
 // ReSharper disable once RedundantUsingDirective - used in #if #else - do not remove
@@ -33,6 +34,8 @@ public class RGSequenceManager : MonoBehaviour
     public GameObject sequenceEditor;
 
     public GameObject deleteSequenceDialog;
+    
+    public Button reloadButton;
     
     private static RGSequenceManager _this;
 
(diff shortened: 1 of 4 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/RGUtils.cs
```diff
@@ -15,6 +15,14 @@ namespace RegressionGames
 {
     public static class RGUtils
     {
+        // <summary>
+        // Determines if the game is currently running on an Android or Apple mobile device
+        // </summary>
+        public static bool IsMobile()
+        {
+            return Application.platform == RuntimePlatform.Android || Application.platform == RuntimePlatform.IPhonePlayer;
```
vontell, line 23: I don't mind this, but this won't solve the problem for developers making mobile games in the editor. I suppose my biggest concern is more "portrait mode" than mobile specifically (i.e. just making sure our UI is usable for vertical devices)

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
```diff
@@ -91,6 +94,11 @@ public void Awake()
 
     public void Start()
     {
+        if (RGUtils.IsMobile())
```
vontell, line 97: In relation to the comment I make at the end of this file, I suppose what I am thinking might be `if (RGUtils.IsMobile() && is in portrait mode, i.e. screen height is larger than screen width)`

vontell (commented):
Just a comment/conversation on when this should be applied - TLDR I want to make sure that portrait mode in the editor is still usable?

src/gg.regression.unity.bots/Runtime/Scripts/RGUtils.cs
addisonbgross: ah I see where you're coming from. I'll make some updates and re-ping you 👍🏻

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
```diff
@@ -91,6 +94,15 @@ public void Awake()
...
+        {
+            SetMobileView();
+        }
+        else if (RGUtils.IsPortraitView())
```
RG-nAmKcAz, line 101: Here @addisonbgross , I think this should be an `if`

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
```diff
@@ -135,6 +147,35 @@ public void DeleteSequenceByPath(string path)
...
+     * <summary>
+     * When this component is viewed on a mobile device:
+     * - Scale the entire canvas so that all components are larger
+     * - Disable the reload Sequence/Segment button
+     * - Hide the create Sequence button
+     * </summary>
+     */
+    public void SetMobileView()
+    {
+        GetComponent<CanvasScaler>().referenceResolution = new Vector2(800, 600);
```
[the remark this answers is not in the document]
addisonbgross: From what I'm reading, since our Canvas Scaler component has its `match` field set to `0.5`, the scaling will be the same if the viewport is 800 x 600 or 600 x 800. In practice it seems to scale the same
image

addisonbgross:
@vontell @RG-nAmKcAz, I've made some changes. Please let me know if these 'rules' make sense:

When on mobile or using the Unity Device Simulator:
- Scale everything in our UI to be larger
- Disable the edit, copy, and delete actions for Sequences
- Hide the 'create sequence button'. Creating or editing Sequences on mobile is too cramped

When the Unity viewport is in a portrait orientation:
- Scale everything in our UI to be larger

I don't think we should disable anything if the game isn't running on mobile, but is in a portrait orientation. Curious to hear what you both think

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
addisonbgross: I dropped [a comment below](https://github.com/Regression-Games/RGUnityBots/pull/362#issuecomment-2501965271) that will hopefully address this comment 🤞🏻

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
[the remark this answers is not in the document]
vontell: For clarity regarding clarity on portrait vs mobile concerns, and since the first operation of this function is identical to that of SetPortraitView, I wonder if it may be better to instead do the following:

1. Rename `SetPortraitView()` to "UpscaleUI" or something just to represent the operation of making things better.
2. Then in this `SetMobileView()` function, you deactivate things

Not something I'm too worried about so this is a nit, but it could make things more clear wrt to the comments that Zack had

vontell (approved):

src/gg.regression.unity.bots/Runtime/Scripts/RGOverlay/RGSequenceManager.cs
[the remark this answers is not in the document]
addisonbgross: @vontell good call on the function rename, done 👍🏻 

@RG-nAmKcAz the portrait and mobile connection is that I'm making it so that the UI is scaled to be larger in both scenarios. When running a build on mobile, scale the UI up (and change a few other small things). When running a game in Unity + the viewport is in a portrait orientation, scale the UI up.

The overall goal is just to make the UI a bit larger in situations where it becomes harder to use (eg: on a small screen). Hopefully I've made some sense

PR #367, opened 2024-12-10
[REG-2200] Create TCP server to communicate with dashboard

abeizer:
## Tcp Server

`RGTcpServer` can start + stop a TCP listener on localhost:8085. `RGTcpManager` starts and stops the server as needed, and ensures functionality in both edit and play mode. 

Unity, for some reason, makes it very difficult to preserve variables when you transition between modes. I couldn't find any viable way to start the listener once and then maintain it across modes, so instead the manager class shuts down the server whenever we exit the current mode, then restarts it on `Start` (which is called when we enter the new mode). 

## Message handling

`RGTcpServer` handles the client handshake, but any messages after that are processed by callbacks exposed in the server class and implemented in `RGTcpManager`. The manager will be where we ultimately coordinate all of the data flowing between the client and Unity project. This PR introduces a handful of messages:
* receive ping from client and respond with pong
* send the available sequences to the client (this currently happens once on handshake)
* send the active sequence to the client (play mode only)
* receive request from client to play a given sequence by its resourcePath (play mode only)

## Dashboard

[View the client project here](https://github.com/Regression-Games/RGGameDashboard)

`RGTcpManager` creates a menu option for launching a `RegressionGames.exe` from the `RGUnityBots` Runtime/Resources directory. The .exe is included in this PR (if you want to try it out for yourself), but I'm not sure whether we want to check this file in now since it is extremely WIP.

The dashboard is designed to reconnect to the `RGTcpServer` whenever it shuts down. You can see data disappear from the dashboard for a moment when Unity switches between edit and play, but we currently send all the info it needs on handshake so it's more of a "blip" right now. Through the dashboard, you can view available sequences in the project, start one of them when you're in play mode, and see which sequence is currently running (if any).

## Out of scope / known issues / to solve in the future

* Client sometimes fails to connect when Unity is first launched. I believe it's creating a listener but some process like script reload is wiping the domain. I suspect that I'll need to handle before  + after compilation events. For now, pressing play successfully starts the server and populates the dashboard so there's an easy workaround
* This PR only includes an .exe application, so dashboard is windows-only for now
* Connection string is hardcoded to localhost:8085 on both ends
* The dashboard currently cannot differentiate whether Unity is in edit or play mode
* Available sequences are sent to the client whenever the mode changes. If you modify the project contents they won't automatically be reflected in the dashboard - you'd either need to refresh or relaunch the app through the Regression Games menu

## Loom

https://www.loom.com/share/2524a84744504a5090bcfc8c44da986f

---


src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/RGTcpManager.cs
```diff
@@ -0,0 +1,245 @@
...
+using System.IO;
+using System.Linq;
+using System.Net.Sockets;
+using JetBrains.Annotations;
+using RegressionGames.RemoteOrchestration.Models;
+using RegressionGames.RemoteOrchestration.Types;
+using RegressionGames.StateRecorder;
+using RegressionGames.StateRecorder.BotSegments;
+using RegressionGames.StateRecorder.BotSegments.Models;
+using UnityEditor;
```
RG-nAmKcAz, line 13: We're going to have issues in built games by always including UnityEditor 

see my other comment about how we may want to split this up

src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/RGTcpManager.cs
```diff
@@ -0,0 +1,245 @@
...
+            RGTcpServer.OnClientHandshake += OnClientHandshake;
+            RGTcpServer.ProcessClientMessage -= ProcessClientMessage;
+            RGTcpServer.ProcessClientMessage += ProcessClientMessage;
+            StartCoroutine(RGSequenceManager.ResolveSequenceFiles(ProcessResolvedSequences));
+        }
+        
+        /// <summary>
+        /// Assign or reset variables when the editor transitions to or from play mode
+        /// </summary>
+        private void OnPlayModeStateChanged(PlayModeStateChange state)
```
RG-nAmKcAz, line 70: I don't know this for certain, but I would bet that the 'i can't keep static things across a reload' is less to do with play vs edit mode and more to do with the code compilation and the context reset that happens because of it.

I would suggest we experiment with ...

triggers around `CompilationPipeline.compilationStarted` and `CompilationPipeline.compilationFinished`


I realize that is in the UnityEditor namespace and thus limited to functionality in the editor.


We may be able to make our lives easier here by separating out the management of the TCP server to editor specific and build specific code.

Make the build specific code do nothing if it is in the editor that we're playing, and put the editor specific code in the Editor package itself.


I am also curious if this alone would help us solve the instance problem if we were to make the tcp server variables instance variables instead of static, and manage a static on the editor class itself that represents the active TCPServer object.

RG-nAmKcAz (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/RGTcpManager.cs
RG-nAmKcAz: After discussing this, ... this PR will just handle adding the # checks so that production builds still work outside the editor on merge

abeizer: Okay I got a working build of BossRoom with my changes. The dashboard (client no.3) doesn't seem to be able to play sequences or receive the active sequence in a build, but there isn't currently a way for the user to open the dashboard in a build anyway... something we'll definitely need to address but later.

Also hooked into the `DidReloadScripts` editor event to make sure the client always receives necessary data on editor start.

RG-nAmKcAz (approved):

src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/RGTcpManager.cs
RG-nAmKcAz: suggestions for ways to open in a build you've likely already considered
- in game hotkey
- manually launch it from outside the game

batu (approved):
This looks good to me! I have nothing to add beyond what Zack mentioned. When I was writing the janky server for runespawn, I faced a similar issue and, well, didn't solve it.

Now that we have this, at some point we can consider moving runespawn tcp functionality to here as well.

PR #368, opened 2024-12-11
[REG-2217] Send available segments to dashboard

abeizer:
## What has been done

* Server now sends available segments to the client
* Added ability to start playing a segment
* Added ability to stop whatever is current playing in-game
* Updated the dashboard .exe to match contents of https://github.com/Regression-Games/RGGameDashboard/pull/1

## Out of Scope

Interacting with segments outside of playing + stopping them. We will be getting more interaction with segments (creation and editing) in upcoming PRs.

---


src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/RGTcpManager.cs
```diff
@@ -151,6 +177,12 @@ private void ProcessResolvedSequences(IDictionary<string, (string, BotSequence)>
             m_availableBotSequences = sequences.Select(kvp => new AvailableBotSequence(kvp.Key, kvp.Value.Item2)).ToList();
             SendAvailableSequences();
         }
+
+        private void LoadAvailableSegments()
```
RG-nAmKcAz, line 181: based on the method naming, I don't like how these methods implicitly send the status as well, maybe rename them to better represent that they do that ?

src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/RGTcpManager.cs
```diff
@@ -130,6 +134,27 @@ private void Update()
...
+                m_shouldStopReplay = false;
+            } 
+            else if (m_startPlayingSegment != null)
+            {
+                var botManager = RGBotManager.GetInstance();
+                if (botManager != null)
+                {
+                    botManager.OnBeginPlaying();
+                }
+                var playbackController = FindObjectOfType<BotSegmentsPlaybackController>();
```
RG-nAmKcAz, line 147: this is fine for the initial POC, but we'll need to decide when we're doing away with the in game overlay

for now.. they could easily conflict

either way, we should probably at least check here that so segment/sequence is running in case the client UI is somehow out of sync

src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/TcpMessage.cs
```diff
@@ -17,17 +18,22 @@ public enum TcpMessageType
...
         
         Pong,
         
-        // info about the available sequences for this game instance
+        // info about the available file-based resources for this game instance
         AvailableSequences,
+        AvailableSegments,
         
         // info about the currently-running sequence (or segment)
         ActiveSequence,
```
RG-nAmKcAz, line 39: Future (non-blocking), I have a task to create a status manager to better isolate the current status of which segment we're on, what the current error state is, whether exploration is active, and the current exploration error state.  We'll want to be able to sync that to the client, but I wanted to put this comment here for visibility into what is coming

RG-nAmKcAz (changes requested):
Mostly looks good, a few small requests to rename some methods and do an error case check.

src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/RGTcpManager.cs
abeizer: Updated the name to indicate that this does send segments after loading them

src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/RGTcpManager.cs
abeizer: Play now only works if something isn't already playing

src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/TcpMessage.cs
abeizer: Yeah, would love to adopt Addison's mockups into this dashboard. We have a lot more freedom here to show detailed information about the current segment, progression through a sequence, any errors, etc.

RG-nAmKcAz (approved):

addisonbgross (approved):

PR #370, opened 2024-12-16
[REG-2230, REG-2231] Implement Validations Within Segments

vontell:
Note to reviewers: I am headed to lunch rn so will make my own self-review changes when I get back, but I didn't want to stall having people take a look at this, so opening it now.

This review is now ready... This PR implements an initial validations-in-editor feature. This doesn't report or record the results of the validations (as that is going to be done next as part of [this ticket](https://linear.app/regression-games/issue/REG-2231/save-bot-segment-validation-results-to-a-json-file-and-viewable-within)), but it implements all of the internal-facing parts of this. More particularly, it does the following:

* Adds a new `validations` field to the bot segments which will contain the types of validations to run for that segment
* Implements a `Script` validation type that is essentially a C# file inheriting from a MonoBehaviour-esque class specifically for writing validations.
* Adds logic to the bot segment execution that will prepare these validation scripts and then run them. However, right now, the only type is a Script that manages its own update loop, so the ProcessValidations function doesn't really do much. I implemented things like pausing and cleanup, too.
* It is technically possible for a segment to finish running before validations can be tried... so there is a new endCriteria of "ValidationsCompleted" that makes the segment wait until all validations are met. **This is not typically needed from the user, it was really only for situations where you have a validation-only segment**.

Here is where I have some questions and might need some help with some debates:

* There are some tricky timing concerns with this... if the bot segment ends instantly (say there is one endCriteria that is instantly met), the validation script might only be loaded after it missed its chance to execute. Any thoughts on this?
* The way that validations are terminated is very tricky... for example, some validations are in the format of "this should never be true." This means that technically, this validation never ends... right now, this would be terminated by the endCriteria being met, in which case the validation script is told to stop, but what if they have a ValidationsCompleted endCriteria?

I'll make some notes about other concerns in the PR - TLDR; I am very open to changing things in here, there are some concepts and ways this work that I am not tied to.

If you are curious, here is the associated Boss Room PR: https://github.com/Regression-Games/RGBossRoom/pull/81


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
```diff
@@ -143,6 +143,40 @@ public void PauseAction()
             }
         }
 
+        public void ProcessValidation()
+        {
+            // Go through each validation and process them. If they have not been started yet, this will also
+            // start them.
+            foreach (var validation in validations)
+            {
+                validation.ProcessValidation(Replay_SegmentNumber);
```
vontell, line 152: Note to myself in my self-review - I need to remove these replay segment numbers, they are not needed by my code I think...

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/SegmentValidations/ScriptSegmentValidationData.cs
```diff
@@ -0,0 +1,200 @@
...
+                }
+
+                if (_myGameObject != null)
+                {
+                    if (!_myGameObject.TryGetComponent(_typeToCreate, out _))
+                    {
+                        ((IRGSegmentValidationData)this).StopValidation(segmentNumber);
+                    }
+
+                    // TODO(vontell): How do I want to support rogue scripts that keep running?
```
vontell, line 123: Ideas here? I think I am going to add a timeout option to these validation scripts...

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentsPlaybackController.cs
```diff
@@ -455,6 +458,7 @@ private void EvaluateBotSegments()
                         BotSegment firstActionSegment = _nextBotSegments[0];
                         try
                         {
+                            firstActionSegment.ProcessValidation();
```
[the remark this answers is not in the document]
RG-nAmKcAz: Nit: I would put this AFTER doing the action

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
RG-nAmKcAz: leave them in and use them to prefix all log statements with `(#) `... makes it much easier to correlate visually what segment validations are running for

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentsPlaybackController.cs
```diff
@@ -486,6 +490,7 @@ private void EvaluateBotSegments()
                         nextBotSegmentIndex == 0,
                         nextBotSegment.Replay_SegmentNumber,
                         nextBotSegment.Replay_ActionCompleted,
+                        nextBotSegment.Replay_ValidationsCompleted,
                         nextBotSegment.endCriteria
                     );
```
RG-nAmKcAz, line 496: about line 520 in your new code.. there is logic for waiting for actions to complete even if the user didn't specify the end criteria to wait for actions to complete....

I feel like this should follow a similar pattern...

After the action has stopped AND we're in that hasMatched endCriteria section...
we then would tell the validation to stop.  At this point, any needed criteria that should have matched and haven't would fail.  However, this would solve you issue where you have things that should 'never' be a condition.. at this point you would mark them as passed if they were still correct.

src/gg.regression.unity.bots/Runtime/Scripts/Validation/RGValidateBehaviour.cs
```diff
@@ -0,0 +1,240 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using StateRecorder.BotSegments.Models.SegmentValidations;
+using UnityEngine;
+using Object = UnityEngine.Object;
+
+namespace RegressionGames.Validation
+{
+    public class RGValidateBehaviour: MonoBehaviour
```
RG-nAmKcAz, line 11: We should model this differently.  Instead of spawning monobehaviours and running into timing issues where endCriteria match before we've even spawned these, why not handle instantiation/evaluation as part of our botsegment evaluation loop like actions/endConditions are done ?

So instead of these being behaviours that self process on the Update loop, they would be implementations of an interface that a ValidationEvaluator behaviour driven by our botsegment update loop would process through on each update pass.

This would remove your ambiguous timing issues.

The one user code difference is that these wouldn't be MonoBehaviour objects anymore, but given that we aren't parenting these to anything useful anyway, I see that as an improvement.  The APIs the user needs to find objects are still available to them MonoBehaviour or not because we'll be calling them from the main thread Update loop.

RG-nAmKcAz, line 11: We should model this differently. Instead of spawning monobehaviours and running into timing issues where endCriteria match before we've even spawned these, why not handle instantiation/evaluation as part of our botsegment evaluation loop like actions/endConditions are done ?

So instead of these being behaviours that self process on the Update loop, they would be implementations of an interface that a ValidationEvaluator behaviour driven by our botsegment update loop would process through on each update pass.

This would remove your ambiguous timing issues.

The one user code difference is that these wouldn't be MonoBehaviour objects anymore, but given that we aren't parenting these to anything useful anyway, I see that as an improvement. The APIs the user needs to find objects are still available to them MonoBehaviour or not because we'll be calling them from the main thread Update loop.

nAmKcAz (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentsPlaybackController.cs
vontell: Perfect, that is exactly what I was looking for.

Does this comment mean that I should _only_ do this? What did you think of the "ValidationsCompleted" endCriteria type? I could see a case for both.

src/gg.regression.unity.bots/Runtime/Scripts/Validation/RGValidateBehaviour.cs
vontell: I was going to counter with why I want it to be a monobehaviour, but after going through the functions and properties available in that class, I suppose I can't support my claim... I was thinking maybe they'd want access to things like mouse event handlers, etc... but with the opinionated way I want them to use this class, they shouldn't do that.

TLDR; I'll try out your approach since it will simplify some of the internals as well. Easy to change in the future if needed. I put what I was writing below so you can see my thought process that I abandoned while writing lol

----


Hmmm that is one option... yeah I was primarily concerned with the ability to use things that monobehaviours make available with ease.

But I guess to clarify, I feel like my timing issues weren't because of the update loop - actually, the suggested approach is somewhat equivalent but I am just using another gameobject's update loop to call my class's fake update loops, instead of having unity call it directly (since Unity goes through all update loops serially).

For "they would be implementations of an interface that a ValidationEvaluator class driven..." that also has been added, the validation segments can use the `ProcessValidations` function for that, and my intent here is that most of the types would - it just didn't for this one.

src/gg.regression.unity.bots/Runtime/Scripts/Validation/README.md
```diff
@@ -0,0 +1,248 @@
+# Prototype - Code-based Validations
```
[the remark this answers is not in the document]
vontell: Removed and will add back new docs later today.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/SegmentValidations/ScriptSegmentValidationData.cs
```diff
@@ -0,0 +1,200 @@
...
+                    //     // This represents a segment failure case where the timeout stops the Behaviour that did not find its criteria and end in a timely manner
+                    //     // we destroy the behaviour + gameObject, but don't mark this stopped so that error reporting works
+                    //     DestroyGameObject();
+                    //     _error = $"Behaviour did not complete within maxRuntimeSeconds: {maxRuntimeSeconds.Value}";
+                    //     error = _error;
+                    //     return false;
+                    // }
+
+                    // Validation is expected to perform its actions in its own 'Update' or 'LateUpdate' calls... we don't directly call it
+                    // TODO(vontell): When it reaches its own self-determined end condition, it should destroy itself
```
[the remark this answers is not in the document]
vontell: This will be adjusted as part of my behaviour rewrite

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
vontell: Added!

src/gg.regression.unity.bots/Editor/Scripts/Validation/RGValidationPane.cs
```diff
@@ -0,0 +1,97 @@
+﻿using RegressionGames.Validation;
+using StateRecorder.BotSegments.Models.SegmentValidations;
+using UnityEditor;
+using UnityEngine;
+using UnityEngine.UIElements;
+
+namespace RegressionGames.Editor.Validation
```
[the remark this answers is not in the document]
vontell: This has been removed

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/SegmentValidations/ScriptSegmentValidationData.cs
vontell: I added a timeout option.

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/SegmentValidations/ScriptSegmentValidationData.cs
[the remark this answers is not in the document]
vontell: Removed/added really I guess (added the part where it actually does process based on some outside update loop)

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentsPlaybackController.cs
vontell: Ok I think I got this working in a pretty decent way, was a little weird at first. There are two options:

1. If you add an endCriteria of ValidationsCompleted, it will wait for the validations to be set before ending the segment. There is also a timeout now so you can wait a certain amount of time and then end the segment anyways (this does not stop the next segment from moving on)
2. OR, if you don't provide a ValidationsCompleted criteria, it will automatically stop the validations as soon as all criteria and actions have been met. Technically you can still include a timeout, but I don't think that would be used much in this situation.

src/gg.regression.unity.bots/Runtime/Scripts/Validation/RGValidateBehaviour.cs
vontell: This has now been updated to use the update loop of the bot segment playback controller 👍

vontell:
@RG-nAmKcAz this is now ready for an actual review, thank you for the high level points, I think things are much cleaner now. I do need to work on the merge conflicts since I bet there are a lot of changes in bot segments I might need to worry about... but will try it out.

Also update the boss room PR with the relevant changes: https://github.com/Regression-Games/RGBossRoom/pull/81

RG-nAmKcAz:
@vontell The merge will restructure your code in the Controller completely, so let's do that first

vontell:
Back to you, @RG-nAmKcAz. I do need to add even more tests to bossroom, mine are very very simple, but I will be doing that tonight. No rush to review this tonight of course, but note that I do plan on giving Kachi an experimental branch to play around with since I promised them a timeline (this was intentional, I **needed** to get this to them in the first half of the week), but I expect to continuously making fixes over the next few days depending on whether they use it or not.

Appreciate all the comments so far!

vontell:
Update on this from evening work: I actually need slightly more time on this to think through a few small pieces... see this video if you are interested in seeing what I changed (hint: I reasoned that I need to support validations over sequences or multiple segments to really get any utility out of this). https://www.loom.com/share/31165d0f9a9f471db710b1d3b48f68a3?sid=4c8b3ec9-b2e9-4882-93ab-c295353fe804

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentsPlaybackController.cs
```diff
@@ -359,11 +366,20 @@ private void ProcessBotSegmentAction(BotSegment firstActionSegment, Dictionary<l
                     _explorationDriver.StopExploring(firstActionSegment.Replay_SegmentNumber);
                     _explorationDriver.ReportPreviouslyCompletedAction(firstActionSegment.botAction.data);
                 }
-
+                
+                // Run (and potentially end) the validations for this segment
+                firstActionSegment.ProcessValidation();
```
RG-nAmKcAz, line 371: ProcessValidations calls should be moved back to ProcessBotSegments, just after the action, but not embedded in ProcessBotSegmentAction

abeizer (approved):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/SegmentValidations/SegmentValidationStatus.cs
```diff
@@ -0,0 +1,9 @@
+namespace StateRecorder.BotSegments.Models.SegmentValidations
+{
+    public enum SegmentValidationStatus
+    {
+        PASSED,
+        FAILED,
+        UNKNOWN
```
batu, line 7: Would you mind outlining in what cases it could be UNKNOWN? I am making assumptions but I would rather be clear.

src/gg.regression.unity.bots/Runtime/Scripts/Validation/RGValidationScript.cs
```diff
@@ -0,0 +1,251 @@
...
+                }
+                else if (currentValidator.Mode == ValidationMode.ONCE_TRUE_ALWAYS_TRUE &&
+                         currentValidator.Status == SegmentValidationStatus.UNKNOWN &&
+                         currentStatus == SegmentValidationStatus.PASSED)
+                {
+                    // Debug.LogError($"Validation method {validator.Method.Name} is now passing.");
+                    currentValidator.Status = SegmentValidationStatus.PASSED;
+                    OnValidationsUpdated?.Invoke();
+                } 
+                else if (currentValidator.Mode == ValidationMode.ONCE_TRUE_ALWAYS_TRUE &&
```
batu, line 193: Courtesy of O1:

Here, if the validator has never passed before (it is UNKNOWN) and the current check fails, you immediately mark it as FAILED. This could be problematic depending on the intended semantics:

The documentation states: "Once there is a frame that the given condition turns true, it must always be marked as true." It does not explicitly say that if the condition fails before ever becoming true, it should never have a chance to pass later.

By failing immediately on a single failed invocation before a pass, you deny the validator any future opportunity to eventually pass. This might not be what you want logically.

src/gg.regression.unity.bots/Runtime/Scripts/Validation/RGValidationScript.cs
```diff
@@ -0,0 +1,251 @@
...
+                }
+                else if (validator.Mode == ValidationMode.EVENTUALLY_TRUE && validator.Status is SegmentValidationStatus.UNKNOWN or SegmentValidationStatus.FAILED)
+                {
+                    //Debug.LogError($"Validation method {validator.Method.Name} should eventually pass, but it never did.");
+                    validator.Status = SegmentValidationStatus.FAILED;
+                    OnValidationsUpdated?.Invoke();
+                }
+                else if (validator.Mode == ValidationMode.ONCE_TRUE_ALWAYS_TRUE && validator.Status == SegmentValidationStatus.UNKNOWN)
+                {
+                    //Debug.LogError($"Validation method {validator.Method.Name} was never passed or failed, so it is technically passed.");
```
batu, line 243: This is counter intuitive to me. I would have expected this to fail. I would think: "If it never passed or failed, it technically didn't pass."

Can you explain it more?

src/gg.regression.unity.bots/Runtime/Scripts/Validation/RGValidationScript.cs
```diff
@@ -0,0 +1,251 @@
...
+                        Frequency = attribute.Frequency,
+                        Mode = attribute.Mode
+                    });
+                    RGDebug.Log("[RGValidator] Activated " + method.Name);
+                } 
+            }
+            
+        }
+
+        public void AssertAsTrue(string message = null)
```
batu, line 68: It took me a minute to understand how currentStatus can be updated within ProcessValidations. Please confirm that the idea is for the validators to call these functions? If so would you mind documenting that?

batu (changes requested):
I would like to better understand
`//Debug.LogError($"Validation method {validator.Method.Name} was never passed or failed, so it is technically passed.");` but beyond that it looks good.

I skimmed the infrastructure and set up parts, nothing popped out to me but I am deferring to Zack's comments there.

batu:
@vontell For my understanding and speeding up reviews, this is a good example of a PR that I would have appreciated a top level walkthrough in the form of a video. I am realizing that I spend a good amount of time identifying which of the 30+ file changes I should give my attention and trying to simulate the code flow to resolve assumptions. Things clicked when I got to [BossRoom/.../ValidateUI.cs](https://github.com/Regression-Games/RGBossRoom/blob/52f9c0c184b947ef3f5f638e93edb1b0e8fefda1/Assets/Scripts/RegressionGames/ValidateUI.cs) but it took a sec. I think the 'code walkthrough video' helps especially when I don't have the context in (SDK in this case) in my recent memory.

nAmKcAz (approved):
Aaron, marking this approved while you test / iterate as I agree with the fundamental design.

nAmKcAz (approved):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/SegmentValidations/SegmentValidationStatus.cs
vontell: Comments added 👍

src/gg.regression.unity.bots/Runtime/Scripts/Validation/RGValidationScript.cs
vontell: You are correct. I have removed this check!

src/gg.regression.unity.bots/Runtime/Scripts/Validation/RGValidationScript.cs
vontell: Yeah this is a good philosophical point... I could be swayed either way. I suppose you are right that in these cases, the user would expect that if it never turned true, they expected it to since they made this test. Going to make this a failure instead.

src/gg.regression.unity.bots/Runtime/Scripts/Validation/RGValidationScript.cs
vontell: Yes that is correct! Added some more docs.

vontell:
@batu back to you! Really like your suggestion of a loom video, I will do that in the future upfront, but also for other reviewers here is that video: https://www.loom.com/share/d6e9a506ccbd4605bdb7bdaa0d2448a0?sid=8ce1048a-d759-481f-b059-8fd58e98c046

batu (approved):

PR #372, opened 2024-12-17
[REG-2213] Create playback status manager

RG-nAmKcAz:
- Creates a Playback Status Manager Behaviour that holds/manages the current status of segment/sequence playback
- Moves all on screen display and logging from the playback controller into this new Behaviour
- Updates the exploration logic to be far less log spammy and flicker the screen less

---


RG-nAmKcAz:
@abeizer This one is really targeted at your work, but it also helped me cleanup some of the mess in the playback controller.

There is now a new Behaviour `BotSegmentPlaybackStatusManager` that manages the status of playback to make it easy to see where we are and what the current error states are.

abeizer (approved):
It'll be great to have all this info easily available for the UI. Maybe after these initial layouts for CV segments we can rework the "current play status" in the dashboard to look more like Addison's past mockups

vontell (approved):
Nice! Thanks for putting this together to make Abby's work easier. I also think encapsulating like this could be helpful in making more debug features in the future

PR #374, opened 2024-12-18
[REG-2230] Operate on BotSegmentLists so we can do validations there

RG-nAmKcAz:
- Changes the container to use lists of BotSegmentLists instead of lists of botsegments so that we can do validations at the BotSegmentList level

---


src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentsPlaybackController.cs
```diff
@@ -108,6 +110,21 @@ private void ProcessBotSegments()
                         ProcessBotSegmentAction(nextBotSegment, transformStatuses, entityStatuses);
                     }
 
+                    // Run (and potentially end) the validations for this segment
+                    nextBotSegment.ProcessValidation();
+
+                    // Also run the validations for the botsegmentList if they exist
+                    foreach (var validation in _botSegmentListValidations)
+                    {
+                        validation.ProcessValidation(nextBotSegment.Replay_SegmentNumber);
```
vontell, line 119: So this is passing down the segment number, but the validation is not really part of that segment, it is part of that segment list as a whole... do we think this could confuse the user? This is why I used -1 for sequences, and then in the logs that gets converted to "SEQUENCE"

vontell (commented):
I'll check out the branch and see if it aligns with what I expect, but in the meantime just one comment around the logging of work on validations that are not technically running as part of a particular segment

vontell (changes requested):
(forgot to mark as request changes)

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentsPlaybackController.cs
RG-nAmKcAz: i want to see the active segment in the logs as this runs so I can see what segment I was on as the validation was running

RG-nAmKcAz: they are absolutely being evaluated during that currently active segment, and should log as such

vontell: Sounds good to me 👍 I can get behind that logic

vontell (approved):

RG-nAmKcAz:
@vontell  feel free to merge (will go into your branch) ... as soon as you're done testing / happy with it

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/BotSegmentsPlaybackController.cs
```diff
@@ -192,7 +209,7 @@ private void ProcessBotSegments()
                         {
                             if (_nextBotSegments.Count < 2)
                             {
-                                var next = _dataPlaybackContainer.DequeueBotSegment();
+                                var next = _dataPlaybackContainer.DequeueBotSegment(out _botSegmentListValidations);
```
vontell, line 212: @RG-nAmKcAz quick question on this actually regarding the lookahead logic. Does this mean that the currently running validation will be lost and replaced? I am not full understanding this code yet but I wanted to check if this means that validations that are currently running and not yet finished may then be running forever, since now we only operate on these new validations. Maybe we don't need to worry about that because we run StopValidations on the data container itself which does hold onto the reference...

RG-nAmKcAz: i'll huddle with you on this

Stop I think we're fine

but you correctly point out that we may not be considering all the proper validations... I want to discuss if we get rid of this look ahead logic all together or not

vontell:
Closing this PR as it has now been merged into the other feature branch from #375

PR #373, opened 2024-12-18
[REG-2231] Ability to save validation results and view results in the logs

vontell:
This will not be merged in directly, this is just demonstrating the changes from my other PR to this one, to enable saving of the JSON. Please see the full video in my other PR for more on this in the meantime!

PR #375, opened 2024-12-19
[REG-2230 Bot Segment List Support + REG-2231 Saving the results]

vontell:
Ok so my branch structure to fix things got really messed up... so I started by working off of Zack's branch to support bot segment lists, and then I pulled in my REG-2231 branch to resolve a compilation error caused by the new location of validations in some of these datastructures. The unfortunate result of this, which I should have done differently, is that my adjustments to Zack's patch PR into REG-2230 is now interwined with my REG-2231 changes.

Anyways, this PR brings this small saving results feature and the bot segment list fixes into my original PR, which has already been approved. Zack and I also already discussed these changes so nothing should be a surprise here hopefully... once I merge this into my original branch, we can do a final review of that PR.

The alternative is that we merge the aaron/reg-2230 branch in main first and then this into main, so we don't need to re-review that other PR, but then we have a half-working prototype in main which I want to avoid...

Once this is approved and merged, we can get rid of #374 and #373


vontell:
Approved by other team members that this is fine to merge in and review all at once

PR #376, opened 2024-12-20
[REG-2222] Support creating and editing segments through the dashboard app 

abeizer:
## What has been done
* Added TCP message types for the client to request a specific segment and overwrite the contents of a specific segment using its `resourcePath`
* Helper methods added to `BotSegment` to facilitate reading + writing segment files
* Reworked TCPServer logic to properly handle fragmented messages

## Out of scope
* Creating new BotSequences and viewing/editing the contents of BotSequences
* server ack on segment save -> I want to do this but ran out of time before the holiday break

---

### DIFF
--- a/src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/TcpMessage.cs
+++ b/src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/TcpMessage.cs
@@ -10,39 +10,6 @@ using RegressionGames.StateRecorder.JsonConverters;
 
 namespace RegressionGames.ClientDashboard
 {
-    public enum TcpMessageType
-    {
-        // =====================
-        // client -> server
-        // =====================
-        
-        Ping,
-        
-        // client requests to play a resource with the given resourcePath
-        PlaySequence,
-        PlaySegment,
-        
-        // stops any currently-running sequence/segments
-        StopReplay,
-        
-        // =====================
-        // server -> client
-        // =====================
-        
-        Pong,
-        
-        // info about the available file-based resources for this game instance
-        AvailableSequences,
-        AvailableSegments,
-        
-        // info about the currently-running sequence (or segment)
-        ActiveSequence,
-        
-        // sent prior to closing Unity windows.
-        // this tells any running client windows to also close.
-        CloseConnection
-    }
-
     public interface ITcpMessageData : IStringBuilderWriteable { }
     
     /// <summary>
@@ -148,6 +115,8 @@ namespace RegressionGames.ClientDashboard
                 IntJsonConverter.WriteToStringBuilder(stringBuilder, currentSegment.apiVersion);
                 
                 // not normally serialized
+                stringBuilder.Append(",\"path\":");
+                StringJsonConverter.WriteToStringBuilder(stringBuilder, currentSegment.path);
                 stringBuilder.Append(",\"resourcePath\":");
                 StringJsonConverter.WriteToStringBuilder(stringBuilder, currentSegment.resourcePath);
                 stringBuilder.Append(",\"type\":");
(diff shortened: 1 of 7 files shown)


src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/TcpMessage.cs
```diff
@@ -195,4 +164,91 @@ public override string ToString()
...
+        public void WriteToStringBuilder(StringBuilder stringBuilder)
+        {
+            stringBuilder.Append("{\"resourcePath\":");
+            StringJsonConverter.WriteToStringBuilder(stringBuilder, resourcePath);
+            stringBuilder.Append("}");
+        }
+        
+        public override string ToString()
+        {
+            StringBuilder sb = new StringBuilder(1000);
```
nAmKcAz, line 181: make this a readonly threadlocal so we avoid re-allocating/gc'ing it

src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/TcpMessage.cs
```diff
@@ -195,4 +164,91 @@ public override string ToString()
...
+        public void WriteToStringBuilder(StringBuilder stringBuilder)
+        {
+            stringBuilder.Append("{\"jsonObject\":");
+            stringBuilder.Append(jsonContent); 
+            stringBuilder.Append("}");
+        }
+        
+        public override string ToString()
+        {
+            StringBuilder sb = new StringBuilder(1000);
```
nAmKcAz, line 205: make this a readonly threadlocal so we avoid re-allocating/gc'ing it

src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/TcpMessage.cs
```diff
@@ -195,4 +164,91 @@ public override string ToString()
...
+            stringBuilder.Append("{\"segmentList\":");
+            segmentList.WriteToStringBuilder(stringBuilder);
+            stringBuilder.Append(",\"resourcePath\":");
+            StringJsonConverter.WriteToStringBuilder(stringBuilder, resourcePath);
+            stringBuilder.Append("}");
+        }
+        
+        public override string ToString()
+        {
+            StringBuilder sb = new StringBuilder(1000);
```
nAmKcAz, line 228: make this a readonly threadlocal so we avoid re-allocating/gc'ing it

src/gg.regression.unity.bots/Runtime/Scripts/ClientDashboard/TcpMessage.cs
```diff
@@ -195,4 +164,91 @@ public override string ToString()
...
+        public void WriteToStringBuilder(StringBuilder stringBuilder)
+        {
+            stringBuilder.Append("{\"resourcePath\":");
+            StringJsonConverter.WriteToStringBuilder(stringBuilder, resourcePath);
+            stringBuilder.Append("}");
+        }
+        
+        public override string ToString()
+        {
+            StringBuilder sb = new StringBuilder(1000);
```
nAmKcAz, line 248: make this a readonly threadlocal so we avoid re-allocating/gc'ing it

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
```diff
@@ -360,6 +373,60 @@ public void WriteToStringBuilder(StringBuilder stringBuilder)
 
             return results;
         }
+        
+        /**
+         * <summary>
+         * Saves a BotSegmentList to a file
+         * </summary>
+         * <param name="segmentList">The BotSegmentList to save to file</param>
+         * <param name="resourcePath">The segment's resourcePath, if it already exists. If null, a new path will be generated for it.</param>
```
nAmKcAz, line 382: should probably update this comment to indicate that the resource path should NOT include the '.json' suffix

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
```diff
@@ -360,6 +373,60 @@ public void WriteToStringBuilder(StringBuilder stringBuilder)
...
+
+        /**
+         * <summary>
+         * Delete the BotSegment or BotSegmentList with the given resourcePath 
+         * </summary>
+         */
+        public static void Delete(string resourcePath)
+        {
+            resourcePath = resourcePath.Replace('\\', '/');
+            if (!resourcePath.StartsWith("Assets/"))
```
nAmKcAz, line 422: This is only valid if running in the editor.  In built runtime this should point to persistent data path.

nAmKcAz (changes requested):

src/gg.regression.unity.bots/Runtime/Scripts/StateRecorder/BotSegments/Models/BotSegment.cs
abeizer: Deletion isn't allowed during a build right? This should be similar to what we're doing from the overlay currently, just for a segment versus a sequence

nAmKcAz: If you create a copy/override of a sequence/segment packaged in the build, it is written to the persistent data path.  You CAN delete that override copy.  

There is also an existing code path that deletes the prior existing after writing the new one when renaming.

vontell (approved):
No changes other than what has already been requested, approving this so we can get these off the notifications list.

